PR #11, opened 2017-09-27
hash operator() must not throw exceptions

vladon:
hash operator() must not throw exceptions, so let it be noexcept


tituswinters:
We can't do that in this case - the standard doesn't specify this and absl::optional is a pre-adopted type.  
We could probably comment somewhere that (many) hash containers don't allow throwing if you'd like to be explicit about it.

vladon:
@tituswinters the Standard guarantees that any member of `std::hash` (including `std::hash::operator()`) will not throw exceptions

tituswinters:
http://eel.is/c++draft/unord.hash#3 "except as noted otherwise"
http://eel.is/c++draft/optional.hash "are not guaranteed to be noexcept"

Am I missing something?

gennadiycivil:
Looks like this is conclusive. Closing

PR #19, opened 2017-10-01
Make any_internal::FastTypeId() constexpr

Bronek:
This means removing all side effects from FastTypeId(). So rather than instantiate dummy_var in the first call to FastTypeId(), move this responsibility to the linker, and only read its address during execution - guaranteed to never change.

### DIFF
--- a/.ci/abseil-cpp.json
+++ b/.ci/abseil-cpp.json
@@ -1,0 +1,8 @@
+// This is a relaxed JSON format, you can have comments in it.
+// This is a list of configuration for the job that does not specify a configuration.
+[
+    {"node": "linux-x86_64"},
+    {"node": "ubuntu_16.04-x86_64"},
+    {"node": "darwin-x86_64"},
+    {"node": "windows-x86_64"}
+]


calabrese-google (approved):
What is the rationale for rewriting this to be constexpr when it is internal and the callers are not constexpr?

absl/types/any.h
```diff
@@ -103,14 +111,10 @@ namespace any_internal {
...
-inline size_t FastTypeId() {
+constexpr inline size_t FastTypeId() {
   static_assert(sizeof(char*) <= sizeof(size_t),
                 "ptr size too large for size_t");
-
-  // This static variable isn't actually used, only its address, so there are
-  // no concurrency issues.
-  static char dummy_var;
-  return reinterpret_cast<size_t>(&dummy_var);
+  return reinterpret_cast<size_t>(&TypeTag<Type>::dummy_var);
```
calabrese-google: Actually, backtracking a bit -- this will not make the call constexpr because of the reinterpret_cast. What is the precise intent here in making it constexpr, given no constexpr uses? We can make this constexpr if we change FastTypeId to not return a size_t, but rather return the pointer directly.

Bronek: I would expect at the very least this will get rid of static initialisation in the function scope - which compiler needs to enforce synchronization of (as per section 6.7 , paragraph 4). I imagine under any sane optimizer this synchronization would be removed, but trying to make optimizer work easier here and the intent more obvious - both to compiler and to reader.

Yes it is slightly embarrassing that this function should not be allowed to be `constexpr` by a conforming compiler. If returned value was `const void*` or perhaps `const char *`, that would not have been a problem. I would even defer the cast to size_t (or perhaps more fittingly, `uintptr_t`) until the body of `GetObjTypeId()` where we have conditional expression on runtime condition anyway.

calabrese-google: Could you update the return to simply be `const void*` instead, and remove the `reinterpret_cast` as a part of this change? This would require cascading changes to the return type in `IdForType()` and `GetObjTypeId()` as well. Thanks.

Bronek: Thanks Matt, will do.

Bronek:
I found that with this change I can add `constexpr` to two overloads of `any_cast` (these taking/returning pointers), but that would be stepping on the toes of the C++ committee, so will leave it out.

calabrese-google:
Sorry for the delay -- we are currently in the process of making external pull requests faster/easier, but we don't yet have linting in place. The change to the `GetObjTypeId()` function goes over our 80 column limit. Could you wrap the line?

calabrese-google:
`GetObjTypeId()` also cannot be `constexpr`, as the `unique_ptr` comparison is not `constexpr`.

calabrese-google:
Sorry, one last request (I promise). Can you rebase?

Things will be much smoother soon.

bazel-io:

bazel-io:

Bronek:
Replaced with https://github.com/abseil/abseil-cpp/pull/28

PR #20, opened 2017-10-02
Don't calculate the length of a constexpr string at runtime

sabbakumov:


tituswinters:
It isn't required, but AFAIK every supported compiler will inline the call to strlen on a compile-time constant string.   Do you have evidence to the contrary?

sabbakumov:
No, I don't have.
I think it's a good idea to use "constexpr" to its fullest.

tituswinters:
On the one hand, yes, constexpr is a good thing when available. On the other hand, it's already being done that way implicitly.  (Admittedly, without enforcement.)  

On net, I'd rather not make the code harder to read for no technical benefit.  If any of our supported compilers were failing to inline that strlen call and it seemed like it mattered, I'd gladly go down this path.  For now, I'd rather keep it simple.

PR #24, opened 2017-10-06
Experimental: Change to SFINAE test to support MSVC 2015 R1

vslashg:
Do not submit this; it is an experimental PR.


bazel-io:

PR #28, opened 2017-10-06
Make any_internal::FastTypeId() and IdForType() constexpr

Bronek:
This means removing all side effects from FastTypeId(). So rather than instantiate dummy_var in the first call to FastTypeId(), move this responsibility to the linker, and only read its address during execution - guaranteed to never change. This allows for more optimization opportunities, with more explicit uses of constexpr

### DIFF
--- a/absl/types/any.h
+++ b/absl/types/any.h
@@ -103,8 +103,8 @@ template <typename Type>
 constexpr char TypeTag<Type>::dummy_var;
 
 // FastTypeId<Type>() evaluates at compile/link-time to a unique pointer for the
-// passed in type. These are meant to be good match for keys into maps or straight
-// up comparisons.
+// passed in type. These are meant to be good match for keys into maps or
+// straight up comparisons.
 template<typename Type>
 constexpr inline const void* FastTypeId() {
   return &TypeTag<Type>::dummy_var;
--- a/absl/types/any.h
+++ b/absl/types/any.h
@@ -383,16 +383,6 @@ class any {
 #if ABSL_ANY_DETAIL_HAS_RTTI
     virtual const std::type_info& Type() const noexcept = 0;
 #endif  // ABSL_ANY_DETAIL_HAS_RTTI
-
-    // Note that on 64-bit (unix) systems size_t is 64-bit while int is 32-bit and
-    // the compiler will happily and quietly assign such a 64-bit value to a
-    // 32-bit integer. While a client should never do that it SHOULD still be safe,
-    // assuming the BSS segment doesn't span more than 4GiB.
-    size_t type_id() const noexcept {
-      static_assert(sizeof(void*) <= sizeof(size_t),
-                    "ptr size too large for size_t");
-      return reinterpret_cast<size_t>(ObjTypeId());
-    }
   };
 
   // Hold a value of some queryable type, with an ability to Clone it.
@@ -431,8 +421,7 @@ class any {
   }
 
   const void* GetObjTypeId() const {
-    return obj_ == nullptr ? any_internal::FastTypeId<void>()
-                           : obj_->ObjTypeId();
+    return obj_ ? obj_->ObjTypeId() : any_internal::FastTypeId<void>();
   }
 
   // `absl::any` nonmember functions //
--- a/absl/types/any.h
+++ b/absl/types/any.h
@@ -94,23 +94,20 @@ namespace absl {
 
 namespace any_internal {
 
-// FastTypeId<Type>() evaluates at compile/link-time to a unique integer for the
-// passed in type. Their values are neither contiguous nor small, making them
-// unfit for using as an index into a vector, but a good match for keys into
-// maps or straight up comparisons.
-// Note that on 64-bit (unix) systems size_t is 64-bit while int is 32-bit and
-// the compiler will happily and quietly assign such a 64-bit value to a
-// 32-bit integer. While a client should never do that it SHOULD still be safe,
-// assuming the BSS segment doesn't span more than 4GiB.
-template<typename Type>
-inline size_t FastTypeId() {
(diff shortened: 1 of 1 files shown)


bazel-io:

absl/types/any.h
```diff
@@ -94,23 +94,20 @@ namespace absl {
...
-// passed in type. Their values are neither contiguous nor small, making them
-// unfit for using as an index into a vector, but a good match for keys into
-// maps or straight up comparisons.
-// Note that on 64-bit (unix) systems size_t is 64-bit while int is 32-bit and
-// the compiler will happily and quietly assign such a 64-bit value to a
-// 32-bit integer. While a client should never do that it SHOULD still be safe,
-// assuming the BSS segment doesn't span more than 4GiB.
+template <typename Type>
+struct TypeTag {
+  constexpr static char dummy_var = 0;
```
EricWF, line 99: If Abseil supports C++17 inline variables, this should be that when available.

absl/types/any.h
```diff
@@ -94,23 +94,20 @@ namespace absl {
...
-inline size_t FastTypeId() {
-  static_assert(sizeof(char*) <= sizeof(size_t),
-                "ptr size too large for size_t");
-
-  // This static variable isn't actually used, only its address, so there are
-  // no concurrency issues.
-  static char dummy_var;
-  return reinterpret_cast<size_t>(&dummy_var);
+constexpr inline const void* FastTypeId() {
+  return &TypeTag<Type>::dummy_var;
```
EricWF, line 110: If you have a constexpr `std::addressof` here, you'll want that to avoid ADL on the template parameter `Type`.

absl/types/any.h
```diff
@@ -415,16 +422,17 @@ class any {
...
     using NormalizedType =
         typename std::remove_cv<typename std::remove_reference<T>::type>::type;
 
     return any_internal::FastTypeId<NormalizedType>();
   }
 
-  size_t GetObjTypeId() const {
-    return obj_ == nullptr ? any_internal::FastTypeId<void>() : obj_->type_id();
+  const void* GetObjTypeId() const {
+    return obj_ == nullptr ? any_internal::FastTypeId<void>()
```
EricWF:
```c++
return obj_ ? obj_->ObjTypeId() : any_internal::FastTypeId<void>()
```
Or is the above order oddly intended?

absl/types/any.h
```diff
@@ -382,10 +379,20 @@ class any {
...
+    virtual const void* ObjTypeId() const noexcept = 0;
 #if ABSL_ANY_DETAIL_HAS_RTTI
     virtual const std::type_info& Type() const noexcept = 0;
 #endif  // ABSL_ANY_DETAIL_HAS_RTTI
+
+    // Note that on 64-bit (unix) systems size_t is 64-bit while int is 32-bit and
+    // the compiler will happily and quietly assign such a 64-bit value to a
+    // 32-bit integer. While a client should never do that it SHOULD still be safe,
+    // assuming the BSS segment doesn't span more than 4GiB.
+    size_t type_id() const noexcept {
```
EricWF: We're casting a pointer to a integral type. I'm suspect we should use `uintptr_t`.

absl/types/any.h
```diff
@@ -382,10 +379,20 @@ class any {
    public:
     virtual ~ObjInterface() = default;
     virtual std::unique_ptr<ObjInterface> Clone() const = 0;
-    virtual size_t type_id() const noexcept = 0;
+    virtual const void* ObjTypeId() const noexcept = 0;
 #if ABSL_ANY_DETAIL_HAS_RTTI
     virtual const std::type_info& Type() const noexcept = 0;
 #endif  // ABSL_ANY_DETAIL_HAS_RTTI
+
+    // Note that on 64-bit (unix) systems size_t is 64-bit while int is 32-bit and
```
EricWF: Does this comment still apply? Where does `int` get into the mix?

EricWF:
Also, tests?

absl/types/any.h
calabrese-google: We do not yet branch for C++14 and C++17, but doing so has been talked about.

absl/types/any.h
calabrese-google: There is no ADL here since `dummy_var` is always of type `char`, so `std::addressof` does not need to be used. The type `Type` does not matter. Side note: We do not have a portable constexpr `std::addressof` as it cannot be implemented without intrinsics. It *might* be possible for us to implement per supported configuration, but that has not yet been investigated and I suspect will only work for some.

absl/types/any.h
calabrese-google: This entire function can simply be removed as it is no longer used.

absl/types/any.h
EricWF: @calabrese-google Ah, sorry my mistake. I just saw the template parameter and reacted.

FYI here is the detection libc++ uses for `__builtin_addressof`. 

```c++
#if !__has_builtin(__builtin_addressof) && _GNUC_VER < 700
# define _LIBCPP_HAS_NO_BUILTIN_ADDRESSOF
#endif```

File: absl/types/any.h [hunk 6]
```diff
@@ -94,23 +94,20 @@ namespace absl {
...
+template <typename Type>
+struct TypeTag {
+  constexpr static char dummy_var = 0;
+};
+
+template <typename Type>
+constexpr char TypeTag<Type>::dummy_var;
+
+// FastTypeId<Type>() evaluates at compile/link-time to a unique pointer for the
+// passed in type. These are meant to be good match for keys into maps or straight
```
calabrese-google: Comment exceeds 80 columns.

absl/types/any.h
calabrese-google: Thanks. We'll likely do this at some point.

PR #50, opened 2017-11-26
Enhancing tests for unescaping strings

brenoguim:
These tests enhance the coverage of CUnescapeInternal method from absl/strings/escaping.cc.

I want to start contributing, so this first pull request serves more as experience.

I took the CUnescapeInternal method and systematically inserted small bugs to check the test coverage. From that experience, I found a few modifications that went undetected and added tests specifically for them.

I'm not sure how valuable this is since this function looks very stable. But I believe the PR would still be a good addition to the test suite in case this is changed for optimizations in the future.

See below a set of changes that would go undetected (if done individually).

```
--- a/absl/strings/escaping.cc
+++ b/absl/strings/escaping.cc
@@ -95,6 +95,7 @@ bool CUnescapeInternal(absl::string_view source, bool leave_nulls_escaped,
   const char* p = source.data();
   const char* end = source.end();
   const char* last_byte = end - 1;
+//const char* last_byte = end - 0;
 
   // Small optimization for case where source = dest and there's no escaping
   while (p == d && p < end && *p != '\\') p++, d++;
@@ -131,9 +132,11 @@ bool CUnescapeInternal(absl::string_view source, bool leave_nulls_escaped,
           const char* octal_start = p;
           unsigned int ch = *p - '0';
           if (p < last_byte && is_octal_digit(p[1])) ch = ch * 8 + *++p - '0';
+//        if (p <= last_byte && is_octal_digit(p[2])) ch = ch * 8 + *++p - '0';
           if (p < last_byte && is_octal_digit(p[1]))
+//        if (p <= last_byte && is_octal_digit(p[1]))
             ch = ch * 8 + *++p - '0';      // now points at last digit
           if (ch > 0xff) {
+//        if (ch >= 0xff) {
             if (error) {
               *error = "Value of \\" +
                        std::string(octal_start, p + 1 - octal_start) +
@@ -155,11 +158,14 @@ bool CUnescapeInternal(absl::string_view source, bool leave_nulls_escaped,
         case 'x':
         case 'X': {
           if (p >= last_byte) {
+//        if (p > last_byte) {
             if (error) *error = "String cannot end with \\x";
             return false;
+//        /*return false;*/
           } else if (!absl::ascii_isxdigit(p[1])) {
             if (error) *error = "\\x cannot be followed by a non-hex digit";
             return false;
+//        /*return false;*/
           }
           unsigned int ch = 0;
           const char* hex_start = p;
@@ -194,6 +200,7 @@ bool CUnescapeInternal(absl::string_view source, bool leave_nulls_escaped,
                        std::string(hex_start, p + 1 - hex_start);
             }
             return false;
+//        /*return false;*/
           }
           for (int i = 0; i < 4; ++i) {
             // Look one char ahead.
@@ -205,6 +212,7 @@ bool CUnescapeInternal(absl::string_view source, bool leave_nulls_escaped,
                          std::string(hex_start, p + 1 - hex_start);
               }
               return false;
+//          /*return false;*/
             }
           }
           if ((rune == 0) && leave_nulls_escaped) {
@@ -227,6 +235,7 @@ bool CUnescapeInternal(absl::string_view source, bool leave_nulls_escaped,
                        std::string(hex_start, p + 1 - hex_start);
             }
             return false;
+//        /*return false;*/
           }
           for (int i = 0; i < 8; ++i) {
             // Look one char ahead.
@@ -235,12 +244,14 @@ bool CUnescapeInternal(absl::string_view source, bool leave_nulls_escaped,
               // is within the Unicode limit, but do advance p.
               uint32_t newrune = (rune << 4) + hex_digit_to_int(*++p);
               if (newrune > 0x10FFFF) {
+//            if (newrune >= 0x10FFFF) {
                 if (error) {
                   *error = "Value of \\" +
                            std::string(hex_start, p + 1 - hex_start) +
                            " exceeds Unicode limit (0x10FFFF)";
                 }
                 return false;
+//            /*return false;*/
               } else {
                 rune = newrune;
               }
@@ -250,6 +261,7 @@ bool CUnescapeInternal(absl::string_view source, bool leave_nulls_escaped,
                          std::string(hex_start, p + 1 - hex_start);
               }
               return false;
+//          /*return false;*/
             }
           }
           if ((rune == 0) && leave_nulls_escaped) {
@@ -265,6 +277,7 @@ bool CUnescapeInternal(absl::string_view source, bool leave_nulls_escaped,
         default: {
           if (error) *error = std::string("Unknown escape sequence: \\") + *p;
           return false;
+//      /*return false;*/
         }
       }
       p++;                                 // read past letter we escaped


```

### DIFF
--- a/absl/strings/escaping_test.cc
+++ b/absl/strings/escaping_test.cc
@@ -281,6 +281,98 @@ TEST_F(CUnescapeTest, UnescapesMultipleUnicodeNulls) {
                    "\0", 5), result_string_);
 }
 
+TEST(CUnescape, NonNullTerminatedOctal1) {
+  absl::string_view reference("\\111", 2);
+
+  std::string unescaped_str;
+  EXPECT_TRUE(absl::CUnescape(reference, &unescaped_str));
+
+  EXPECT_EQ(unescaped_str, "\1");
+}
+
+TEST(CUnescape, NonNullTerminatedOctal2) {
+  absl::string_view reference("\\1z1", 3);
+
+  std::string unescaped_str;
+  EXPECT_TRUE(absl::CUnescape(reference, &unescaped_str));
+
+  EXPECT_EQ(unescaped_str, "\1z");
+}
+
+TEST(CUnescape, NonNullTerminatedHex) {
+  absl::string_view reference("\\x11", 3);
+
+  std::string unescaped_str;
+  EXPECT_TRUE(absl::CUnescape(reference, &unescaped_str));
+
+  EXPECT_EQ(unescaped_str, "\x1");
+}
+
+TEST(CUnescape, NonNullTerminatedInvalidHex) {
+  absl::string_view reference("\\x1", 2);
+
+  std::string unescaped_str;
+  EXPECT_FALSE(absl::CUnescape(reference, &unescaped_str));
+}
+
+TEST(CUnescape, NonNullTerminatedInvalidUtf1) {
+  absl::string_view reference("\\u11111", 5);
+
+  std::string unescaped_str;
+  EXPECT_FALSE(absl::CUnescape(reference, &unescaped_str));
+}
+
+TEST(CUnescape, NonNullTerminatedInvalidUtf2) {
+  absl::string_view reference("\\u1z11", 6);
+
+  std::string unescaped_str;
+  EXPECT_FALSE(absl::CUnescape(reference, &unescaped_str));
+}
+
+TEST(CUnescape, NonNullTerminatedInvalidUtf3) {
+  absl::string_view reference("\\U0000000000000", 9);
+
+  std::string unescaped_str;
+  EXPECT_FALSE(absl::CUnescape(reference, &unescaped_str));
+}
+
+TEST(CUnescape, InvalidHex) {
+  std::string unescaped_str;
+  EXPECT_FALSE(absl::CUnescape("\\xz", &unescaped_str));
+}
+
+TEST(CUnescape, InvalidUtf) {
+  std::string unescaped_str;
+  EXPECT_FALSE(absl::CUnescape("\\U1z111111", &unescaped_str));
+}
+
+TEST(CUnescape, InvalidEscaping) {
+  std::string unescaped_str;
+  EXPECT_FALSE(absl::CUnescape("\\z", &unescaped_str));
+}
+
+TEST(CUnescape, OctalLimits) {
+  std::string reference = "\\377";
+  std::string expected = "\377";
+
+  std::string unescaped_str;
+  EXPECT_TRUE(absl::CUnescape(reference, &unescaped_str));
+  EXPECT_EQ(unescaped_str, expected);
+
+  EXPECT_FALSE(absl::CUnescape("\\400", &unescaped_str));
+}
+
+TEST(CUnescape, UtfULimits) {
+  std::string reference = "\\U0010FFFF";
+  std::string expected = "\U0010FFFF";
+
+  std::string unescaped_str;
+  EXPECT_TRUE(absl::CUnescape(reference, &unescaped_str));
+  EXPECT_EQ(unescaped_str, expected);
+
+  EXPECT_FALSE(absl::CUnescape("\\U00110000", &unescaped_str));
+}
+
 static struct {
   absl::string_view plaintext;
   absl::string_view cyphertext;
@@ -636,101 +728,6 @@ TEST(HexAndBack, HexStringToBytes_and_BytesToHexString) {
 
   std::string hex_result = absl::BytesToHexString(bytes_expected);
   EXPECT_EQ(hex_only_lower, hex_result);
-}
-
-TEST(NonNullTerminatedStrings, OctalConversion) {
-  absl::string_view reference("\\111", 2);
-  std::string expected = "\1";
-
-  std::string unescaped_str;
-  EXPECT_TRUE(absl::CUnescape(reference, &unescaped_str));
-
-  EXPECT_EQ(unescaped_str, expected);
-}
-
-TEST(NonNullTerminatedStrings, InvalidHexConversion) {
-  absl::string_view reference("\\x1", 2);
-
-  std::string unescaped_str;
-  EXPECT_FALSE(absl::CUnescape(reference, &unescaped_str));
-}
-
-TEST(NonNullTerminatedStrings, HexConversion) {
-  absl::string_view reference("\\x11", 3);
-  std::string expected = "\x1";
-
-  std::string unescaped_str;
-  EXPECT_TRUE(absl::CUnescape(reference, &unescaped_str));
-  EXPECT_EQ(unescaped_str, expected);
-}
-
-TEST(NonNullTerminatedStrings, UtfConversion_u) {
-  absl::string_view invalid("\\u11111", 5);
-
-  std::string unescaped_str;
-  EXPECT_FALSE(absl::CUnescape(invalid, &unescaped_str));
-}
-
-TEST(NonNullTerminatedStrings, UtfConversion_U) {
-  absl::string_view invalid("\\U0000000000000", 9);
-
-  std::string unescaped_str;
-  EXPECT_FALSE(absl::CUnescape(invalid, &unescaped_str));
-}
-
-TEST(NonNullTerminatedStrings, InvalidOctal) {
-  absl::string_view reference("\\1z1", 3);
-
-  std::string unescaped_str;
-  EXPECT_TRUE(absl::CUnescape(reference, &unescaped_str));
-
-  EXPECT_EQ(unescaped_str, "\1z");
-}
-
-TEST(NonNullTerminatedStrings, InvalidHex) {
-  absl::string_view reference("\\xz", 3);
-
-  std::string unescaped_str;
-  EXPECT_FALSE(absl::CUnescape(reference, &unescaped_str));
-}
-
-TEST(NonNullTerminatedStrings, InvalidUtf_u_RightSizeButInvalidChar) {
-  absl::string_view reference("\\u1z11", 6);
-
-  std::string unescaped_str;
-  EXPECT_FALSE(absl::CUnescape(reference, &unescaped_str));
-}
-
-TEST(Limits, Octal) {
-  std::string reference = "\\377";
-  std::string expected = "\377";
-
-  std::string unescaped_str;
-  EXPECT_TRUE(absl::CUnescape(reference, &unescaped_str));
-  EXPECT_EQ(unescaped_str, expected);
-
-  EXPECT_FALSE(absl::CUnescape("\\400", &unescaped_str));
-}
-
-TEST(Limits, Utf_U) {
-  std::string reference = "\\U0010FFFF";
-  std::string expected = "\U0010FFFF";
-
-  std::string unescaped_str;
-  EXPECT_TRUE(absl::CUnescape(reference, &unescaped_str));
-  EXPECT_EQ(unescaped_str, expected);
-
-  EXPECT_FALSE(absl::CUnescape("\\U00110000", &unescaped_str));
-}
-
-TEST(Invalids, Utf_U_RightSizeButInvalidChar) {
-  std::string unescaped_str;
-  EXPECT_FALSE(absl::CUnescape("\\U1z111111", &unescaped_str));
-}
-
-TEST(Invalids, InvalidEscaping) {
-  std::string unescaped_str;
-  EXPECT_FALSE(absl::CUnescape("\\z", &unescaped_str));
 }
 
 }  // namespace
--- a/absl/strings/escaping_test.cc
+++ b/absl/strings/escaping_test.cc
@@ -638,4 +638,108 @@ TEST(HexAndBack, HexStringToBytes_and_BytesToHexString) {
   EXPECT_EQ(hex_only_lower, hex_result);
 }
 
+TEST(OffByOneErrors, OctalConversion) {
+  const char creference[] = "\\111";
+  absl::string_view reference(creference, 2);
+  std::string expected = "\1";
+
+  std::string unescaped_str;
+  EXPECT_TRUE(absl::CUnescape(reference, &unescaped_str));
+
+  EXPECT_EQ(unescaped_str, expected);
+}
+
+TEST(OffByOneErrors, InvalidHexConversion) {
+  const char creference[] = "\\x1";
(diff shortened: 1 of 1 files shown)


absl/strings/escaping_test.cc
```diff
@@ -635,4 +635,108 @@ TEST(HexAndBack, HexStringToBytes_and_BytesToHexString) {
   EXPECT_EQ(hex_only_lower, hex_result);
 }
 
+TEST(OffByOneErrors, OctalConversion) {
+  const char creference[] = "\\111";
+  absl::string_view reference(creference, 2);
```
brenoguim: Perhaps it's best to just write:

```
absl::string_view reference("\\111", 2);
```

absl/strings/escaping_test.cc
```diff
@@ -635,4 +635,108 @@ TEST(HexAndBack, HexStringToBytes_and_BytesToHexString) {
   EXPECT_EQ(hex_only_lower, hex_result);
 }
 
+TEST(OffByOneErrors, OctalConversion) {
```
brenoguim: Better name suggestion other than "OffByOneErrors"?

mbxx:
Thanks for the pull request and glad to hear you want to start contributing!  This is definitely a stable interface, but adding tests is still a good thing.

The tests all seem reasonable, so I think I'm just going to have some comments about naming and organization.

absl/strings/escaping_test.cc
```diff
@@ -635,4 +635,99 @@ TEST(HexAndBack, HexStringToBytes_and_BytesToHexString) {
   EXPECT_EQ(hex_only_lower, hex_result);
 }
 
+TEST(NonNullTerminatedStrings, OctalConversion) {
```
mbxx: These tests all test absl::CUnescape, so let's make them look like the other absl::CUnescape tests.  Let's put them up with the other CUnescape tests and have them start with `TEST(CUnescape`, like this:

```
TEST(CUnescape, NonNullTerminatedOctalConversion) {
  ...
```

absl/strings/escaping_test.cc
```diff
@@ -635,4 +635,99 @@ TEST(HexAndBack, HexStringToBytes_and_BytesToHexString) {
...
+}
+
+TEST(NonNullTerminatedStrings, InvalidHex) {
+  absl::string_view reference("\\xz", 3);
+
+  std::string unescaped_str;
+  EXPECT_FALSE(absl::CUnescape(reference, &unescaped_str));
+}
+
+TEST(NonNullTerminatedStrings, InvalidUtf_u_RightSizeButInvalidChar) {
```
mbxx: Google Test asks that the names of tests not contain underscores: https://github.com/google/googletest/blob/master/googletest/docs/FAQ.md#why-should-not-test-case-names-and-test-names-contain-underscore

absl/strings/escaping_test.cc
```diff
@@ -635,4 +635,99 @@ TEST(HexAndBack, HexStringToBytes_and_BytesToHexString) {
...
+  absl::string_view reference("\\1z1", 3);
+
+  std::string unescaped_str;
+  EXPECT_TRUE(absl::CUnescape(reference, &unescaped_str));
+
+  EXPECT_EQ(unescaped_str, "\1z");
+}
+
+TEST(NonNullTerminatedStrings, InvalidHex) {
+  absl::string_view reference("\\xz", 3);
```
mbxx: It's not obvious to me why you're specifying the size of the `string_view` here.  Wouldn't `std::string reference = "\\xz";` have the same behavior?

mbxx (changes requested):
Thanks you for the additional tests!  I think this should all be fine after a few minor stylistic changes.

absl/strings/escaping_test.cc
brenoguim: You are right. I thought I needed to adjust the size to get the right behavior, and ended up with the full string without realizing it :)

By the way this test is designed to catch this error:

```
        case 'x':                                                                                                       
        case 'X': {                                                                                                     
          if (p >= last_byte) {                                                                                         
            if (error) *error = "String cannot end with \\x";                                                           
            return false;                                                                                               
          } else if (!absl::ascii_isxdigit(p[1])) {                                                                     
            if (error) *error = "\\x cannot be followed by a non-hex digit";                                            
            //return false;      <---- This is the only test failing if this return statement is missing                                                                                                                                                  
          }    
```

absl/strings/escaping_test.cc
brenoguim: Makes sense, will do :)

absl/strings/escaping_test.cc
brenoguim: Ouch, rookie mistake. Thanks!

brenoguim: It's hard to make that look good without underscores :(

I also noticed some other tests have underscores. That could be a low hanging issue for new comers.

absl/strings/escaping_test.cc
```diff
@@ -278,6 +278,98 @@ TEST_F(CUnescapeTest, UnescapesMultipleUnicodeNulls) {
                    "\0", 5), result_string_);
 }
 
+TEST(CUnescape, NonNullTerminatedOctal1) {
+  absl::string_view reference("\\111", 2);
+
+  std::string unescaped_str;
```
mbxx: I'm probably being too pedantic, but let's make these tests more consistent with the tests above.  If we don't, someone is likely to wonder why they're different when there isn't really a reason.

The CUnescape test framework has a `result_string_` member variable.  Let's reuse that here.  We'll have to change `TEST` to `TEST_F` to make that work.

We can also rename `reference` to `original_string` to match the above tests.

absl/strings/escaping_test.cc
```diff
@@ -278,6 +278,98 @@ TEST_F(CUnescapeTest, UnescapesMultipleUnicodeNulls) {
...
+  EXPECT_TRUE(absl::CUnescape(reference, &unescaped_str));
+
+  EXPECT_EQ(unescaped_str, "\1");
+}
+
+TEST(CUnescape, NonNullTerminatedOctal2) {
+  absl::string_view reference("\\1z1", 3);
```
mbxx: Does this follow a different code path than `absl::string_view reference("\\1z");` would?  If not, we should choose the simpler test input.  We could then rename it `InvalidOctal` and put it with the other similar tests below.

absl/strings/escaping_test.cc
```diff
@@ -278,6 +278,98 @@ TEST_F(CUnescapeTest, UnescapesMultipleUnicodeNulls) {
...
+
+TEST(CUnescape, NonNullTerminatedInvalidUtf1) {
+  absl::string_view reference("\\u11111", 5);
+
+  std::string unescaped_str;
+  EXPECT_FALSE(absl::CUnescape(reference, &unescaped_str));
+}
+
+TEST(CUnescape, NonNullTerminatedInvalidUtf2) {
+  absl::string_view reference("\\u1z11", 6);
```
mbxx: You're using the entire `string_view` here.  How about dropping the explicit length and changing the name of the test from `NonNullTerminatedInvalidUtf2` to `InvalidShortUnicode`?

It would then most naturally go with the other `Invalid...` tests below.

absl/strings/escaping_test.cc
```diff
@@ -278,6 +278,98 @@ TEST_F(CUnescapeTest, UnescapesMultipleUnicodeNulls) {
...
+  std::string unescaped_str;
+  EXPECT_FALSE(absl::CUnescape(reference, &unescaped_str));
+}
+
+TEST(CUnescape, InvalidHex) {
+  std::string unescaped_str;
+  EXPECT_FALSE(absl::CUnescape("\\xz", &unescaped_str));
+}
+
+TEST(CUnescape, InvalidUtf) {
```
mbxx: We could name this `InvalidLongUnicode` out of symmetry with `InvalidShortUnicode`.

absl/strings/escaping_test.cc
```diff
@@ -278,6 +278,98 @@ TEST_F(CUnescapeTest, UnescapesMultipleUnicodeNulls) {
...
+  std::string unescaped_str;
+  EXPECT_FALSE(absl::CUnescape("\\U1z111111", &unescaped_str));
+}
+
+TEST(CUnescape, InvalidEscaping) {
```
mbxx: `InvalidEscaping` is a bit vague.  Maybe call this `InvalidEscapedCharacter`?

absl/strings/escaping_test.cc
```diff
@@ -278,6 +278,98 @@ TEST_F(CUnescapeTest, UnescapesMultipleUnicodeNulls) {
...
+  std::string expected = "\377";
+
+  std::string unescaped_str;
+  EXPECT_TRUE(absl::CUnescape(reference, &unescaped_str));
+  EXPECT_EQ(unescaped_str, expected);
+
+  EXPECT_FALSE(absl::CUnescape("\\400", &unescaped_str));
+}
+
+TEST(CUnescape, UtfULimits) {
```
mbxx: This is really about Unicode rather than UTF, so how about calling this `UnicodeLimits`?

absl/strings/escaping_test.cc
brenoguim: Cool, thanks for the tip. Don't worry about being pedantic, that's how good code is done. :)

I'll fix that.

absl/strings/escaping_test.cc
```diff
@@ -278,6 +278,98 @@ TEST_F(CUnescapeTest, UnescapesMultipleUnicodeNulls) {
...
+TEST(CUnescape, NonNullTerminatedOctal2) {
+  absl::string_view reference("\\1z1", 3);
+
+  std::string unescaped_str;
+  EXPECT_TRUE(absl::CUnescape(reference, &unescaped_str));
+
+  EXPECT_EQ(unescaped_str, "\1z");
+}
+
+TEST(CUnescape, NonNullTerminatedHex) {
```
mbxx: What do you think about naming this test `HexUnescapingDoesNotOverrunBuffer`?  Naming is hard and subjective, but that name describes what the test is trying to do a little better to me.

absl/strings/escaping_test.cc
```diff
@@ -278,6 +278,98 @@ TEST_F(CUnescapeTest, UnescapesMultipleUnicodeNulls) {
...
+}
+
+TEST(CUnescape, NonNullTerminatedInvalidHex) {
+  absl::string_view reference("\\x1", 2);
+
+  std::string unescaped_str;
+  EXPECT_FALSE(absl::CUnescape(reference, &unescaped_str));
+}
+
+TEST(CUnescape, NonNullTerminatedInvalidUtf1) {
```
mbxx: I prefer test names to describe how they differ rather than having numbers at the end where possible.  How about calling this `ShortUnicodeUnescapingDoesNotOverrunBuffer`?  (I hope that isn't so long it causes a line wrap.)

mbxx (changes requested):
I worry that I'm being really nit-picky here, but I just like clear names and think it's worth spending a little extra time to be clear.  Thanks for bearing with me :)

Also, sorry about not sending all of my comments in one batch--I'm still learning the GitHut code review interface.

absl/strings/escaping_test.cc
mbxx: Definitely not a rookie mistake, as evidenced by the other code that violates it :)

Sure, this could be a fine thing for someone wanting to get their feet wet.

brenoguim:
Sorry for the delay here. I'll need another week or so. It's software release week back at the office. :)

mbxx:
No worries, take your time :)

mbxx:
Hi @brenoguim, just checking in.  Are you still interested in doing the clean up on this PR?  If not, I'd be happy to take it over for you one way or another.  Totally up to you, and no rush if this is still something you'd like to do :)

brenoguim:
Oh gosh! I forgot about this. End of the year is a busy time!

I will continue on this this week. I definitely have interest in finishing what i started :)

mbxx:
Cool :)  Feel free to reach out if there's anything I can do to help.

absl/strings/escaping_test.cc
brenoguim: They kind of follow the same code path, but are designed to catch different "off by one" mistakes.

The relevant code piece is this one:
escaping.cc
```
 144           if (p < last_byte && is_octal_digit(p[1])) ch = ch * 8 + *++p - '0';                                          
 145           if (p < last_byte && is_octal_digit(p[1]))                                                                                                                        
 146             ch = ch * 8 + *++p - '0';      // now points at last digit    
```

While the first test catches things like:

```
 -144           if (p < last_byte && is_octal_digit(p[1])) ch = ch * 8 + *++p - '0';
 +144           if (p <= last_byte && is_octal_digit(p[1])) ch = ch * 8 + *++p - '0';                                          
 145           if (p < last_byte && is_octal_digit(p[1]))                                                                                                                        
 146             ch = ch * 8 + *++p - '0';      // now points at last digit    
```
This second test is the only one that would fail on the following off-by-one error:
```
 -144           if (p < last_byte && is_octal_digit(p[1])) ch = ch * 8 + *++p - '0';
 +144           if (p < last_byte && is_octal_digit(p[2])) ch = ch * 8 + *++p - '0';                                          
 145           if (p < last_byte && is_octal_digit(p[1]))                                                                                                                        
 146             ch = ch * 8 + *++p - '0';      // now points at last digit    
```

mbxx: Oh, I see what you're after, thanks.  You can actually trigger that second case without specifying the size of the string_view:

```
TEST_F(CUnescapeTest, UnescapeOctal) {
  std::string original_string = "\\1z3";
  EXPECT_TRUE(absl::CUnescape(original_string, &result_string_));
  EXPECT_EQ("\1z3", result_string_);
}
```

Let's do that.  It keeps the test minimal.  By specifying the size, I assumed that's what was being tested.  Then we can rename `NonNullTerminatedOctal2` to `Octal` and rename `NonNullTerminatedOctal1` to `NonNullTerminatedOctal`, and then everything looks good to me.

Thanks for bearing with me on this, and sorry about this slow review.

ahedberg:
Hi there--any plans to resume this pull request? If not, I'm inclined to close it by the end of the week, as it's been inactive for quite some time. We're happy to provide help if needed.

brenoguim:
Hi!
Sorry for not finishing the work here. I won't be take this further :(
Please go ahead and close it.

Thanks!

Issue #34, opened 2017-10-15
Don't #include <windows.h>


Remotion:
Currently it appears to be impossible to compile Abseil using VS 2017 without errors.

Adding something like this resolve this problem with windows.h includes.

    # ifndef WIN32_LEAN_AND_MEAN
    #   define WIN32_LEAN_AND_MEAN
    # endif
    # ifndef VC_EXTRALEAN
    #   define VC_EXTRALEAN
    # endif
    # ifndef NOMINMAX
    #   define NOMINMAX
    # endif
    #include <windows.h>

zhangxy988:
@Remotion "Currently it appears to be impossible to compile Abseil using VS 2017 without errors." Could you please provide details on what specific compiling errors you are getting and how to reproduce the errors?
I just tried running all unit tests with VS 2017 15.3 and it worked fine (except for something related to expression SFINAE with container_test which I don't think is related to windows.h).

Remotion:
I did not use Bazel to build, instate I more or less simple dropped files into existing VS2017.4 project and just tried to compile it.

There ware a lot of compiler errors, for example with 
    #define min
that collided with 
    std::min

I see now here [WIN32_LEAN_AND_MEAN](https://github.com/abseil/abseil-cpp/blob/master/absl/copts.bzl#L105) so this probably why there is no problems if build using bazel.

zhangxy988:
So let me rephrase the issue:
Abseil #includes <windows.h> from some files (e.g. clock.cc, waiter.h) where it's conflicting with usage of `std::min` because <windows.h> #defines `min` as a macro.
I think the general approach for this kind of issue is to configure the build to predefine `WIN32_LEAN_AND_MEAN`, just like what we did in our bazel buid file.
Defining `WIN32_LEAN_AND_MEAN` in code (before each `#include <windows.h>`)  doesn't solve this issue in all cases because it depends on the header inclusion order. For example, <windows.h> could be #included by another non-Abseil header file before any Abseil file is #included (which might not have `WIN32_LEAN_AND_MEAN` defined).
So I would suggest fixing this issue at the build system level rather than at source code level.

JonathanDCohen:
Friendly ping here :)  @Remotion are you still having trouble compiling on Windows?

zivsha:
FYI, just cloned abseil to my `cmake` project (with `googletest` and `cctz` with Sarcasm's PR for CMake support) and only issue I'm having is with the usual `std::max`:
[...]

PR #55, opened 2017-11-28
MSVC fixes

rongjiecomputer:
With this PR, abseil can be built with CMake on Windows successfully.

- `NOMINMAX` is required to prevent `windows.h` from defining `min` and `max` macros, which mess up with things like `std::numeric_limits<T>::max()` as seen in #34 (Also added to Bazel).
- `_CRT_SECURE_NO_WARNINGS` is required to stop MSVC from complaining about the usage of insecure `getenv` over secure `getenv_s`. (Also added to Bazel).
- Deprecated `_sleep` is replaced with Win32 `Sleep` function.
- VS 2015 and above does not understand `-std=c++11` or `/std:c++11`, default is `/std:c++14`.
- `CMAKE_CXX_FIX_MSVC` is removed (not initialized by anyone).

The following code is needed in both abseil side and user side because `add_subdirectory` will not export `CMAKE_CXX_FLAGS` outside the directory.

```cmake
add_definitions(/DNOMINMAX /DWIN32_LEAN_AND_MEAN=1 /D_CRT_SECURE_NO_WARNINGS)  
add_compile_options(/W3 /WX /wd4005 /wd4068 /wd4244 /wd4267 /wd4800)
```

### DIFF
--- a/CMake/README.md
+++ b/CMake/README.md
@@ -43,6 +43,11 @@
 
       set(CMAKE_CXX_FLAGS "-std=c++11 -stdlib=libc++ ${CMAKE_CXX_FLAGS}")
 
+      if (MSVC)
+        add_compile_options(/wd4005 /wd4068 /wd4244 /wd4267 /wd4800)
+        add_definitions(/DNOMINMAX /DWIN32_LEAN_AND_MEAN=1 /D_CRT_SECURE_NO_WARNINGS)
+      endif()
+
       add_subdirectory(googletest)
       add_subdirectory(cctz)
       add_subdirectory(abseil-cpp)
@@ -51,7 +56,7 @@
       target_link_libraries(my_exe absl::base absl::synchronization absl::strings)
 
 
-You will need to create your own CMake files for cctz until https://github.com/google/cctz/pull/54 lands.  As of this writing, that pull request requires -DBUILD_TESTING=OFF as it doesn't correctly export cctz's dependency on Google Benchmark.
+As of this writing, that pull request requires -DBUILD_TESTING=OFF as it doesn't correctly export cctz's dependency on Google Benchmark.
 
     You will find here a non exhaustive list of absl public targets
 
--- a/CMake/README.md
+++ b/CMake/README.md
@@ -44,6 +44,11 @@
       set(CMAKE_CXX_FLAGS "-std=c++11 -stdlib=libc++ ${CMAKE_CXX_FLAGS}")
 
       if (MSVC)
+        # /wd4005  macro-redefinition
+        # /wd4068  unknown pragma
+        # /wd4244  conversion from 'type1' to 'type2'
+        # /wd4267  conversion from 'size_t' to 'type2'
+        # /wd4800  force value to bool 'true' or 'false' (performance warning)
         add_compile_options(/wd4005 /wd4068 /wd4244 /wd4267 /wd4800)
         add_definitions(/DNOMINMAX /DWIN32_LEAN_AND_MEAN=1 /D_CRT_SECURE_NO_WARNINGS)
       endif()
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -27,8 +27,13 @@ include(AbseilHelpers)
 
 # config options
 if (MSVC)
-  add_definitions(/DNOMINMAX /DWIN32_LEAN_AND_MEAN=1 /D_CRT_SECURE_NO_WARNINGS)
+  # /wd4005  macro-redefinition
+  # /wd4068  unknown pragma
+  # /wd4244  conversion from 'type1' to 'type2'
+  # /wd4267  conversion from 'size_t' to 'type2'
+  # /wd4800  force value to bool 'true' or 'false' (performance warning)
   add_compile_options(/W3 /WX /wd4005 /wd4068 /wd4244 /wd4267 /wd4800)
+  add_definitions(/DNOMINMAX /DWIN32_LEAN_AND_MEAN=1 /D_CRT_SECURE_NO_WARNINGS)
 else()
   set(ABSL_STD_CXX_FLAG "-std=c++11" CACHE STRING "c++ std flag (default: c++11)")
 endif()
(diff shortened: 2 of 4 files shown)


CMake/README.md
```diff
@@ -43,6 +43,11 @@
 
       set(CMAKE_CXX_FLAGS "-std=c++11 -stdlib=libc++ ${CMAKE_CXX_FLAGS}")
 
+      if (MSVC)
+        add_compile_options(/wd4005 /wd4068 /wd4244 /wd4267 /wd4800)
```
JonathanDCohen: Please add a comment explaining these compile options.  Right now they just look like magic numbers.

CMakeLists.txt
```diff
@@ -26,7 +26,12 @@ include(AbseilHelpers)
 
 
 # config options
-set(ABSL_STD_CXX_FLAG "-std=c++11" CACHE STRING "c++ std flag (default: c++11)")
+if (MSVC)
+  add_definitions(/DNOMINMAX /DWIN32_LEAN_AND_MEAN=1 /D_CRT_SECURE_NO_WARNINGS)
+  add_compile_options(/W3 /WX /wd4005 /wd4068 /wd4244 /wd4267 /wd4800)
```
JonathanDCohen: Same here

CMakeLists.txt
```diff
@@ -26,7 +26,12 @@ include(AbseilHelpers)
 
 
 # config options
-set(ABSL_STD_CXX_FLAG "-std=c++11" CACHE STRING "c++ std flag (default: c++11)")
+if (MSVC)
```
JonathanDCohen, line 29: I'm assuming that one of these options sets the default c++ standard to C++11? Abseil doesn't support earlier versions of the standard.

absl/copts.bzl
```diff
@@ -103,7 +103,9 @@ MSVC_FLAGS = [
     "/wd4244",  # conversion from 'type1' to 'type2', possible loss of data
     "/wd4267",  # conversion from 'size_t' to 'type', possible loss of data
     "/wd4800",  # forcing value to bool 'true' or 'false' (performance warning)
+    "/DNOMINMAX",  # Don't define min and max macros (windows.h)
     "/DWIN32_LEAN_AND_MEAN",  # Don't bloat namespace with incompatible winsock versions.
+    "/D_CRT_SECURE_NO_WARNINGS",  # Don't warn about usage of insecure C functions
```
JonathanDCohen, line 108: I'm uncomfortable about this.  I'd rather mark each `getenv` call with a #warning pragma so we can be explicit about calling unsafe functions.

JonathanDCohen (changes requested):

absl/copts.bzl
rongjiecomputer: I am not sure if this is practical. There are just too many C functions marked as insecure (`fopen`, `strcpy`, `getenv`, `rand`, `memcpy`, `printf` etc.). Using #warning pragma will be too verbose.

See the full list at https://msdn.microsoft.com/en-us/library/wd3wzwts.aspx

CMakeLists.txt
rongjiecomputer: VS 2015 and above only has C++14 (default) and C++17 (for VS 2017). They do not have C++11 mode.

See https://msdn.microsoft.com/en-us/library/mt490614.aspx

CMake/README.md
```diff
@@ -43,6 +43,16 @@
 
       set(CMAKE_CXX_FLAGS "-std=c++11 -stdlib=libc++ ${CMAKE_CXX_FLAGS}")
 
+      if (MSVC)
+        # /wd4005  macro-redefinition
+        # /wd4068  unknown pragma
+        # /wd4244  conversion from 'type1' to 'type2'
+        # /wd4267  conversion from 'size_t' to 'type2'
+        # /wd4800  force value to bool 'true' or 'false' (performance warning)
+        add_compile_options(/wd4005 /wd4068 /wd4244 /wd4267 /wd4800)
```
rongjiecomputer, line 52: Consider these warning flags as temporary fix just to get Abseil to build on Windows. We should eventually fix these warnings or at least reduce the number of occurrence until we can just use #pragma explicitly.

rongjiecomputer:
With google/googletest#1337, Abseil can be built with `clang-cl` on Windows as well.

absl/copts.bzl
JonathanDCohen: Okay, that's fine for now.  I wonder if it's practical to have shims which use teh safe versions provided by msvc and C11 when they're available and fallback to the unsafe versions otherwise.

CMake/README.md
JonathanDCohen: Thanks. As a nit, please document that these are temporary

JonathanDCohen (approved):

absl/copts.bzl
rongjiecomputer: `-D_CRT_SECURE_CPP_OVERLOAD_STANDARD_NAMES=1` can be used to tell MSVC to automatically convert unsafe CRT function calls to safe CRT functions (only works for certain cases).

The so-called "safe" CRT functions are just "unsafe/normal" CRT functions with extra bound-checks, so they tend to have slightly _worse performance_. Abseil code probably already does enough bound-checks, so explicit "safe" CRT functions aren't useful.

`D_CRT_SECURE_NO_WARNINGS` is a very common flag used by many cross-platform projects, I won't worry too much about it. See https://github.com/llvm-mirror/llvm/blob/master/cmake/modules/HandleLLVMOptions.cmake#L336

See https://msdn.microsoft.com/en-us/library/ms175759.aspx

CMake/README.md
rongjiecomputer: On second thought, I think trying to silence MSVC about these warnings by modifying the code might be impossible. MSVC is more noisy than GCC and Clang. Some projects have even longer list (see LLVM and Chromium for instance).

zhangxy988:
Thanks for the PR. Sorry for the delay.
FYI, I'll need to run some test before I can merge this.

PR #57, opened 2017-12-04
FreeBSD compability

blodan:
Hi!

I was a user of Tensorflow on FreeBSD until after 1.2.0 because they started using abseil (which doesnt seem to be FreeBSD compatible).

I fixed alot of errors with this PR but am now stuck with this error when running the tests:
```
ERROR: /usr/home/daniely/abseil-cpp/absl/base/BUILD.bazel:374:1: Linking of rule '//absl/base:low_level_alloc_test' failed (Exit 1)
bazel-out/local_freebsd-fastbuild/bin/absl/base/_objs/low_level_alloc_test/absl/base/internal/low_level_alloc_test.pic.o: In function `std::__1::unordered_map<int, absl::base_internal::(anonymous namespace)::BlockDesc, std::__1::hash<int>, std::__1::equal_to<int>, std::__1::allocator<std::__1::pair<int const, absl::base_internal::(anonymous namespace)::BlockDesc> > >::operator[](int const&)':
absl/base/internal/low_level_alloc_test.cc:(.text+0x1ff4): undefined reference to `ceilf'
bazel-out/local_freebsd-fastbuild/bin/absl/base/_objs/low_level_alloc_test/absl/base/internal/low_level_alloc_test.pic.o: In function `std::__1::__hash_table<std::__1::__hash_value_type<int, absl::base_internal::(anonymous namespace)::BlockDesc>, std::__1::__unordered_map_hasher<int, std::__1::__hash_value_type<int, absl::base_internal::(anonymous namespace)::BlockDesc>, std::__1::hash<int>, true>, std::__1::__unordered_map_equal<int, std::__1::__hash_value_type<int, absl::base_internal::(anonymous namespace)::BlockDesc>, std::__1::equal_to<int>, true>, std::__1::allocator<std::__1::__hash_value_type<int, absl::base_internal::(anonymous namespace)::BlockDesc> > >::rehash(unsigned long)':
absl/base/internal/low_level_alloc_test.cc:(.text+0x3aa9): undefined reference to `ceilf'
absl/base/internal/low_level_alloc_test.cc:(.text+0x3bc6): undefined reference to `ceilf'
```

I can't seem to find a solution to the above problem, would you guys have any pointers to get me further? (This is when running bazel test absl/...:all )

The segfault in Tensorflow on FreeBSD that lead me to do this patch was this backtrace:
```
#0  0x000000080e12c20c in absl::base_internal::NumCPUs () from /usr/local/lib/python3.6/site-packages/tensorflow/python/../libtensorflow_framework.so
(gdb) bt
#0  0x000000080e12c20c in absl::base_internal::NumCPUs () from /usr/local/lib/python3.6/site-packages/tensorflow/python/../libtensorflow_framework.so
#1  0x000000080e12c0a9 in _GLOBAL__sub_I_spinlock.cc () from /usr/local/lib/python3.6/site-packages/tensorflow/python/../libtensorflow_framework.so
#2  0x000000080e1cedd2 in __do_global_ctors_aux () from /usr/local/lib/python3.6/site-packages/tensorflow/python/../libtensorflow_framework.so
#3  0x000000080db578de in _init () from /usr/local/lib/python3.6/site-packages/tensorflow/python/../libtensorflow_framework.so
```


absl/BUILD.bazel
```diff
@@ -50,3 +50,11 @@ config_setting(
     },
     visibility = [":__subpackages__"],
 )
+
+config_setting(
```
derekmauro: It doesn't look like this is actually used anywhere.  Is it actually needed?

absl/base/config.h
```diff
@@ -254,7 +254,7 @@
 // POSIX.1-2001.
 #ifdef ABSL_HAVE_MMAP
 #error ABSL_HAVE_MMAP cannot be directly set
-#elif defined(__linux__) || defined(__APPLE__) || defined(__ros__) || \
+#elif defined(__linux__) || defined(__APPLE__) || defined(__FreeBSD__) || defined(__ros__) || \
```
derekmauro: https://github.com/abseil/abseil-cpp/blob/master/CONTRIBUTING.md#coding-style

We have a style-guide that code needs to adhere to. One of the rules we have is that C++ source files have an 80 char limit per line.  Can you make sure your code adheres to this?

You might be able to fix this automatically by running "git clang-format".

absl/base/config.h
```diff
@@ -265,7 +265,7 @@
 // functions as defined in POSIX.1-2001.
 #ifdef ABSL_HAVE_PTHREAD_GETSCHEDPARAM
 #error ABSL_HAVE_PTHREAD_GETSCHEDPARAM cannot be directly set
-#elif defined(__linux__) || defined(__APPLE__) || defined(__ros__)
+#elif defined(__linux__) || defined(__APPLE__) || defined(__FreeBSD__) || defined(__ros__)
```
derekmauro: Same comment about 80 chars.

absl/debugging/internal/stacktrace_config.h
```diff
@@ -24,7 +24,7 @@
 // First, test platforms which only support a stub.
 #if ABSL_STACKTRACE_INL_HEADER
 #error ABSL_STACKTRACE_INL_HEADER cannot be directly set
-#elif defined(__native_client__) || defined(__APPLE__) || \
+#elif defined(__native_client__) || defined(__APPLE__) || defined(__FreeBSD__) || \
```
derekmauro: Same comment about 80 chars.

absl/debugging/internal/stacktrace_powerpc-inl.inc
```diff
@@ -47,7 +47,7 @@ static inline void *StacktracePowerPCGetLR(void **sp) {
   return *(sp+2);
 #elif defined(_CALL_SYSV)
   return *(sp+1);
-#elif defined(__APPLE__) || (defined(__linux__) && defined(__PPC64__))
+#elif defined(__APPLE__) || defined(__FreeBSD__) || (defined(__linux__) && defined(__PPC64__))
```
derekmauro: Same comment about 80 chars.

absl/base/internal/raw_logging.cc
```diff
@@ -34,7 +34,7 @@
 //
 // This preprocessor token is also defined in raw_io.cc.  If you need to copy
 // this, consider moving both to config.h instead.
-#if defined(__linux__) || defined(__APPLE__) || defined(__Fuchsia__)
+#if defined(__linux__) || defined(__APPLE__) || defined(__FreeBSD__) || defined(__Fuchsia__)
```
derekmauro: Same comment about 80 chars.

derekmauro (changes requested):
I *think* the build errors like "undefined reference to `ceilf'" are because you need to explicitly like against the math library.  We have a few unit tests that don't use the Google Test framework.  It seems like we are relying on the Google Test framework to link in the math library (and the pthread library) for the majority of test targets.  I think you will have to add "-lm" to the linkopts section of test targets that don't use Google Test.

https://github.com/abseil/abseil-cpp/blob/master/absl/base/BUILD.bazel#L381 is an example of one of the places where I think you'd have to add "-lm" to the "//conditions:default" list.

Can you try this and let us know if it works?  I don't have access to a FreeBSD machine to test on currently.

absl/BUILD.bazel
blodan: It wasnt used at the time, but now it is used to specify the math link option.

absl/base/config.h
blodan: Fixed

absl/base/config.h
blodan: Fixed

absl/base/internal/raw_logging.cc
blodan: Fixed

absl/debugging/internal/stacktrace_config.h
blodan: Fixed

absl/debugging/internal/stacktrace_powerpc-inl.inc
blodan: Fixed

blodan:
Thanks for the input about the linking option, that solved almost all the issues that still existed.

All tests except one now passes.

The test thats failing is meta/type_traits TestTrivialCopyCtor but I have no clue where to even begin with that one.

The output is as follows 
```
absl/meta/type_traits_test.cc:257: Failure
Value of: (absl::is_trivially_copy_constructible<std::pair<int, char*>>::value)
  Actual: false
Expected: true
absl/meta/type_traits_test.cc:259: Failure
Value of: (absl::is_trivially_copy_constructible<std::pair<int, Trivial>>::value)
  Actual: false
Expected: true
absl/meta/type_traits_test.cc:261: Failure
Value of: (absl::is_trivially_copy_constructible< std::pair<int, TrivialCopyCtor>>::value)
  Actual: false
Expected: true
[  FAILED  ] TypeTraitsTest.TestTrivialCopyCtor (0 ms)
```

absl/base/BUILD.bazel
```diff
@@ -38,6 +38,10 @@ cc_library(
         "internal/spinlock_wait.h",
     ],
     copts = ABSL_DEFAULT_COPTS,
+    linkopts = select({
```
derekmauro: Let's try to avoid this repeated code everywhere.

https://github.com/abseil/abseil-cpp/blob/master/absl/copts.bzl#L118 has a ABSL_DEFAULT_COPTS var.  Can you add an ABSL_DEFAULT_LINKOPTS var to copts.bzl?

I think you can leave out the specific case of "//absl:freebsd".  Just add both "-lm" and "-pthread" to "//conditions:default".  As far as I know all the POSIX build environments supported by Abseil need both, but we are probably somehow implicitly using "-lm" on Linux.  It should be harmless to add them both as defaults and not special case FreeBSD here.

derekmauro (changes requested):

derekmauro:
@JonathanDCohen Do you have any idea what might be going on with type_traits_test (see https://github.com/abseil/abseil-cpp/pull/57#issuecomment-349242090)?

JonathanDCohen:
@derekmauro Let me take a look

@blodan do you know which compiler this is failing on?

absl/base/BUILD.bazel
blodan: I've now changed to a global linkopts var and dropped the freebsd definition.

blodan:
@JonathanDCohen 

```
FreeBSD clang version 4.0.0 (tags/RELEASE_400/final 297347) (based on LLVM 4.0.0)
Target: x86_64-unknown-freebsd11.1
Thread model: posix
InstalledDir: /usr/bin
```

JonathanDCohen:
@blodan 
Since you're using clang, I'm also assuming you're using libc++.  Is this true? 

In that case, it's interesting that the [static assert in absl::is_trivially_copy_constructible](https://github.com/abseil/abseil-cpp/blob/master/absl/meta/type_traits.h#L223) isn't firing, indicating that something about clang's std::is_trivially_copy_constructible may be borked on FreeBSD.

I don't have a FreeBSD machine to test this on unfortunately, so we're going to have to work together on this one.  Could you output the different components of absl::is_trivially_copy_constructible (as in __has_trivial_copy(T), std::is_copy_constructible<T>::value, is_trivially_destructible<T>::value) and the value of std::is_trivially_cop_constructible<T> for std::pair<int, char*>?

JonathanDCohen:
@blodan Matt Calabrese (@calabrese-google) is pushing an internal cl to remove the dependency on std::pair to see if that resolves the issue.  Please try running our tests again after our daily push tomorrow

blodan:
@JonathanDCohen I'll test the commit once it is pushed!

@JonathanDCohen Seing as we seem to be in completly different timezones I made you one better, I setup a dedicated server for you to test this on;
```
46.246.120.51
```

Credentials are tmpuser / temp123temp123

The repo is cloned and all other tests are built so you don't have to wait for anything;
```
cd ~/abseil-cpp && bazel test absl/...:all
```

JonathanDCohen:
@blodan I goe an error "46.246.120.51 refused to connect."

blodan:
@derekmauro @JonathanDCohen after your changes to type_traits all tests now builds so this should now be mergeable :)

@JonathanDCohen strange, you sure you used port 22 as that server did not have a firewall, I'll take it now anyhow as the issue has been resolved :)

derekmauro:
I haven't forgotten about this.  I'm testing to make sure the BUILD.bazel changes don't break any of the other platforms we support for some of our Google-internal users.  I'll let you know the results and if I can actually merge this as is.

absl/base/BUILD.bazel
```diff
@@ -17,6 +17,7 @@
 load(
     "//absl:copts.bzl",
     "ABSL_DEFAULT_COPTS",
+    "ABSL_DEFAULT_LINKOPTS",
```
derekmauro: I discussed this further with the team.  We aren't sure this viral ABSL_DEFAULT_LINKOPTS thing is the correct approach.  In fact, we've even started to reconsider whether ABSL_DEFAULT_COPTS is the correct approach.

Another possible approach is to use a CROSSTOOL file for this.  See https://github.com/bazelbuild/bazel/wiki/About-the-CROSSTOOL and https://github.com/bazelbuild/bazel/blob/master/tools/cpp/CROSSTOOL#L114.  There is even a way to set a custom CROSSTOOL file as the default for a Bazel WORKSPACE.

Until we resolve this, can you change this pull request to only include the changes to C++ files?  By this is mean limit this pull request to `defined(__FreeBSD__)` and the like.

You then can work around the linker issues with `bazel build --linkopt=-lm --linkopt=-pthread //path/to:target` or by defining a custom CROSSTOOL file for your project.

We can then open a separate issue where we can figure out the correct thing to do with our compiler and linker flags.

Sorry for conflicting requests and for requesting more changes, we are just trying to get this right.

blodan:
@derekmauro I've now removed the default linkopts again, tests succeed when --linkopt=-lm --linkopt=-phread is being used

derekmauro (approved):

PR #61, opened 2017-12-11
Fix build under MSVC

umogSlayer:
MSVC defines macros named `min` and `max` which conflict with
std::min/std::max.

### DIFF
--- a/absl/base/internal/low_level_alloc.cc
+++ b/absl/base/internal/low_level_alloc.cc
@@ -35,6 +35,9 @@
 #include <sys/mman.h>
 #include <unistd.h>
 #else
+#ifndef NOMINMAX
+#define NOMINMAX
+#endif
 #include <windows.h>
 #endif
 
--- a/absl/base/internal/spinlock_win32.inc
+++ b/absl/base/internal/spinlock_win32.inc
@@ -14,6 +14,9 @@
 //
 // This file is a Win32-specific part of spinlock_wait.cc
 
+#ifndef NOMINMAX
+#define NOMINMAX
+#endif
 #include <windows.h>
 #include <atomic>
 #include "absl/base/internal/scheduling_mode.h"
--- a/absl/base/internal/sysinfo.cc
+++ b/absl/base/internal/sysinfo.cc
@@ -15,6 +15,9 @@
 #include "absl/base/internal/sysinfo.h"
 
 #ifdef _WIN32
+#ifndef NOMINMAX
+#define NOMINMAX
+#endif
 #include <shlwapi.h>
 #include <windows.h>
 #else
--- a/absl/debugging/internal/stacktrace_win32-inl.inc
+++ b/absl/debugging/internal/stacktrace_win32-inl.inc
@@ -37,6 +37,9 @@
 #ifndef ABSL_DEBUGGING_INTERNAL_STACKTRACE_WIN32_INL_H_
 #define ABSL_DEBUGGING_INTERNAL_STACKTRACE_WIN32_INL_H_
 
+#ifndef NOMINMAX
+#define NOMINMAX
+#endif
 #include <windows.h>    // for GetProcAddress and GetModuleHandle
 #include <cassert>
 
--- a/absl/synchronization/internal/waiter.cc
+++ b/absl/synchronization/internal/waiter.cc
@@ -17,6 +17,9 @@
 #include "absl/base/config.h"
 
 #ifdef _WIN32
+#ifndef NOMINMAX
+#define NOMINMAX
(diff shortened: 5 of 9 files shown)


umogSlayer:
I've seen the #55, but I think the way I did it is way better, because it doesn't require users to have extra knowledge how to build with this library (i.e. extra compiler flags when building theirs code).
And I personally consider macros with names such as `min`/`max` to be very harmful and I don't want to have them anywhere near my code.

tituswinters:
I'm not sure this scales.  @zhangxy988 has probably thought about this more deeply than I.

zhangxy988:
Sorry for the late reply. As I stated in https://github.com/abseil/abseil-cpp/issues/34#issuecomment-337633532, I think this is an issue that should be addressed at the build system level rather than at source code level. I am willing to hear about the situations where
* you can't add `NOMINMAX` flag in the build system.
* you don't want `NOMINMAX` in some translation units.
These are the cases I think we need source-code-level remedy.

umogSlayer:
For example when building code that has pure C parts in it, and you would like to use those `min`/`max` (in pure C code). At the same time I wouldn't want to have those `min`/`max` defined for C++ code.

If this issue seems minor then I don't really have problems with leaving it like this.

So there are two points:
1. No extra knowledge is required to build _with_ `abseil` (from build system point of view).
2. Pure C code doesn't have to be changed.

zhangxy988:
So quick summary:
Downside of `#define NOMINMAX` at every `#include <windows.h>`:
1. This needs to be done EVERYWHERE in the code base and is thus not scalable. It's even not feasible in some cases where you don't own the code you depend on.
2. If it's not done EVERYWHERE, it depends on the header inclusion order: the first `#include <windows.h>` in a translation unit needs to have `NOMINMAX` defined before it.

Downside of using `-DNOMINMAX` compiler flag at build system level:
1. Requires some knowledge at build-system level and some work to add the compiler flag.
2. In cases where `NOMINMAX` is unwanted, special handling are needed, i.e. `-UNOMINMAX` for some translation units. I don't think these cases are restricted to pure C code.

In practice, our internal toolchain has been fine just having `-DNOMINMAX` as a compiler flag and in some rare case where it's not needed, we have `-UNOMINMAX` at package level or target level to suppress it.

IMO, I dislike the downside of `#define NOMINMAX` in source code more than `-DNOMINMAX`, and that's why I am reluctant to accept this PR. Here is an option I want to explore:

* see if I can remove `#include <windows.h>` from headers in Abseil, so Abseil doesn't "pollute" the users code.
* for .cc files, maybe we can add `#define NOMINMAX` if `std::min|max` is used to make it more friendly for users who don't want to add `-DNOMINMAX`.

My goal is to keep this as a patch of limited scope rather than promoting it as a best practice. The recommended way is still `-DNOMINMAX` at build system level. I'll investigate more and get back to your.

umogSlayer:
Thanks for productive discussion. I see your point and I don't insist on getting this PR through.
And I agree, removing `#include <windows.h>` from headers is the best option.

I would like to share our experience with you.
At our company inclusion paired with `#define NOMINMAX` actually scales better than defining it in a build system. The reason is a person who writes platform specific code usually knows better than a person who writes build scripts.
Situation: a) a linux developer changes build scripts a bit alongside with source code changes, b) sends them to review, c) his code is accepted, d) CI rejects builds since code doesn't compile on MSVC, e) linux developer in question spends another day or two to fix build problems (since he's not that familiar with MSVC and cannot build it locally).

PS: should I close this PR?

zhangxy988:
@umogSlayer : thanks for sharing the situation in your company. I am not sure I understand it fully and clearly:
* The build scripts: Are they shared by all platforms? Does each platform has its own build configurations? If a global `-DNOMINMAX` exist for Windows build, I don't understand how a Linux developer could possibly break Window build, by removing `-DNOMINMAX` accidentally maybe?
* "The reason is a person who writes platform specific code usually knows better than a person who writes build scripts" I don't fully agree. The build configuration for each platform should be maintained by developers working on the specific platform. In other words, build scripts/system configuration should be platform-specific.

Please feel free to close this PR as it is not likely to be accepted as is.
I will look into the alternative option I proposed above https://github.com/abseil/abseil-cpp/pull/61#issuecomment-352815372.
And we can continue this discussion either here (if you prefer) or in #34 .

zhangxy988:
I took a look at removing windows.h from the headers but there doesn't seem to be an easy way to do this. For the reasons I stated above, we still suggest users to `-DNOMINMAX` at build system level.
I am closing this PR. The windows.h issue will be tracked by #34 .

PR #62, opened 2017-12-11
Replace dynamic type dispatch with static type dispatch in strings/numbers.h

umogSlayer:
a) fixes warnings issued by MSVC (warning C4127: conditional expression is
constant),
b) removes runtime "signedness" check for integer-like types.

### DIFF
--- a/absl/strings/numbers.h
+++ b/absl/strings/numbers.h
@@ -32,8 +32,10 @@
 #include <string>
 #include <type_traits>
 
+#include "absl/base/internal/identity.h"
 #include "absl/base/macros.h"
 #include "absl/base/port.h"
+#include "absl/meta/type_traits.h"
 #include "absl/numeric/int128.h"
 #include "absl/strings/string_view.h"
 
@@ -75,11 +77,11 @@ ABSL_MUST_USE_RESULT bool SimpleAtob(absl::string_view str, bool* value);
 namespace absl {
 namespace numbers_internal {
 
-// safe_strto?() functions for implementing SimpleAtoi()
-bool safe_strto32_base(absl::string_view text, int32_t* value, int base);
-bool safe_strto64_base(absl::string_view text, int64_t* value, int base);
-bool safe_strtou32_base(absl::string_view text, uint32_t* value, int base);
-bool safe_strtou64_base(absl::string_view text, uint64_t* value, int base);
+// safe_strto_int_base<?>() functions for implementing SimpleAtoi()
+bool safe_strto_int_base(absl::string_view text, int32_t* value, int base);
+bool safe_strto_int_base(absl::string_view text, int64_t* value, int base);
+bool safe_strto_int_base(absl::string_view text, uint32_t* value, int base);
+bool safe_strto_int_base(absl::string_view text, uint64_t* value, int base);
 
 static const int kFastToBufferSize = 32;
 static const int kSixDigitsToBufferSize = 16;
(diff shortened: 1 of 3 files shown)


absl/strings/numbers.h
```diff
@@ -100,26 +100,80 @@ static const int kSixDigitsToBufferSize = 16;
...
+struct Identity final
+{
+	using type = any_type;
+};
+
+// Checks whether this int_type is signed or not, does it using underlying_type for enumerations.
+template<typename int_type>
+using IsSignedIntType =
+  std::is_signed<
+    typename std::conditional_t<
```
tituswinters: conditional_t is a C++14 API.  Gotta wait a few more years before we can rely on that in Abseil, I'm afraid.
That said, absl/meta/type_traits.h has an absl::conditional_t.

absl/strings/numbers.h
```diff
@@ -100,26 +100,80 @@ static const int kSixDigitsToBufferSize = 16;
...
+{
+  return safe_strto64_base(text, value, base);
+}
+template<> inline bool safe_strto_base<uint64_t>(absl::string_view text, uint64_t* value, int base)
+{
+  return safe_strtou64_base(text, value, base);
+}
+
+template<typename any_type>
+struct Identity final
```
tituswinters: This already exists in absl/base/internal/identity.h - maybe it should just move to meta/type_traits? Regardless, for now you should use that one.

absl/strings/numbers.h
```diff
@@ -100,26 +100,80 @@ static const int kSixDigitsToBufferSize = 16;
...
+template<typename int_type>
+using IsSignedIntType =
+  std::is_signed<
+    typename std::conditional_t<
+      std::is_enum<int_type>::value,
+      std::underlying_type<int_type>,
+      Identity<int_type>>::type>;
+
+template<typename int_type>
+using PrintableIntType =
```
tituswinters: PrintableIntType is a slightly weird name for this - it's printable only in the sense that we've written overloads for it. 

Perhaps something more like IntToSupportedInt or IntTo32Or64? I'd definitely name it more actively - it's more metafunction than type.

absl/strings/numbers.h
```diff
@@ -100,26 +100,80 @@ static const int kSixDigitsToBufferSize = 16;
...
+  std::conditional_t<
+    IsSignedIntType<int_type>::value,
+    std::conditional_t<
+      (sizeof(int_type) > 32 / 8),
```
tituswinters: IIRC, we're still aiming to not assume bits-per-byte.  Rather than 32/8, sizeof(int32_t)? (here and below)

absl/strings/numbers.h
```diff
@@ -137,31 +191,11 @@ ABSL_MUST_USE_RESULT bool SimpleAtoi(absl::string_view s, int_type* out) {
...
-      uint64_t val;
-      parsed = numbers_internal::safe_strtou64_base(s, &val, 10);
-      *out = static_cast<int_type>(val);
-    } else {  // 32-bit
-      uint32_t val;
-      parsed = numbers_internal::safe_strtou32_base(s, &val, 10);
-      *out = static_cast<int_type>(val);
-    }
-  }
+  using ParsedType = numbers_internal::PrintableIntType<int_type>;
```
tituswinters: ParsedType seems like an odd name: it isn't parsed from anything.  SupportedInt?

tituswinters (changes requested):
I'll also try to get the original maintainer of this to take a look.  

I'm curious: is this not being optimized already?  Is this just to address the warnings on MSVC?

umogSlayer:
Thanks for comments.
Basically yes, it's mainly to address MSVC warnings since optimizers should optimize that code away.
But interestingly enough this code generates 2 warnings on MSVC, the one I wrote and narrowing type conversion warning (this one I don't want to disable). Narrowing type conversion happens because (as it seems) compiler front-end generates all the code paths and on dead paths we actually have narrowing conversion.

absl/strings/numbers.h
umogSlayer: Fixed

absl/strings/numbers.h
umogSlayer: Fixed

absl/strings/numbers.h
umogSlayer: Sounds good to me. Fixed.

absl/strings/numbers.h
umogSlayer: Renamed to `IntLikeTo32Or64`, wrote a comment explaining meaning of this metafunction.

tituswinters:
Update: we're looking at various options for potentially restructuring that to address the MSVC warnings - we might go this way or we might find a different refactoring. I expect it'll get resolved/decided by early next week either way.

tituswinters:
The internal maintainer took a different approach, which I *think* should still resolve the MSVC warnings you were seeing.  Check the current HEAD and see if that resolves your issue? If not, we'll finish this off next week. Thanks.

umogSlayer:
I've looked at current `master` branch and those changes don't address the `C4127: conditional expression is
constant` warning. Since this line of code hasn't been changed: `if (sizeof(i) > 32 / 8) {`. But they address the narrowing cast warning.

umogSlayer:
I've changed my code a bit. Maybe this version is ok.

umogSlayer:
I have one question though. After implementing these changes I have failing tests. Is it intended use of `FastIntToBuffer` with custom "integer wrapper" like MyInteger at `numbers_test.cc:111`?

tituswinters:
Since it's internal, I guess that could in theory be relaxed. But I do like the more-explicit code as-is. 

Here's a thought: can we just #pragma suppress the MSVC warning for that block?  That feels potentially less awkward to me, overall.

johnforestolson:
Hello! Following up on this PR. Vladimir, any thoughts on Titus' suggestion?

ahedberg:
One more ping on this PR - any thoughts on Titus' suggestion, @umogSlayer? Are you still seeing MSVC warnings, and would #pragma suppression clear them up?

I'm inclined to close this PR by the end of the week if it remains inactive.

umogSlayer:
Hello!
Sorry for such a late response, had basically no time lately.

I don't like the idea to disable warnings for blocks of code. It seems to me that such approach is not scalable, if someone in clang or gcc adds the same warning then you'll need to handle that case separately.

Maybe it's better to use something like `if constexpr (...)` simulation?

ahedberg:
Apologies for the delay in following up here.

We're still leaning towards the pragma solution. The warnings are all false in those cases, and disabling them locally is safe and correct. A SFINAE-based solution feels excessive here. Thoughts?

ahedberg:
Given the lack of activity on this PR, I think we should close it--please reopen if you're still seeing MSVC warnings.

PR #72, opened 2018-01-03
Fix builds with BUILD_TESTING=OFF.

coryan:
Abseil requires the gtest/gmock/gtest_main targets even for
builds where BUILD_TESTING=OFF, where those targets are not
used.

The attached [CMakeLists.txt](https://github.com/abseil/abseil-cpp/files/1601827/CMakeLists.txt) can be used to reproduce the problem.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -69,9 +69,12 @@ find_package(Threads REQUIRED)
 
 ## check targets
 check_target(cctz)
-check_target(gtest)
-check_target(gtest_main)
-check_target(gmock)
+
+if (CONFIG_TESTING)
+  check_target(gtest)
+  check_target(gtest_main)
+  check_target(gmock)
+endif (CONFIG_TESTING)
 
 # -fexceptions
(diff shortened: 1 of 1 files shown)


coryan:
Any thoughts?  Should I cancel this PR because it is not a good idea?  Should I tweak it in some way?

derekmauro:
I assigned this to @JonathanDCohen, but he is on vacation now. He will get to this when he gets back. Sorry for the delay.

JonathanDCohen:
Just getting around to this.  Sorry for the wait.

https://github.com/abseil/abseil-cpp/issues/74 appears related -- I think the resolution is to pull out absl/base/exception_safety_testing.{h,cc} into their own CMake target and only compile them in test-only mode.

JonathanDCohen:
https://github.com/abseil/abseil-cpp/pull/69 seems to subsume this PR -- any objections to closing this one for the more complete change?

coryan:
No problem.

PR #77, opened 2018-01-10
spinlock.h : fix typo

bmusin:
contentionS


derekmauro:
This isn't a typo, it is a Googlism - 'contentionz' is the thing we use to collect contention profiles on servers.

PR #84, opened 2018-01-25
Converted absl_utility to Header-only CMake Target

peteigel:
When generating a build with CMake, I got the following error:
```
-- Configuring done
CMake Error: Cannot determine link language for target "absl_utility".
CMake Error: CMake can not determine linker language for target: absl_utility
```

Currently absl_utility has no actual sources, so there really isn't a link language. To correct this, I simply replicated the Header-only CMake config in [absl_algorithm](https://github.com/abseil/abseil-cpp/blob/master/absl/algorithm/CMakeLists.txt).

### DIFF
--- a/absl/utility/CMakeLists.txt
+++ b/absl/utility/CMakeLists.txt
@@ -19,19 +19,12 @@ list(APPEND UTILITY_PUBLIC_HEADERS
   "utility.h"
 )
 
-
-
-list(APPEND UTILITY_SRC
-  ${UTILITY_PUBLIC_HEADERS}
-)
-
 absl_header_library(
   TARGET
     absl_utility
   EXPORT_NAME
     utility
 )
-
 
 
 #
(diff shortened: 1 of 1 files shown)


peteigel:
I signed it!

johnforestolson:
Thanks! We'll review your change and get back to you...

absl/utility/CMakeLists.txt
```diff
@@ -25,13 +25,9 @@ list(APPEND UTILITY_SRC
   ${UTILITY_PUBLIC_HEADERS}
```
JonathanDCohen: does this UTILITY_SRC need to be defined?  We don't have it in algorithm and it seems redundant.

rongjiecomputer: @JonathanDCohen Indeed it is not needed. Unlike Bazel which requires header dependency to be stated in `BUILD` file explicitly, CMake will just let the target build system to figure this out themselves.

JonathanDCohen: Could you remove this too as part of the PR?

peteigel: Of course. It was an oversight on my part to leave it in.

PR #91, opened 2018-02-09
Annotate printf-like functions as such

gjasny:
This solves the following error:

```
../absl/base/internal/raw_logging.cc:86:34: error: format string is not a string literal [-Werror,-Wformat-nonliteral]
  int n = vsnprintf(*buf, *size, format, ap);
```


derekmauro:
Thank you for the contribution.

I was hoping this would let us remove
https://github.com/abseil/abseil-cpp/blob/8311b41974e50f2753921ba986bbd188f9bc28a0/absl/copts.bzl#L60
but it looks like there is another non-literal in `CheckHex()` in `str_cat_test.cc`. I'll follow up seperately and see if I can fix that one.

derekmauro (approved):

PR #99, opened 2018-03-23
Add configurable support for selecting mutex_nonprod.

deeglaze:
Our platform needs to be able to select the nonprod implementation of absl::Mutex. We can add --ABSL_USE_NONPROD_MUTEX=1 to our .bazelrc and compiler_flag: "-DABSL_USE_NONPROD_MUTEX" in our CROSSTOOL.

### DIFF
--- a/absl/synchronization/BUILD.bazel
+++ b/absl/synchronization/BUILD.bazel
@@ -24,6 +24,13 @@ package(default_visibility = ["//visibility:public"])
 
 licenses(["notice"])  # Apache 2.0
 
+config_setting(
+    name = "use_nonprod_mutex",
+    values = {
+      "define": "ABSL_USE_NONPROD_MUTEX=1",
+    },
+)
+
 # Internal data structure for efficiently detecting mutex dependency cycles
 cc_library(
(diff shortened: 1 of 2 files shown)


derekmauro:
It looks like you're a Googler. Can you use the internal code review process?

From https://github.com/abseil/abseil-cpp/blob/master/CONTRIBUTING.md:
`If you are a Googler, it is preferable to first create an internal CL and have it reviewed and submitted. The code propagation process will deliver the change to GitHub.`

In regards to the change itself, we really didn't want to support manual setting of using mutex_nonprod, but I don't see any other way to do it other than what you suggested.  I think of you make the condition `defined(ABSL_LOW_LEVEL_ALLOC_MISSING) && defined(ABSL_USE_NONPROD_MUTEX)` (instead of OR) we might accept this. But please tell us about your platform.

zhangxy988:
As our contribution policy, if this change is required inside Google, please follow internal workflow rather than opening a github issue.
Particularly for this issue, please raise a internal ticket so we can discuss your platform and how Abseil should support it.
Closing this.

PR #117, opened 2018-05-05
Set the FOLDER property for Abseil targets.

PetterS:
https://cmake.org/cmake/help/v3.0/prop_tgt/FOLDER.html
This only affects IDEs and will further only take effect if

    set_property(GLOBAL PROPERTY USE_FOLDERS ON)

is set in the main CMakeLists.txt of the project.

I found using this quite helpful when adding Abseil to larger CMake
projects.

### DIFF
--- a/CMake/AbseilHelpers.cmake
+++ b/CMake/AbseilHelpers.cmake
@@ -16,6 +16,11 @@
 
 include(CMakeParseArguments)
 
+# The IDE folder for Abseil that will be used if Abseil is included in a CMake
+# project that sets
+#    set_property(GLOBAL PROPERTY USE_FOLDERS ON)
+# For example, Visual Studio supports folders.
+set(ABSL_IDE_FOLDER Abseil)
 
 #
 # create a library in the absl namespace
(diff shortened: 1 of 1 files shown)


CMake/AbseilHelpers.cmake
```diff
@@ -49,6 +49,8 @@ function(absl_library)
     PUBLIC ${ABSL_COMMON_INCLUDE_DIRS} ${ABSL_LIB_PUBLIC_INCLUDE_DIRS}
     PRIVATE ${ABSL_LIB_PRIVATE_INCLUDE_DIRS}
   )
+  # Add all Abseil targets to a a folder in the IDE for organization.
+  set_property(TARGET ${_NAME} PROPERTY FOLDER Abseil)
```
JonathanDCohen: Nit: Please set "Abseil" in a variable like ABSL_IDE_FOLDER or some other descriptive name instead of using the string directly.

JonathanDCohen (approved):
Approved pending one small nit.

JonathanDCohen:
Thanks for the usability update!

PetterS:
Done!

Issue #114, opened 2018-05-02
CMake Refactor: Update interface of absl_library()


JonathanDCohen:
A long term goal for us is to be able to generate our CMakeLists.txt files from our Bazel BUILD files so as to relieve engineering overhead.  The first step for this is to make the corresponding cc_library and absl_library invocations in Bazel and CMake respectively have the same set of arguments, allowing us to easily generate one from the other.

rongjiecomputer:
I believe this is done with `absl_cc_{library,test}` now. Time to close this issue?

Issue #123, opened 2018-05-14
Match CMake compilation flags to Bazel compilation flags.


JonathanDCohen:
Our Bazel compilation flags are in https://github.com/abseil/abseil-cpp/blob/master/absl/copts.bzl.  We need these to be the same when building with CMake

Faisalsalipada:
Ok

asoffer:
This has been completed.

Issue #111, opened 2018-05-02
Step 6: Allow Abseil to be installed and used with find_package()


JonathanDCohen:
This has been a contentious one, but yes, Abseil is going to support the `git clone; make; make install` workflow.

This will come at the same time as Abseil cuts our first Long Term Support Branch -- basically cuts of our source code stuck in versioned inline namespaces which we will maintain for a length of time with emergency security bugfixes and the like, and which will also be tagged for use by package managers and the like.  Ideally, the `find_package()` workflow will only work with Long Term Support branches of Abseil -- it's the only safe way to use a precompiled version of abseil, as we can make sure that symbols don't clash unless they're identical (as long as they're compiled and linked with the same settings, etc).  So the supported usages of abseil will be either live-at-head using Bazel, a mirror, a git submodule, or config-time downloading of source as in googletest's model, or using a precompiled Long Term Support branch of Abseil.  It's unclear as of yet how this can be accomplished -- discussion on this point will happen closer to the point in time we're ready for this step.

JonathanDCohen:
#106

darix:
+1

pifon2a:
Are there any updates? [Google Cartographer](https://github.com/googlecartographer/cartographer) project now depends on Abseil. Unfortunately, since we are using ROS and catkin, we have to maintain the CMake build as well. We ended up using `ExternalProject_Add` in [FindAbseil.cmake](https://github.com/googlecartographer/cartographer/blob/master/cmake/modules/FindAbseil.cmake), but it is not optimal. Really looking forward to having `make install` in Abseil.

alexweej:
Is there any "work in progress" branch for this? I'm having a little stab myself in order to integrate with our organization's packaging system.

pifon2a:
@Mizux, are you planning to submit your CMake Install rules?

Mizux:
@pifon2a The current plan is to push internally one by one all the #199 commits, then maybe revamp (i.e. rebase/rework) my `Add CMake Install rules (abseil#111)` on top of it...

We will remove `absl_test()`, `absl_library()`, `absl_headeronly_library()` and `absl_test()` so we will "just" have to add install rules on `absl_cc_library()` and `absl_cc_test()`...

shardator:
is it now even possible to build Google Cartographer without internet access or a fake github repo?

pifon2a:
Unfortunately, not yet. After abseil has proper support for  `make install`, it will be possible.

[...]

PR #136, opened 2018-06-29
[CMake] Implement absl_cc_library as Bazel's cc_library 

rongjiecomputer:
`absl_cc_library` is a `cc_library`-like CMake function to replace `absl_library` and `absl_header_library`.

Please do not start any migration until `absl_cc_binary` and `absl_cc_test` are implemented in future PRs. I migrated a few targets just to showcase how `absl_cc_library` can be used.

#114

@JonathanDCohen PTAL

### DIFF
--- a/CMake/AbseilHelpers.cmake
+++ b/CMake/AbseilHelpers.cmake
@@ -121,7 +121,7 @@ function(absl_cc_library)
     ${ARGN}
   )
 
-  if (NOT ABSL_CC_LIB_TESTONLY OR BUILD_TESTING)
+  if (NOT ABSL_CC_LIB_TESTONLY OR ABSL_RUN_TESTS)
     set(_NAME "absl_${ABSL_CC_LIB_NAME}")
     string(TOUPPER ${_NAME} _UPPER_NAME)
 
--- a/CMake/AbseilHelpers.cmake
+++ b/CMake/AbseilHelpers.cmake
@@ -62,7 +62,114 @@ function(absl_library)
   endif()
 endfunction()
 
+#
+# CMake function to imitate Bazel's cc_library rule.
+#
+# Parameters:
+# NAME: name of target (see Note)
+# HDRS: List of public header files for the library
+# SRCS: List of source files for the library
+# DEPS: List of other libraries to be linked in to the binary targets
+# COPTS: List of private compile options
+# DEFINES: List of public defines
+# LINKOPTS: List of link options
+# VISIBILITY_PUBLIC: Add this so that this library will be exported under absl:: (see Note).
+# TESTONLY: When added, this target will only be built if user passes -DBUILD_TESTING=ON to CMake.
+#
+# Note:
+#
+# By default, absl_cc_library will always create a library named absl_${NAME},
+# which means other targets can only depend this library as absl_${NAME}, not ${NAME}.
+# This is to reduce namespace pollution.
+#
+# absl_cc_library(
+#   NAME
+#     awesome_lib
+#   HDRS
+#     "a.h"
+#   SRCS
+#     "a.cc"
+# )
+# absl_cc_library(
+#   NAME
+#     fantastic_lib
+#   SRCS
+#     "b.cc"
+#   DEPS
+#     absl_awesome_lib # not "awesome_lib"!
+# )
+#
+# If VISIBILITY_PUBLIC is set, absl_cc_library will also create an alias absl::${NAME}
+# for public use.
+#
+# absl_cc_library(
+#   NAME
+#     main_lib
+#   ...
+#   VISIBILITY_PUBLIC
+# )
+#
+# User can then use the library as absl::main_lib (although absl_main_lib is defined too).
+#
+# TODO: Implement "ALWAYSLINK"
 
+function(absl_cc_library)
+  cmake_parse_arguments(ABSL_CC_LIB
+    "DISABLE_INSTALL;VISIBILITY_PUBLIC;TESTONLY"
+    "NAME"
+    "HDRS;SRCS;COPTS;DEFINES;LINKOPTS;DEPS"
+    ${ARGN}
+  )
+
+  if (NOT ABSL_CC_LIB_TESTONLY OR BUILD_TESTING)
+    set(_NAME "absl_${ABSL_CC_LIB_NAME}")
+    string(TOUPPER ${_NAME} _UPPER_NAME)
+
+    # Check if this is a header-only library
+    if (ABSL_CC_LIB_SRCS)
+      set(_SRCS ${ABSL_CC_LIB_SRCS})
+      list(FILTER _SRCS INCLUDE REGEX "\.cc$")
+      list(LENGTH _SRCS ABSL_CC_LIB_SRCS_LEN)
+    else()
+      set(ABSL_CC_LIB_SRCS_LEN 0)
+    endif()
+
+    if(ABSL_CC_LIB_SRCS_LEN)
+      add_library(${_NAME} STATIC ${ABSL_CC_LIB_SRCS} ${ABSL_CC_LIB_HDRS})
+    else()
+      set(__dummy_header_only_lib_file "${CMAKE_CURRENT_BINARY_DIR}/${_NAME}_header_only_dummy.cc")
+
+      if(NOT EXISTS ${__dummy_header_only_lib_file})
+        file(WRITE ${__dummy_header_only_lib_file}
+          "/* generated file for header-only cmake target */
+
+          namespace absl {
+            // single meaningless symbol
+            void ${_NAME}__header_fakesym() {}
+          }  // namespace absl")
+      endif()
+
+      add_library(${_NAME} ${__dummy_header_only_lib_file} ${ABSL_CC_LIB_HDRS})
+    endif()
+
+    target_compile_options(${_NAME} PRIVATE ${ABSL_COMPILE_CXXFLAGS} ${ABSL_CC_LIB_COPTS})
+    target_link_libraries(${_NAME}
+      PUBLIC ${ABSL_CC_LIB_DEPS}
+      PRIVATE ${ABSL_CC_LIB_LINKOPTS}
+    )
+    target_compile_definitions(${_NAME} PUBLIC ${ABSL_CC_LIB_DEFINES})
+
+    target_include_directories(${_NAME}
+      PUBLIC ${ABSL_COMMON_INCLUDE_DIRS}
+    )
+    # Add all Abseil targets to a a folder in the IDE for organization.
+    set_property(TARGET ${_NAME} PROPERTY FOLDER ${ABSL_IDE_FOLDER})
+
+    if(ABSL_CC_LIB_VISIBILITY_PUBLIC)
+      add_library(absl::${ABSL_CC_LIB_NAME} ALIAS ${_NAME})
+    endif()
+  endif()
+endfunction()
 
 #
 # header only virtual target creation
--- a/CMake/AbseilHelpers.cmake
+++ b/CMake/AbseilHelpers.cmake
@@ -73,7 +73,7 @@ endfunction()
 # COPTS: List of private compile options
 # DEFINES: List of public defines
 # LINKOPTS: List of link options
-# VISIBILITY_PUBLIC: Add this so that this library will be exported under absl:: (see Note).
+# PUBLIC: Add this so that this library will be exported under absl:: (see Note).
 # TESTONLY: When added, this target will only be built if user passes -DBUILD_TESTING=ON to CMake.
 #
 # Note:
@@ -99,14 +99,14 @@ endfunction()
 #     absl_internal_awesome_lib # not "awesome_lib"!
 # )
 #
-# If VISIBILITY_PUBLIC is set, absl_cc_library will also create an alias absl::${NAME}
+# If PUBLIC is set, absl_cc_library will also create an alias absl::${NAME}
 # for public use in addition to absl_internal_${NAME}.
 #
 # absl_cc_library(
 #   NAME
 #     main_lib
 #   ...
-#   VISIBILITY_PUBLIC
+#   PUBLIC
 # )
 #
 # User can then use the library as absl::main_lib (although absl_internal_main_lib is defined too).
@@ -115,7 +115,7 @@ endfunction()
 
 function(absl_cc_library)
   cmake_parse_arguments(ABSL_CC_LIB
-    "DISABLE_INSTALL;VISIBILITY_PUBLIC;TESTONLY"
+    "DISABLE_INSTALL;PUBLIC;TESTONLY"
     "NAME"
     "HDRS;SRCS;COPTS;DEFINES;LINKOPTS;DEPS"
     ${ARGN}
@@ -123,7 +123,6 @@ function(absl_cc_library)
 
   if (NOT ABSL_CC_LIB_TESTONLY OR ABSL_RUN_TESTS)
     set(_NAME "absl_internal_${ABSL_CC_LIB_NAME}")
-    string(TOUPPER ${_NAME} _UPPER_NAME)
 
     # Check if this is a header-only library
     if (ABSL_CC_LIB_SRCS)
@@ -157,7 +156,7 @@ function(absl_cc_library)
       target_compile_definitions(${_NAME} INTERFACE ${ABSL_CC_LIB_DEFINES})
     endif()
 
-    if(ABSL_CC_LIB_VISIBILITY_PUBLIC)
+    if(ABSL_CC_LIB_PUBLIC)
       add_library(absl::${ABSL_CC_LIB_NAME} ALIAS ${_NAME})
     endif()
   endif()
--- a/CMake/AbseilHelpers.cmake
+++ b/CMake/AbseilHelpers.cmake
@@ -136,34 +136,26 @@ function(absl_cc_library)
 
     if(ABSL_CC_LIB_SRCS_LEN)
       add_library(${_NAME} STATIC ${ABSL_CC_LIB_SRCS} ${ABSL_CC_LIB_HDRS})
+      target_include_directories(${_NAME}
+        PUBLIC ${ABSL_COMMON_INCLUDE_DIRS})
+      target_compile_options(${_NAME}
+        PRIVATE ${ABSL_COMPILE_CXXFLAGS} ${ABSL_CC_LIB_COPTS})
+      target_link_libraries(${_NAME}
+        PUBLIC ${ABSL_CC_LIB_DEPS}
+        PRIVATE ${ABSL_CC_LIB_LINKOPTS}
+      )
+      target_compile_definitions(${_NAME} PUBLIC ${ABSL_CC_LIB_DEFINES})
+
+      # Add all Abseil targets to a a folder in the IDE for organization.
+      set_property(TARGET ${_NAME} PROPERTY FOLDER ${ABSL_IDE_FOLDER})
     else()
-      set(__dummy_header_only_lib_file "${CMAKE_CURRENT_BINARY_DIR}/${_NAME}_header_only_dummy.cc")
-
-      if(NOT EXISTS ${__dummy_header_only_lib_file})
-        file(WRITE ${__dummy_header_only_lib_file}
-          "/* generated file for header-only cmake target */
-
-          namespace absl {
-            // single meaningless symbol
-            void ${_NAME}__header_fakesym() {}
-          }  // namespace absl")
-      endif()
-
-      add_library(${_NAME} ${__dummy_header_only_lib_file} ${ABSL_CC_LIB_HDRS})
+      add_library(${_NAME} INTERFACE)
+      target_include_directories(${_NAME} INTERFACE ${ABSL_COMMON_INCLUDE_DIRS})
+      target_link_libraries(${_NAME}
+        INTERFACE ${ABSL_CC_LIB_DEPS} ${ABSL_CC_LIB_LINKOPTS}
+      )
+      target_compile_definitions(${_NAME} INTERFACE ${ABSL_CC_LIB_DEFINES})
     endif()
-
-    target_compile_options(${_NAME} PRIVATE ${ABSL_COMPILE_CXXFLAGS} ${ABSL_CC_LIB_COPTS})
-    target_link_libraries(${_NAME}
-      PUBLIC ${ABSL_CC_LIB_DEPS}
-      PRIVATE ${ABSL_CC_LIB_LINKOPTS}
-    )
-    target_compile_definitions(${_NAME} PUBLIC ${ABSL_CC_LIB_DEFINES})
-
-    target_include_directories(${_NAME}
-      PUBLIC ${ABSL_COMMON_INCLUDE_DIRS}
-    )
-    # Add all Abseil targets to a a folder in the IDE for organization.
-    set_property(TARGET ${_NAME} PROPERTY FOLDER ${ABSL_IDE_FOLDER})
 
     if(ABSL_CC_LIB_VISIBILITY_PUBLIC)
       add_library(absl::${ABSL_CC_LIB_NAME} ALIAS ${_NAME})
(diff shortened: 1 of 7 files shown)


CMake/AbseilHelpers.cmake
```diff
@@ -62,7 +62,114 @@ function(absl_library)
...
+#
+# Parameters:
+# NAME: name of target (see Note)
+# HDRS: List of public header files for the library
+# SRCS: List of source files for the library
+# DEPS: List of other libraries to be linked in to the binary targets
+# COPTS: List of private compile options
+# DEFINES: List of public defines
+# LINKOPTS: List of link options
+# VISIBILITY_PUBLIC: Add this so that this library will be exported under absl:: (see Note).
```
JonathanDCohen: Maybe just PUBLIC?  Or is that reserved by cmake?

CMake/AbseilHelpers.cmake
```diff
@@ -62,7 +62,114 @@ function(absl_library)
...
+# TESTONLY: When added, this target will only be built if user passes -DBUILD_TESTING=ON to CMake.
+#
+# Note:
+#
+# By default, absl_cc_library will always create a library named absl_${NAME},
```
JonathanDCohen: Here's an idea, while we're here.  If VISIBILITY_PUBLIC is not set, the prefix should be absl_internal_blah, otherwise it should be absl::blah.  What do you think?  I'm okay with "internal" appearing a bunch of times in the name.

CMake/AbseilHelpers.cmake
```diff
@@ -62,7 +62,114 @@ function(absl_library)
...
 
+function(absl_cc_library)
+  cmake_parse_arguments(ABSL_CC_LIB
+    "DISABLE_INSTALL;VISIBILITY_PUBLIC;TESTONLY"
+    "NAME"
+    "HDRS;SRCS;COPTS;DEFINES;LINKOPTS;DEPS"
+    ${ARGN}
+  )
+
+  if (NOT ABSL_CC_LIB_TESTONLY OR BUILD_TESTING)
```
JonathanDCohen: should this be ABSL_RUN_TESTS?

CMake/AbseilHelpers.cmake
```diff
@@ -62,7 +62,114 @@ function(absl_library)
...
+      list(FILTER _SRCS INCLUDE REGEX "\.cc$")
+      list(LENGTH _SRCS ABSL_CC_LIB_SRCS_LEN)
+    else()
+      set(ABSL_CC_LIB_SRCS_LEN 0)
+    endif()
+
+    if(ABSL_CC_LIB_SRCS_LEN)
+      add_library(${_NAME} STATIC ${ABSL_CC_LIB_SRCS} ${ABSL_CC_LIB_HDRS})
+    else()
+      set(__dummy_header_only_lib_file "${CMAKE_CURRENT_BINARY_DIR}/${_NAME}_header_only_dummy.cc")
```
JonathanDCohen: Let's take the chance to improve header-only libs while we're here.  Can we not make header-only libs INTERFACE libraries?  Then we just declare their deps as normal and the headers get picked up by ABSL_COMMON_INCLUDE_DIRS.

absl/strings/CMakeLists.txt
```diff
@@ -14,88 +14,71 @@
...
-absl_library(
-  TARGET
-    absl_strings
-  SOURCES
-    ${STRINGS_SRC}
-  PUBLIC_LIBRARIES
-    ${STRINGS_PUBLIC_LIBRARIES}
-  EXPORT_NAME
+absl_cc_library(
+  NAME
```
JonathanDCohen: I would rather we didn't immediately migrate something so widely used as strings.  I think the exception-safety testing is fine, it has a small footprint.  If you want another trial case, let's try something internal instead.

JonathanDCohen (changes requested):

JonathanDCohen:
Thanks for this!  It's been gathering dust on my todo list for a while

CMake/AbseilHelpers.cmake
rongjiecomputer: You mean replace `BUILD_TESTING` with `ABSL_RUN_TESTS`? There is no difference actually.

When `ABSL_RUN_TESTS` is not set, `CTest` won't be included and `BUILD_TESTING` won't be specified. Un-specified variable has default value of "false" when accessed.

CMake/AbseilHelpers.cmake
rongjiecomputer: Yes we can! Thanks for the reminder. Done.

CMake/AbseilHelpers.cmake
rongjiecomputer: `PUBLIC` is not reserved by CMake, but I just thought that `VISIBILITY_PUBLIC` is easier to understand for Abseil team since Bazel uses `//visibility:public` to mean that "targets outside Abseil can depend on it". I can change it to `PUBLIC` if you think that is better.

CMake/AbseilHelpers.cmake
rongjiecomputer: I realized that `BUILD_TESTING` is global, not project-specific, so depending on `BUILD_TESTING` will force consumer to run Abseil tests when they just want to run their own tests. Done.

absl/strings/CMakeLists.txt
rongjiecomputer: Done.

CMake/AbseilHelpers.cmake
rongjiecomputer: `absl_internal_` preffix will reduce the chance of name collision, but we might still face name collision *within* Abseil itself (some targets have name that are too general like `test_util` in `//absl/time` package). I have a longer response at https://groups.google.com/forum/#!topic/abseil-io/_Pq2NuCuYcc

I am settling at `absl_` preffix for now since Abseil's CMake files use `absl_` preffix too (e.g. `absl_throw_delegate`), so I don't have to do too much manual editing when showing some possible migrations in this PR.

JonathanDCohen: I'm not really worried about name collisions as much as I'm trying to make it as inconvenient as possible to rely on abseil internals, since we have no true hiding mechanism as we do in Bazel.  Does that make more sense?

CMake/AbseilHelpers.cmake
JonathanDCohen: I think just PUBLIC is fine

CMake/AbseilHelpers.cmake
```diff
@@ -62,7 +62,106 @@ function(absl_library)
...
+
+function(absl_cc_library)
+  cmake_parse_arguments(ABSL_CC_LIB
+    "DISABLE_INSTALL;VISIBILITY_PUBLIC;TESTONLY"
+    "NAME"
+    "HDRS;SRCS;COPTS;DEFINES;LINKOPTS;DEPS"
+    ${ARGN}
+  )
 
+  if (NOT ABSL_CC_LIB_TESTONLY OR ABSL_RUN_TESTS)
```
JonathanDCohen, line 124: This all seems wrong.  For testonly, I'm not actually sure there's a way to do that in cmake.  That requires that a library be depended on only by test targets.  My guess is we'd have to do something with the cache to mark targets as test only and then check against it when we declare link libraries for tests and libs?  Anyways I think it can be added later.  For now, we can have testonly be a pure documentation thing.

As for if(ABSL_RUN_TESTS), I don't really understand why we only want to define the libraries when we're running tests.  That test wants to be in the cc_test rule, not the cc_library rule.

CMake/AbseilHelpers.cmake
```diff
@@ -62,7 +62,106 @@ function(absl_library)
...
+    if (ABSL_CC_LIB_SRCS)
+      set(_SRCS ${ABSL_CC_LIB_SRCS})
+      list(FILTER _SRCS INCLUDE REGEX "\.cc$")
+      list(LENGTH _SRCS ABSL_CC_LIB_SRCS_LEN)
+    else()
+      set(ABSL_CC_LIB_SRCS_LEN 0)
+    endif()
+
+    if(ABSL_CC_LIB_SRCS_LEN)
+      add_library(${_NAME} STATIC ${ABSL_CC_LIB_SRCS} ${ABSL_CC_LIB_HDRS})
```
JonathanDCohen: Can we split these out into library and header-only-library helper functions?

CMake/AbseilHelpers.cmake
```diff
@@ -62,7 +62,106 @@ function(absl_library)
...
+      target_include_directories(${_NAME}
+        PUBLIC ${ABSL_COMMON_INCLUDE_DIRS})
```
JonathanDCohen, line 142: Continuing to try to improve our technique here, we could make an absl_common INTERFACE library which exposes ABSL_COMMON_FOO in order for us to migrate away from the global vars.

CMake/AbseilHelpers.cmake
rongjiecomputer: Done.

CMake/AbseilHelpers.cmake
```diff
@@ -62,7 +62,106 @@ function(absl_library)
...
+      target_compile_options(${_NAME}
+        PRIVATE ${ABSL_COMPILE_CXXFLAGS} ${ABSL_CC_LIB_COPTS})
```
JonathanDCohen, line 145: As far as I can tell ABSL_COMPILE_CXXFLAGS is never set anywhere.  Using it was probably a bug in the old cc_library function.

CMake/AbseilHelpers.cmake
rongjiecomputer: Done.

CMake/AbseilHelpers.cmake
rongjiecomputer: When `cc_library.testonly` is set to 1, it means that only `cc_test` can depend on it (1), which also means that user who don't care about Abseil's tests can also simply skip building these libraries.

CMake does not have the ability to verify this (1), but since we are just translating Bazel `BUILD` files to CMake (2), if CMake files are wrong, so are Bazel `BUILD` files and Bazel build should fail immediately (i.e. "proof by Bazel").

(2): This is the purpose of `absl_cc_library` and the forthcoming `absl_cc_test`.

CMake/AbseilHelpers.cmake
rongjiecomputer: I think it was there because the original author wants to use it to implement `ABSL_DEFAULT_COPTS` and `ABSL_TEST_COPTS` from Bazel. I think it is better to leave it for now and revisit later when we deal with #123.

CMake/AbseilHelpers.cmake
rongjiecomputer: I don't think we really need to go that far to avoid global variable. If we keep the number of global variables low and make sure they have `ABSL_` preffix, we should be OK. Other global variables like `EXCEPTION_SAFETY_TESTING_SRC` should definitely be removed.

CMake/AbseilHelpers.cmake
rongjiecomputer: That will over-complicate things IMHO. This function will need to pass many variables such as `_NAME`, `ABSL_CC_LIB_{SRCS,HDRS,LINKOPTS,COPTS,DEPS,DEFINES}`. That is just too long.

rongjiecomputer:
I have no more changes to make until next round of review.

CMake/AbseilHelpers.cmake
JonathanDCohen: Yeah agreed that we should try to remove the FOO_SRC variables.  This is fine for now I suppose, although it would be a nice follow-up to keep tidying up the place :)

rongjiecomputer:
@JonathanDCohen Is there anything else I need to address?

rongjiecomputer:
Ping @JonathanDCohen

CMake/AbseilHelpers.cmake
JonathanDCohen: This is fine, can we leave a comment or TODO?

CMake/AbseilHelpers.cmake
```diff
@@ -62,7 +62,105 @@ function(absl_library)
...
+      target_link_libraries(${_NAME}
+        PUBLIC ${ABSL_CC_LIB_DEPS}
+        PRIVATE ${ABSL_CC_LIB_LINKOPTS}
+      )
+      target_compile_definitions(${_NAME} PUBLIC ${ABSL_CC_LIB_DEFINES})
+
+      # Add all Abseil targets to a a folder in the IDE for organization.
+      set_property(TARGET ${_NAME} PROPERTY FOLDER ${ABSL_IDE_FOLDER})
+    else()
```
JonathanDCohen, line 154: Comment here please noting that we're generating a header-only library

JonathanDCohen:
Sorry for the delay, this fell off of my radar

CMake/AbseilHelpers.cmake
rongjiecomputer: Done.

CMake/AbseilHelpers.cmake
rongjiecomputer: Done.

CMake/AbseilHelpers.cmake
```diff
@@ -62,7 +62,107 @@ function(absl_library)
...
+    set(_NAME "absl_internal_${ABSL_CC_LIB_NAME}")
+
+    # Check if this is a header-only library
+    if (ABSL_CC_LIB_SRCS)
+      set(_SRCS ${ABSL_CC_LIB_SRCS})
+      list(FILTER _SRCS INCLUDE REGEX "\.cc$")
+      list(LENGTH _SRCS ABSL_CC_LIB_SRCS_LEN)
+    else()
+      set(ABSL_CC_LIB_SRCS_LEN 0)
+    endif()
```
Mizux, line 136: Is it possible to have a header only lib with SRCS parameter not empty ?

CMake/AbseilHelpers.cmake
```diff
@@ -62,7 +62,107 @@ function(absl_library)
   endif()
 endfunction()
 
+#
+# CMake function to imitate Bazel's cc_library rule.
```
Mizux, line 66: you should add a link to https://docs.bazel.build/versions/master/be/c-cpp.html#cc_library
Usually CMake user don't use bazel

CMake/AbseilHelpers.cmake
```diff
@@ -62,7 +62,107 @@ function(absl_library)
...
+# absl_cc_library(
+#   NAME
+#     fantastic_lib
+#   SRCS
+#     "b.cc"
+#   DEPS
+#     absl_internal_awesome_lib # not "awesome_lib"!
+# )
+#
+# If PUBLIC is set, absl_cc_library will also create an alias absl::${NAME}
```
Mizux: why don't using "absl_" prefix for PUBLIC enable libraries ?

CMake/AbseilHelpers.cmake
rongjiecomputer: Technically this `absl_cc_library` is not for use of normal CMake users. The hope is to create CMake functions that closely match the Bazel rules (at least, as close as needed by Abseil) so that we can generate CMakeFiles.txt from script without human intervention (so that Abseil team don't need to go and learn CMake). It is implementation detail that CMake user as a consumer of Abseil library does not need to know.

CMake/AbseilHelpers.cmake
rongjiecomputer: It is already there. The following code will create `absl_strings` *and* `absl::strings` (alias). CMake user should depend on `absl::strings` only.

```
absl_cc_library(
  NAME
    strings
  PUBLIC
)
```

CMake/AbseilHelpers.cmake
rongjiecomputer: Why is it necessary to have a header-only lib with SRCS parameter? Does Bazel allow other targets to depend on `.h` files declared in `srcs` of a target? At least from what I can see, Abseil's `BUILD` files don't do that.

CMake/AbseilHelpers.cmake
Mizux: no it will create `absl_internal_strings` *and* `absl::strings` line 125
you should have something like this:
```cmake
if (ABSL_CC_LIB_PUBLIC) 
  set(_NAME "absl_${ABSL_CC_LIB_NAME}")
else()
  set(_NAME "absl_internal_${ABSL_CC_LIB_NAME}")
endif()

```

CMake/AbseilHelpers.cmake
Mizux: Exactly, `INTERFACE` with `_SRCS` will be an error (i.e. an ill-formed header only lib).
So you can simply check if SRCS is empty to know if it's a `INTERFACE` or not:
```cmake
if ("${ABSL_CC_LIB_SRCS}" STREQUAL "")
  set(ABSL_CC_LIB_IS_INTERFACE 1)
else()
  set(ABSL_CC_LIB_IS_INTERFACE 0)
endif()
```
then later
```cmake
if(NOT ABSL_CC_LIB_IS_INTERFACE)
  add_library(${NAME} STATIC "")
  target_sources(${NAME} PRIVATE ${_SRCS})
  ...
else(NOT ABSL_CC_LIB_IS_INTERFACE)
  add_library(${NAME} INTERFACE)
endif()
```

CMake/AbseilHelpers.cmake
```diff
@@ -62,7 +62,107 @@ function(absl_library)
...
+    if(ABSL_CC_LIB_SRCS_LEN)
+      add_library(${_NAME} STATIC ${ABSL_CC_LIB_SRCS} ${ABSL_CC_LIB_HDRS})
+      target_include_directories(${_NAME}
+        PUBLIC ${ABSL_COMMON_INCLUDE_DIRS})
+      # TODO(rongjiecomputer): Revisit ABSL_COMPILE_CXXFLAGS when fixing GH#123
+      target_compile_options(${_NAME}
+        PRIVATE ${ABSL_COMPILE_CXXFLAGS} ${ABSL_CC_LIB_COPTS})
+      target_link_libraries(${_NAME}
+        PUBLIC ${ABSL_CC_LIB_DEPS}
+        PRIVATE ${ABSL_CC_LIB_LINKOPTS}
```
Mizux, line 148: should be
```cmake
target_link_options(${_NAME} PRIVATE ${ABSL_CC_LIB_LINKOPTS})
```

src: https://cmake.org/cmake/help/latest/command/target_link_options.html

CMake/AbseilHelpers.cmake
rongjiecomputer: > no it will create absl_internal_strings and absl::strings line 125

Ah, I forgot that the prefix has changed to `absl_internal_` (it was initially just `absl_`, but Jonathan requested to always use `absl_internal_` for non-public targets and I didn't think of creating exception for public targets).

Is it better for user to use `absl_strings` instead of `absl::strings`? Because I want to make it clear to CMake user that only `absl::` can be depended outside Abseil. (No strong objection here, just want your opinion).

CMake/AbseilHelpers.cmake
rongjiecomputer: Abseil still supports CMake 3.1, so can't use `target_link_options`.

src: https://cmake.org/cmake/help/v3.1/manual/cmake-commands.7.html

Mizux: For the record, target_link_options come with cmake 3.13 :disappointed:

CMake/AbseilHelpers.cmake
rongjiecomputer: Just checking, so we don't need to check `ABSL_CC_LIB_SRCS_LEN` anymore right?

EDIT: 

~~It seems like `target_sources` is also a command that is available only in fairly new version of CMake.~~

`target_sources` is available in 3.1 (https://cmake.org/cmake/help/v3.1/command/target_sources.html)

CMake/AbseilHelpers.cmake
Mizux: Since:
1) Abseil force user to use abseil-cpp through `add_subdirectory(abseil-cpp)`,
So meta project who want to add an install rules will fail at configure time if they don't add `absl_string` to one of their `install(TARGETS ...` (that's why so much people want install rule cf #111).

2) Twist:
>  An `ALIAS` target may not be installed or exported.

src: https://cmake.org/cmake/help/latest/command/add_library.html#alias-libraries

-> `1 + 2` -> you can't integrate `abseil-cpp` in your project (and only use `absl::`) unless you'll never install your awesome target...
(i.e. unless target dependency is an imported target, it must be part of your install, aka avoid to install half target in the build tree)
note: for GTEST etc it works to use `add_subdirectory()` tricks since usually no one install test executables (and all their dependencies needed and build in the build_dir aka gtest)

CMake/AbseilHelpers.cmake
rongjiecomputer: So the correct usage of `target_sources` should be this?

```
      target_sources(${_NAME}
          PRIVATE ${ABSL_CC_LIB_SRCS}
          PUBLIC ${ABSL_CC_LIB_HDRS}
      )
```

CMake/AbseilHelpers.cmake
rongjiecomputer: Done.

CMake/AbseilHelpers.cmake
Mizux: You should keep `${ABSL_CC_LIB_HDRS}` private. PUBLIC want absolute out of the source file (I'm currently trying to play with it) :confused: and figure out how is it working.

note: There is also `PUBLIC_HEADER` property but it for install headers purpose but it shouldn't be part of this PR.

rongjiecomputer:
@Mizux Really appreciate your comments, keep them coming! I really need to go to sleep now, so don't wait for my response.

CMake/AbseilHelpers.cmake
Mizux: Hi, I dig a little on this topic (cf comment in google/cctz#83 --> https://github.com/google/cctz/pull/83/files/6cd878965d04595624f512546f59682b165ccd00#diff-af3b638bc2a3e6c650974192a53c7291)
TLDR the correct usage seems (when there is no install rule defined like it is the case for this PR):
```cmake
      target_sources(${_NAME}
          PRIVATE ${ABSL_CC_LIB_SRCS}
          PUBLIC ${ABSL_CC_LIB_HDRS}
      )
```
**BUT** HDRS must use `${CMAKE_CURRENT_SOURCE_DIR}/include file.h`
So as  I say let's put includes in `PRIVATE` for this PR then in an other let's update to PUBLIC if we want...

side note: Once you have install rules you need to use generator also (note: no part of this PR however, (i'll do it in PR182 once I rebase it on this one...))
I do some test in my pet project https://github.com/Mizux/plop if you are curious...

rongjiecomputer: `ABSL_CC_LIB_IS_INTERFACE` and `target_sources(PRIVATE)` done.

Just note that you will need to update both `absl_library` and `absl_cc_library` for your install PR because I have migrated a few private targets to use `absl_cc_library` for demonstration purpose.

Mizux: Yeah I know i need to trash everything and rebuild on top of your PR.
no worry :+1:

Mizux:
LGTM to me thanks @rongjiecomputer 
@JonathanDCohen seems odd that PR don't have any CI job associated to them (you should strongly consider having Travis/Appveyor jobs than only kokoro ones)

CMake/AbseilHelpers.cmake
```diff
@@ -62,7 +62,110 @@ function(absl_library)
...
+# Parameters:
+# NAME: name of target (see Note)
+# HDRS: List of public header files for the library
+# SRCS: List of source files for the library
+# DEPS: List of other libraries to be linked in to the binary targets
+# COPTS: List of private compile options
+# DEFINES: List of public defines
+# LINKOPTS: List of link options
+# PUBLIC: Add this so that this library will be exported under absl:: (see Note).
+# TESTONLY: When added, this target will only be built if user passes -DBUILD_TESTING=ON to CMake.
```
JonathanDCohen: I think you mean -DABSL_RUN_TESTS=ON

JonathanDCohen (commented):
looks good to me besides the small comment change.  @Mizux look good for you?

CMake/AbseilHelpers.cmake
rongjiecomputer: Thanks! Done.

Mizux:
Yes everything sound good to me.

note: I will rework #182 once this one is merged

pifon2a:
@JonathanDCohen  is it ready to merge?

JonathanDCohen (approved):

PR #138, opened 2018-07-03
Remove usage of soon-to-be-deprecated http_archive rule

edbaunton:
Per https://groups.google.com/d/msg/bazel-discuss/dO2MHQLwJF0/2OXHjMAaAAAJ the native rules for http_archive are to be replaced with their skylark implementations. This commit updates the WORKSPACE to use the skylark definitions.

Tested using `bazel build //... --incompatible_remove_native_http_archive --incompatible_remove_native_git_repository`

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -1,4 +1,6 @@
 workspace(name = "com_google_absl")
+load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
+
 # Bazel toolchains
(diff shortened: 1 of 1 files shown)


tituswinters:
We're investigating an unrelated CI failure, otherwise this should be fine to merge I believe.

ahedberg:
Apologies for the delay in merging, and thanks for the contribution!

Issue #98, opened 2018-03-23
absl::Mutex has multiple "no member named '...'" errors when !ABSL_HAVE_MMAP.


deeglaze:
There are some declarations in mutex.h that are in the "else" branch of #ifdef ABSL_INTERNAL_USE_NONPROD_MUTEX that mutex.cc has no similar #ifdef protection around the definitions. This leads to many compiler errors.

derekmauro:
Can post the compile errors and give us details about the platform you are building this on?

deeglaze:
I just determined this is due to some BUILD select expressions that should be added for our platform. I'll have to work from a patched archive in the meantime.

PR #139, opened 2018-07-05
Indicate Solaris / Illumos has mmap support

siepkes:
After bumping into #98 on SmartOS (Illumos derivative) I added Solaris to the list of OS'es with `mmap` support.


tituswinters:
According to https://sourceforge.net/p/predef/wiki/OperatingSystems/ this seems like a reasonable way to do this. I'll run some tests and double check some things, but conceptually this seems fine to merge.

Do note that this isn't an OS we officially support - it's going to be best-effort.

siepkes:
Great, thanks!

ahedberg:
Thanks for your patience, and apologies in the delay in getting this merged. Internal testing shows that the CI errors here are unrelated.

PR #140, opened 2018-07-05
Ensure types like 'month_t' can never be negative

siepkes:
Fixes build failures when using '-Werror=char-subscripts'. I encountered this issue while building Envoy on SmartOS (Solaris / Illumos) with Bazel and GCC 7.

`std::int_fast64_t` can be negative. Since it is directly used to reference positions in arrays the compiler complains about it. Making the types unsigned ensure they can never be negative.

Now the whole thing looks sensible and correct to me but I'm not an abseil-pro. So I recommend taking a good look at this PR to ensure I didn't break anything ;-).

### DIFF
--- a/absl/time/internal/cctz/include/cctz/civil_time_detail.h
+++ b/absl/time/internal/cctz/include/cctz/civil_time_detail.h
@@ -45,11 +45,11 @@ using diff_t = std::int_fast64_t;
 namespace detail {
 
 // Type aliases that indicate normalized argument values.
-using month_t = std::int_fast8_t;   // [1:12]
-using day_t = std::int_fast8_t;     // [1:31]
-using hour_t = std::int_fast8_t;    // [0:23]
-using minute_t = std::int_fast8_t;  // [0:59]
-using second_t = std::int_fast8_t;  // [0:59]
+using month_t = std::uint_fast8_t;   // [1:12]
+using day_t = std::uint_fast8_t;     // [1:31]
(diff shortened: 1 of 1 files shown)


rongjiecomputer:
You should send the change to https://github.com/google/cctz instead. Abseil will import change from cctz later.

tituswinters:
I would recommend against pursuing this.  Use of unsigned to represent "can't be negative" is an attractive nuisance, it's not a good fit for the language rules.  I'll repost below what I believe represents the majority opinion of the standards committee, Google's style guide, etc. 


Signedness is one of the worst things about C++.  

* The standard types are unsigned sizes (size_t), and the majority of standards committee members I've talked to believe that it is a mistake.  The committee just can't fix it.
* The compiler can better optimize (sometimes significantly) for signed types rather than unsigned types, because the defined wraparound behavior is often more expensive (or the signed behavior optimizes away better because it can ignore the endpoints).
* Things like ubsan can check against signed wraparound (because it's undefined), but we must assume unsigned wraparound is intended: even when it's just "vector.size() - 1" and the bug is basically evident to a human.
* Absent anything else, unsigned is only good for representing bitfields and closed-group arithmetic.  However, we're not absent anything else because we have the standard library, so drawing the line is hard.
* The only thing worse than all of this is mixing signed and unsigned continuously.

Push people to iterators and auto, don't pass around unsigned types if at all possible.  Try not to mix.

siepkes:
@tituswinters Thanks for the detailed explanation. I'm going back to the drawing board with this.

PR #152, opened 2018-07-31
fix multiple define problem with ppc64le

clnperez:
building anything, e.g. envoy, that also builds gperftools, results in
the error that StacktracePowerPCDummyFunction was previously defined.

Rename this one, as its only a dummy function and is only used in this
one place.

Signed-off-by: Christy Norman <christy@linux.vnet.ibm.com>

### DIFF
--- a/absl/debugging/internal/stacktrace_powerpc-inl.inc
+++ b/absl/debugging/internal/stacktrace_powerpc-inl.inc
@@ -31,6 +31,8 @@
 #include <cstdint>
 #include <cstdio>
 
+#include "absl/base/attributes.h"
+#include "absl/base/optimization.h"
 #include "absl/base/port.h"
 #include "absl/debugging/stacktrace.h"
 #include "absl/debugging/internal/address_is_readable.h"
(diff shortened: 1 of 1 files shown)


absl/debugging/internal/stacktrace_powerpc-inl.inc
```diff
@@ -150,8 +150,8 @@ static void **NextStackFrame(void **old_sp, const void *uc) {
 }
 
 // This ensures that absl::GetStackTrace sets up the Link Register properly.
-void StacktracePowerPCDummyFunction() __attribute__((noinline));
-void StacktracePowerPCDummyFunction() { __asm__ volatile(""); }
+void AbslStacktracePowerPCDummyFunction() __attribute__((noinline));
```
derekmauro: Thank you very much for finding this!

I suspect we can do even better by marking this static and using the portable symbols we've added since this code was written. Something like:
```
ABSL_ATTRIBUTE_NOINLINE static void AbslStacktracePowerPCDummyFunction() {
  ABSL_BLOCK_TAIL_CALL_OPTIMIZATION();
}
```

I'll run this though the global presubmit tonight since we don't have ppc64le test coverage for OSS contributions.

clnperez: Sure. Let me know if there's anything else I can do!

clnperez:
This works as well (and thanks for catching that mistake in my patch!). :+1:

derekmauro:
There was no real mistake, I just wanted to modernize the code a bit before merging it. I'm going to merge as soon as the internal presubmits pass.

Thanks for your contribution!

clnperez:
Oh, oops. I thought I'd forgotten to change the actual function call from `StacktracePowerPCDummyFunction` to `AbslStacktracePowerPCDummyFunction` when I looked at your diff. n/m. :D

Issue #157, opened 2018-08-11
int128.h warnings with -Wpedantic


ojura:
When building Google Cartographer (which defines `-Wpedantic`) I get warnings like these if Abseil header directories are specified using `-I` instead of `-isystem`:

<details><summary>int128.h warnings</summary>
<p>

```
_deps/abseil-src/absl/numeric/int128.h:88:22: warning: ISO C++ does not support ‘__int128’ for ‘type name’ [-Wpedantic]
     alignas(unsigned __int128)
                      ^~~~~~~~
_deps/abseil-src/absl/numeric/int128.h:102:30: warning: ISO C++ does not support ‘__int128’ for ‘v’ [-Wpedantic]
   constexpr uint128(__int128 v);           // NOLINT(runtime/explicit)
                              ^
_deps/abseil-src/absl/numeric/int128.h:103:39: warning: ISO C++ does not support ‘__int128’ for ‘v’ [-Wpedantic]
   constexpr uint128(unsigned __int128 v);  // NOLINT(runtime/explicit)
                                       ^
_deps/abseil-src/absl/numeric/int128.h:117:31: warning: ISO C++ does not support ‘__int128’ for ‘v’ [-Wpedantic]
   uint128& operator=(__int128 v);
                               ^
_deps/abseil-src/absl/numeric/int128.h:118:40: warning: ISO C++ does not support ‘__int128’ for ‘v’ [-Wpedantic]
   uint128& operator=(unsigned __int128 v);
                                        ^
_deps/abseil-src/absl/numeric/int128.h:142:31: warning: ISO C++ does not support ‘__int128’ for ‘type name’ [-Wpedantic]
   constexpr explicit operator __int128() const;
                               ^~~~~~~~
_deps/abseil-src/absl/numeric/int128.h:142:31: warning: ISO C++ does not support ‘__int128’ for ‘type name’ [-Wpedantic]
_deps/abseil-src/absl/numeric/int128.h:143:40: warning: ISO C++ does not support ‘__int128’ for ‘type name’ [-Wpedantic]
   constexpr explicit operator unsigned __int128() const;
                                        ^~~~~~~~
_deps/abseil-src/absl/numeric/int128.h:143:40: warning: ISO C++ does not support ‘__int128’ for ‘type name’ [-Wpedantic]
_deps/abseil-src/absl/numeric/int128.h:256:57: warning: ISO C++ does not support ‘__int128’ for ‘type name’ [-Wpedantic]
[...]

PR #160, opened 2018-08-21
Add a CMake option for declaring interface headers as system headers 

ojura:
Related to #157.


JonathanDCohen:
Reading the conversation in #157, I don't think exporting all of our headers with -isystem is a smart idea, as we still want to see our own compiler warnings!  In particular, I'd need to see why our users (Cartographer in particular) can't drop -Wpedantic, as we never claimed to support that. If int128 is the only problem, maybe we can use a scalpel instead of a chainsaw: could we add a user define to avoid intrinsics in int128?

ojura:
> scalpel instead of a chainsaw

In that case it would be a shame not to use the intrinsics if they are available. If Cartographer wants to define `-Wpedantic` only for its own code and not for Abseil’s, I think the chainsaw approach is the way to go :-)

> as we still want to see our own compiler warnings

That’s why the option is disabled by default and opt-in, only for abseil users who want to treat their Abseil build tree as if it were installed.

Maybe this will become a non-issue since Cartographer guys mentioned they would be willing to drop `-Wpedantic`. @cschuet @gaschler @pifon2a?

Mizux:
Seems related to #111 to me, I means this is one of the issue of integrating abseil-cpp as `subdirectory()` instead of installed targets..., you force all cmake user to use your supported warning...
edit: well if cartographer use `target_compile_option(... PRIVATE -Wpendantic)` it should not impact abseil-cpp build...

Why don't using something like this instead:
```cmake
target_include_directories(absl... PUBLIC
  $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include/...>
)
target_include_directories(absl... SYSTEM PUBLIC
  $<INSTALL_INTERFACE:include/...>  # <prefix>/include/...
)
```
When building `abseil-cpp` you'll have the warning you want (impose), while **integrating** abseil-cpp in my **own** project (by first install it in my "workspace") I won't get warnings from abseil-cpp headers anymore (i.e. I flag them as -isystem). That's exactly why the [-isystem](https://gcc.gnu.org/onlinedocs/gcc-8.2.0/gcc/Directory-Options.html#Directory-Options) directory option  is for.

note: Yes, you'll keep the warning when build abseil-cpp standalone or integrating it using `add_subdirectory` but i think that's the implicit contract while using `add_subdirectory` in cmake

ojura:
> That’s why the option is disabled by default and opt-in, only for abseil users who want to treat their Abseil build tree as if it were installed.

Oops, it seems that I set the default to be `ON`, contrary to what I said here. Fixed.

ojura:
Can we move forward with this?

ojura:
@EricWF I have reworked this a bit to follow master and look a bit cleaner. Instead of branching on the `target_include_directories` call, we inject the `SYSTEM` keyword into the call if the option is set.

Jules0911 (approved):
Ok first time user so this is really alot

CMake/AbseilHelpers.cmake
```diff
@@ -107,6 +113,7 @@ function(absl_cc_library)
       add_library(${_NAME} STATIC "")
       target_sources(${_NAME} PRIVATE ${ABSL_CC_LIB_SRCS} ${ABSL_CC_LIB_HDRS})
       target_include_directories(${_NAME}
+        ${ABSL_DECLARE_PUBLIC_SYSTEM_HEADERS_KEYWORD}
         PUBLIC
           $<BUILD_INTERFACE:${ABSL_COMMON_INCLUDE_DIRS}>
           $<INSTALL_INTERFACE:${ABSL_INSTALL_INCLUDEDIR}>
```
lilianmoraru, line 116: This should be:
```cmake
      target_include_directories(${_NAME}
        PUBLIC
          $<BUILD_INTERFACE:${ABSL_COMMON_INCLUDE_DIRS}>
      )

      target_include_directories(${_NAME} ${ABSL_DECLARE_PUBLIC_SYSTEM_HEADERS_KEYWORD}
        PUBLIC
          $<INSTALL_INTERFACE:${ABSL_INSTALL_INCLUDEDIR}>
      )
```
The second `target_include_directories` is actually equivalent to `target_...(... SYSTEM INTERFACE ...)`
The same applies to the other `target_include_directories` lower.

This allows having warnings inside the Abseil project, but remove them for the users of the interface(for the actual users that can't modify the code and should not see the warnings).
With this change, the option can be removed and `SYSTEM` used directly.

ojura: @lilianmoraru if we are building Abseil as a subproject within a larger project (using e.g. CMake's `add_subdirectory`), we want to be able to silence these warnings as well. This is why `SYSTEM` should go into build interface as well if the user so desires.

Building as a subproject is strongly suggested in official docs:
https://github.com/abseil/abseil-cpp/blob/master/CMake/README.md

> For API/ABI compatibility reasons, we strongly recommend building Abseil in a subdirectory of your project or as an embedded dependency.

CMakeLists.txt
```diff
@@ -79,6 +79,8 @@ option(ABSL_USE_GOOGLETEST_HEAD
 
 option(ABSL_RUN_TESTS "If ON, Abseil tests will be run." OFF)
 
+option(ABSL_DECLARE_PUBLIC_SYSTEM_HEADERS "If ON, declare public Abseil headers as SYSTEM in order to mute compiler warnings." OFF)
```
lilianmoraru: What's your opinion on activating/disabling this automatically(still an option), by doing something like this?:
```cmake
if("${CMAKE_SOURCE_DIR}" STREQUAL "${CMAKE_CURRENT_SOURCE_DIR}")
    # Don't use system include paths in development
    option(ABSL_DECLARE_PUBLIC_SYSTEM_HEADERS "If ON, declare public Abseil headers as SYSTEM in order to mute compiler warnings." OFF)
else()
    # CMakeLists.txt was included as a subdirectory - disable warnings for the user
    option(ABSL_DECLARE_PUBLIC_SYSTEM_HEADERS "If ON, declare public Abseil headers as SYSTEM in order to mute compiler warnings." ON)
endif()
```

ojura: Nice idea! To avoid having `option` twice  I would set a variable in the if block (`ABSL_IS_SUBDIRECTORY`?) and use it as the default value in `option`.

Mizux: and adding it inside the block here ?
https://github.com/abseil/abseil-cpp/blob/e96ae2203b80d5ae2e0b7abe0c77b722b224b16d/CMakeLists.txt#L35-L41

ojura: Done.

ojura: @Mizux can this be moved forward?

Mizux:
Hi, 

First, sorry for the delay, shame on me !

I'm currently not part of Google (was a temp worker and in an other team see google/or-tools which have to depend on abseil ;) ...) and AFAIK @JonathanDCohen who was "in charge of CMake based build system" for the abseil Team just move to an other cool team (https://jigsaw.google.com/) so you should try to reach an other Abseil Team member in order to make your case...

ps: I'm in favor of the idea behind your change ;)

CMake/AbseilHelpers.cmake
```diff
@@ -110,6 +116,7 @@ function(absl_cc_library)
       add_library(${_NAME} "")
       target_sources(${_NAME} PRIVATE ${ABSL_CC_LIB_SRCS} ${ABSL_CC_LIB_HDRS})
       target_include_directories(${_NAME}
+        ${ABSL_DECLARE_PUBLIC_SYSTEM_HEADERS_KEYWORD}
```
Mizux: Proposal: since `ABSL_DECLARE_PUBLIC_SYSTEM_HEADERS` is only use here would it be possible to use
use a generator expression e.g. (not tested !!!)
```cmake
$<${ABSL_DECLARE_PUBLIC_SYSTEM_HEADERS}:"SYSTEM">
```
and remove lines 83-88 above...

CMake/AbseilHelpers.cmake
```diff
@@ -147,6 +154,7 @@ function(absl_cc_library)
       # Generating header-only library
       add_library(${_NAME} INTERFACE)
       target_include_directories(${_NAME}
+        ${ABSL_DECLARE_PUBLIC_SYSTEM_HEADERS_KEYWORD}
```
Mizux: same here

ojura:
@Mizux thanks for the suggestion, a generator expression is indeed more elegant :)
Pinging @EricWF.

CMakeLists.txt
```diff
@@ -32,14 +32,22 @@ cmake_policy(SET CMP0048 NEW)
...
 project(absl CXX)
 
-# when absl is included as subproject (i.e. using add_subdirectory(abseil-cpp))
-# in the source tree of a project that uses it, install rules are disabled.
+# detect if absl is the root CMake project, or is a subproject in the source tree
+# of a project that uses it (e.g. using add_subdirectory(abseil-cpp))
 if(NOT "^${CMAKE_SOURCE_DIR}$" STREQUAL "^${PROJECT_SOURCE_DIR}$")
+  # disable install rules
   set(ABSL_ENABLE_INSTALL FALSE)
+  set(ABSL_IS_CMAKE_SUBPROJECT TRUE)
```
JonathanDCohen, line 38: Looks like ABSL_ENABLE_INSTALL could just be killed here for logic in ABSL_IS_CMAKE_SUBPROJECT.  IIRC this only serves to make the install CMake rule visible, it doesn't control if Abseil is actually installed anywhere.

JonathanDCohen (commented):
Hey there @Mizux @EricWF !

It's true, I'm no longer on the Abseil team any more but I saw this fly by my inbox so I took a look.  I agree the generators are a better solution.  If I were still maintaining this code I'd be cool with it.

Incidentally, the project I moved to was [this](https://github.com/Jigsaw-Code/?q=outline), specifically

CMakeLists.txt
ojura: Sure! Cleaned up `ABSL_ENABLE_INSTALL.` I had to introduce the variable with the complementary boolean value because I could not do an inline negation of the value of `ABSL_ENABLE_INSTALL` when passing the default value for the `ABSL_DECLARE_PUBLIC_SYSTEM_HEADERS` option.

jnooree:
Any updates here?

Tunealb (commented):
💯

jnooree:
Might be resolved in #1249?

derekmauro:
I think this is fixed.

yasushi-saito:
poor mroo poo red p

ojura:
@derekmauro Sorry, not wanting to be rude, but why was this PR, which is doing the basically same thing as in #1249, left to rot for 4 years?

Anyway glad the thing is resolved now. You might still want to look at the diff of this PR because there are other things in there which I believe might be good to add.

PR #163, opened 2018-08-28
Add reverse_range helper function

miscco:
This adds a simple helper function, that redirects begin/end of a container to rbegin/rend.
It's main usage is for range based loops in reverse order where the following is equivalent:

vector<int> foo;
for (int elem : absl::reverse_range(foo)) {
    ...
}

for (auto rit = foo.rbegin(); rit!= foo.rend(); ++rit) {
    int elem = *rit;
    ...
}

Sadly std::rbegin/std::rend were only introduced with C++14 so i used the C++11 way

### DIFF
--- a/absl/algorithm/container.h
+++ b/absl/algorithm/container.h
@@ -69,6 +69,12 @@ using std::end;
 template <typename C>
 using ContainerIter = decltype(begin(std::declval<C&>()));
 
+// The type of the reverse_iterator given by rbegin(c) (possibly std::rbegin(c)).
+// ContainerReverseIter<const vector<T>> gives vector<T>::const_reverse_iterator,
+// while ContainerReverseIter<vector<T>> gives vector<T>::reverse_iterator.
+template <typename C>
+using ContainerReverseIter = std::reverse_iterator<ContainerIter<C>>;
+
 // An MSVC bug involving template parameter substitution requires us to use
 // decltype() here instead of just std::pair.
 template <typename C1, typename C2>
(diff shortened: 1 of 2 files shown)


miscco:
I signed it!

mbxx:
Thank you for the PR!  The people who are best suited to review this is out until next week, but I wanted to make sure you had a response and to offer a little background on Abseil's direction.

As you probably noticed, Abseil doesn't currently offer a ranges implementation.  That's because we're not clear on what ranges are going to look like when they land in the C++ standard.  Were we to offer a ranges implementation, we'd also want to have a good story for how we can seamlessly migrate our users to the standard version in the future.  (Much in the same way that absl::string_view and absl::optional have the same APIs as std::string_view and std::optional, and how they become typedefs for the standard types when they're available.)  Last I heard, the standards proposal for ranges was going to undergo some changes, and until we have a good idea what the C++ standard will look like, it's hard to plan a migration path.

So while this PR adds a small utility function, it touches on a larger topic that we're thinking about.  I apologize in advance if that means this isn't as simple for us as it might otherwise seem.

I'm curious, is your interest purely in having a `reverse_range` function or are you interested in ranges functionality more broadly?

miscco:
The resemblance with the one aspect of the ranges proposal is more or less coincidental. I just found, that the ability to use range based loops in reverse order improves code quality a lot, as that is a quite common pattern for me.

So while ranges are awesome, i am more interested in the `reverse_range` function

JonathanDCohen:
I would be more willing to accept the addition of absl::rbegin and absl::rend as C++17 shims then the addition of a random api which doesn't anticipate the standard in any way.

Do note, though, that the bar for entry for new API's in abseil is very high, especially ones which haven't baked for a while internally.

miscco:
Yeah, I simply didnt think it through in the context of the emerging ranges feature. So from my Point of view there are only 2. possibilities:

1. Just drop the PR, which is most likely the right thing to do

2. Given that it is a minimal API, add it and deprecate it once ranges are added to abeils/std. Although Hyrum might scream in agony here. I am not 100% sure i know what you ment with absl::rbegin shims?  Just adding simple struct and defining absl::begin and absl::end for them directly?

```c++
template <typename C>
struct reverse_range {
    explicit reverse_range (C& c) : data(c) {}
    C& data;
};

template <typename C>
ContainerReverseIter<C> absl::begin(reverse_range<C>& c) { return std::rbegin(c.data); }

template <typename C>
ContainerReverseIter<C> absl::end(reverse_range<C>& c) { return std::rend(c.data); }
```

JonathanDCohen:
Basically, I mean if we implemented https://en.cppreference.com/w/cpp/iterator/rbegin but in the absl:: namespace, and then it was an alias for std::rbegin when available, that would be something we'd probably be willing to accept.

miscco:
Ok I will have a look at it at the end of the week, as I am currently travelling.

lukamarek:
[abseil-cpp-master.zip](https://github.com/abseil/abseil-cpp/files/2379224/abseil-cpp-master.zip)

miscco:
Given that ranges are in C++20 there is obviously a better solution for this so i am closing this.

Thanks for the comments

PR #170, opened 2018-09-13
Add implementation of the C++14 feature of rbegin and rend

miscco:
Also add cbegin and cend

This came up in #163


aking77:
Please remove this from my site as I have been havked

miscco:
@JonathanDCohen  I have updated the PR to be more correct, as it really only tries to derive from `rbegin` rather than creating a `reverse_iterator` of `begin`.

I have also updated the tests to provide more coverage for less code

Unrelated: Great talk at CppCon.

zhangxy988:
Could you please briefly describe the reason why you want this in Abseil? Specifically, which category of priority does it fit into according to https://github.com/abseil/abseil-cpp/blob/master/CONTRIBUTING.md?

miscco:
In the already closed #163 I proposed an adaptor that inverses a range. This comes n handy when one might want to use a range based loop for loop. @JonathanDCohen pointed out a conflict with the coming ranges library so it was dropped. 

However in the conversation it came up that `begin` and `end` are defined in abseil but not `rbegin` and `rend`, which are equally usefull and generally often used. I would guess that if someone uses `absl::begin` he will quite likely also use `absl::rbegin`

It also feels like an inconsitency that in C++11 there is one but not the other

But quite frankly if that should be dropped i would not care too much

gennadiycivil:
@miscco Thank you again for this contribution. We are sill looking for the rationale why this should be in abseil. We have "priorities" defined in [contribution guidelines](https://github.com/abseil/abseil-cpp/blob/master/CONTRIBUTING.md#priorities) - could you please explain how this change fits?
Thank you again

miscco:
@gennadiycivil As far as I can tell any user of `absl::begin` will eventually try to iterate the container in the reverse direction. Consequently I would say this is a fix for an incompletely implemented API and register it as `Anticipated widespread usage` as it complements `absl::begin`

gennadiycivil:
@miscco Thank you for proposing this change. We had a discussion and decided that this change does not really fit the Abseil philosophy.

Abseil is fundamentally **not** a C++ compatibility backfill library

Thank you again

Issue #113, opened 2018-05-02
CMake Refactor: Remove Global Variables


JonathanDCohen:
Our CMakeLists.txt files have many global variables in the form of default sources, default library dependencies, or just lists of sources and libraries.  These are verbose and unneeded at best, or antipatterns where we could instead use INTERFACE targets at worst.  These should be cleaned up so we have a minimal amount of global variables polluting the CMake namespace. Any globals we can't get rid of should be prefixed with ABSL_

rongjiecomputer:
The number of global variables has been reduced to just a few that has `ABSL_` prefix, so this should not cause any trouble to CMake users. Time to close this issue?

PR #173, opened 2018-09-28
Fix CMake build

StephanDollberg:
Hi guys,

thanks for finally opensourcing Swissmap!

I just tried to build it using the CMake way but it seems that is currently broken as `absl::container` now needs to be a source library.

Fix for that is attached.

I have also enabled one of the hash map tests in the cmake tests to show the problem and avoid regressions. I can also enable the other tests as well if you want to (probably in another PR).

### DIFF
--- a/absl/container/CMakeLists.txt
+++ b/absl/container/CMakeLists.txt
@@ -163,3 +163,13 @@ absl_test(
   PUBLIC_LIBRARIES
     ${TEST_INSTANCE_TRACKER_TEST_PUBLIC_LIBRARIES}
 )
+
+
+absl_test(
+  TARGET
+    raw_hash_set_test
+  SOURCES
+    "internal/raw_hash_set_test.cc"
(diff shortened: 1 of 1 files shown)


absl/container/CMakeLists.txt
```diff
@@ -47,10 +47,16 @@ list(APPEND CONTAINER_INTERNAL_HEADERS
...
+list(APPEND CONTAINER_SRC
+  "internal/raw_hash_set.cc"
+)
 
-absl_header_library(
+absl_library(
   TARGET
     absl_container
+  SOURCES
+    ${CONTAINER_SRC}
```
rongjiecomputer: Please directly add `"internal/raw_hash_set.cc"` here instead of creating CMake global variable `CONTAINER_SRC`. Same goes to other CMake variables.

See #113

StephanDollberg: Everything inlined now.

PR #175, opened 2018-09-28
Mingw gcc doesn't define _MSC_VER, use _WIN32 instead.

sighingnow:


derekmauro:
Thank you for the pull request.

Sorry for not documenting this, but the code you are changing is actually part of another project that we automatically import. Can you please submit your pull request to https://github.com/google/cctz instead?

sighingnow:
See google/cctz#81.

PR #203, opened 2018-10-26
Use _umul128 on Windows to improve performance of Mix() on MSVC.

henrikedin:
I was doing performance measurements (http://bannalia.blogspot.com/2013/11/measuring-lookup-times-for-c-unordered.html as a base) and noticed a significant difference on the new swiss hash tables when compiled with MSVC compared to GCC/Clang on Linux. For simple integer keys find() took a little more than double the time. In practice absl ended up slower than std::unordered until approx. 0.5M elements.

Because MSVC is lacking __int128 intrinsic the code ended up doing uint128 * uint128 multiplication instead of uint64 * uint64 = uint128 which is the behavior we get with __int128 intrinsic on GCC/Clang.

Perhaps for clarity this code should always use _umul128 when available, but I haven't written that detection logic.

With this the performance on MSVC is on par with GCC and Clang in my testing.

### DIFF
--- a/absl/hash/internal/hash.h
+++ b/absl/hash/internal/hash.h
@@ -764,6 +764,7 @@ class CityHashState : public HashStateBase<CityHashState> {
   }
 
   ABSL_ATTRIBUTE_ALWAYS_INLINE static uint64_t Mix(uint64_t state, uint64_t v) {
+#if !defined(_MSC_VER) || !defined(_WIN64)
     using MultType =
         absl::conditional_t<sizeof(size_t) == 4, uint64_t, uint128>;
     // We do the addition in 64-bit space to make sure the 128-bit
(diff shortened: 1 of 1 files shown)


mattcalabrese-google:
Sorry for the delay and thank you for this. We are in the process of experimenting with an update to uint128 multiplication first.

astrelni:
Sorry for leaving this without an update for a while, this should actually have been improved as of https://github.com/abseil/abseil-cpp/commit/7b46e1d31a6b08b1c6da2a13e7b151a20446fa07. Can you please verify?

henrikedin:
Thank you. I ran my test and with that commit the performance is as good as with this PR.

astrelni:
Great, thanks a lot for spotting this.

PR #222, opened 2018-11-20
Ignore ctags files in .gitignore

RedBeard0531:
This is mostly because tools like ag and rg use .gitignore and it was annoying having the tags file pollute the results every time I grepped the codebase

Do you need a CLA for this?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,5 @@ build
 CMakeLists.txt.user
 # Ignore VS Code files
 .vscode/*
+# Ignore ctags files
+tags


ilyapopov:
Ignoring user-specific stuff should ideally be done in a user-wise [.gitignore](https://git-scm.com/docs/gitignore) (somewhere like `$HOME/.config/git/ignore`). `ripgrep` uses this this file (`ag` does not AFAIK). **Update:**  `ag` also uses the user-wide gitignore file.

derekmauro:
I agree with @ilyapopov on this one. Does that work for you @RedBeard0531?

RedBeard0531:
Sure. It's something I've seen in many others, but if you want to keep
yours minimal, I'm fine adding it to private a private config file.

PR #230, opened 2018-11-30
fix CMakeLists.txt in absl/container

lederernc:
absl_compressed_tuple is a header only library. use absl_cc_library properly to make the build work


lederernc:
I signed it!

lederernc:
I signed it!

lederernc:
this does not actually work. back to the drawing board

PR #238, opened 2018-12-21
Fix CMake build - missing hashtablez_sampler build

StephanDollberg:
Newly added file `hashtablez_sampler.cc` is not being built by the cmake
build and hence linking fails.

### DIFF
--- a/absl/container/CMakeLists.txt
+++ b/absl/container/CMakeLists.txt
@@ -22,6 +22,7 @@ absl_cc_library(
     container
   SRCS
     "internal/raw_hash_set.cc"
+    "internal/hashtablez_sampler.cc"
   COPTS
     ${ABSL_DEFAULT_COPTS}
   PUBLIC


absl/container/CMakeLists.txt
```diff
@@ -22,6 +22,7 @@ absl_cc_library(
     container
   SRCS
     "internal/raw_hash_set.cc"
+    "internal/hashtablez_sampler.cc"
```
derekmauro, line 25: See the comment a few lines up on lines 17-19. This target is deprecated and doesn't do anything anyway. I'd rather just remove this target.

StephanDollberg:
Yeah, I saw that but it's still the recommended way per the documentation
and I don't think one can expect people to read every single cmake file.

Personally, I also prefer having a single target but that's just me so I
would understand removing it but then the docs should be updated first.

I can do that as well though, if you want.

derekmauro:
Which documentation are you referring to?

StephanDollberg:
This one https://github.com/abseil/abseil-cpp/blob/master/CMake/README.md

StephanDollberg:
That one is also missing `absl::hash` I think.

derekmauro:
We are doing to have to fix that documentation. We are planning on generating the CMakeLists from Bazel Build files so there is going to be a one-to-one correspondence between targets soon.

StephanDollberg:
Ok, just to be clear, you want me to do that now and remove the `container` target or close this PR and wait for you autogenerate the CMakeLists?

derekmauro:
I'll consult with the team, but I think we are probably going to remove it. This probably won't be merged until Wednesday at the earliest due to the holidays.

PR #243, opened 2019-01-03
_umul128 intrinsic is x64 only and not available on Windows ARM64

ThomsonTan:
Intrinsic `_umul128` is only available on Windows X64, and not available on Windows ARM64, even both defined `_WIN64`.

This issue is exposed when building abseil library in Chromium for Windows ARM64.

EDIT: Resend the PR using the email account I registered with corporate CLA.

### DIFF
--- a/absl/numeric/int128.h
+++ b/absl/numeric/int128.h
@@ -43,10 +43,10 @@
 // builtin type.  We need to make sure not to define operator wchar_t()
 // alongside operator unsigned short() in these instances.
 #define ABSL_INTERNAL_WCHAR_T __wchar_t
-#if defined(_WIN64)
+#if defined(_M_X64)
 #include <intrin.h>
 #pragma intrinsic(_umul128)
-#endif  // defined(_WIN64)
+#endif  // defined(_M_X64)
 #else   // defined(_MSC_VER)
 #define ABSL_INTERNAL_WCHAR_T wchar_t
 #endif  // defined(_MSC_VER)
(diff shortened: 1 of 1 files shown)


astrelni:
Hi @ThomsonTan, thanks for the fix and sorry for the delayed response. Just a question before merging, are you able to give any context / verification on [this documentation for _umul128](https://docs.microsoft.com/en-us/cpp/intrinsics/umul128?view=vs-2017)? It seems to imply _umul128 availability on ARM.

ThomsonTan:
@astrelni this issue happens when building Chromium for Windows ARM64 which also defines `_WIN64` but doesn't define `_umul128`. According to the CRT header file `intrin.h/intrin0.h`, `_umul128` is only available on x64. I'll check with document team and reply it back here.

BTW: when could this change reach Chromium master?

ThomsonTan:
Hi @astrelni , confirmed that it is a document error for intrinsic `_umul128`. It actually doesn't support `ARM`. We'll fix it.

https://github.com/MicrosoftDocs/cpp-docs/pull/702

astrelni (approved):

astrelni:
@ThomsonTan 
Thanks for getting the documentation updated. Chromium forks abseil into a third_party directory, and they seem to update that about once a month. I will try to reach out and see if we can time the next one to include this patch or you can try to upstream it to chromium directly.

ThomsonTan:
Thanks @astrelni . Do I need look into the failure of Linux CI-gcc-opt build? I don't know how could my change affect that.

astrelni:
@ThomsonTan I think it's part of an issue kokoro itself is having right now, I don't think there's anything for you to do about it. I will hopefully get this merged early tomorrow.

PR #262, opened 2019-02-01
Add pragma alternate name support for Windows ARM64

ThomsonTan:
The extra declaration for `_M_IA64` and `_M_AMD64` than `_M_IX86` seems for `__ptr64` type, so it also applies to `_M_ARM64`.


atetubou:
It is better to send this to
https://github.com/google/cctz
instead?

derekmauro:
Please send this patch to https://github.com/google/cctz instead.

ThomsonTan:
Thanks, @atetubou and @derekmauro . The PR was merged to cctz project below. How could I port that to abseil-cpp?

https://github.com/google/cctz/pull/90

derekmauro:
@ThomsonTan - You don't have to do anything else, we will do it for you soon.

ThomsonTan:
@derekmauro it sounds great. When will you port it or update to latest cctz? Curious because Chromium needs this to unblock abseil-cpp unittest build for Windows ARM64.

shaindelschwartz:
Sorry for the delay - the team is offsite this week, so responses are slow. I'll push the latest Abseil now with an update to the latest CCTZ.

Issue #265, opened 2019-02-12
Https Everywhere


OlafvdSpek:
There are a lot of `http:` links in the Abseil files.. shouldn't these be `https:`?

JonathanDCohen:
This is low priority but would be nice for a good first issue.  We want to make sure all the links can actually support https.

SuchismitaDhal:
Hello @JonathanDCohen,
I am new to open source and would like to contribute to the project by making minor fixes. I have gone through the codebase. Can you help me with how to begin with this issue?

derekmauro:
@SuchismitaDhal This issue is about as straightforward as it gets. Find all the non-HTTPS links in the codebase. Check to see if the site supports HTTPS, and if it does, replace the HTTP link with an HTTPS link.

PR #270, opened 2019-03-05
Minor fix: changed HTTP URLs to HTTPS where possible

nik7273:
Issue 265: https://github.com/abseil/abseil-cpp/issues/265
Exhaustively searched and changed HTTP URLs to HTTPS where possible. The occasional HTTP still exists to links such as open-std.org because they don't yet support HTTPS.
This is done so that a secure connection is established when users click on URLs in the repository.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,7 +1,7 @@
 
                                  Apache License
                            Version 2.0, January 2004
-                        http://www.apache.org/licenses/
+                        https://www.apache.org/licenses/
 
    TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
 
@@ -193,7 +193,7 @@
    you may not use this file except in compliance with the License.
    You may obtain a copy of the License at
 
-       http://www.apache.org/licenses/LICENSE-2.0
+       https://www.apache.org/licenses/LICENSE-2.0
 
    Unless required by applicable law or agreed to in writing, software
    distributed under the License is distributed on an "AS IS" BASIS,
--- a/CMake/AbseilHelpers.cmake
+++ b/CMake/AbseilHelpers.cmake
@@ -5,7 +5,7 @@
 # you may not use this file except in compliance with the License.
 # You may obtain a copy of the License at
 #
-#    http://www.apache.org/licenses/LICENSE-2.0
+#    https://www.apache.org/licenses/LICENSE-2.0
 #
 # Unless required by applicable law or agreed to in writing, software
 # distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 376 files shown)


EricWF:
Awesome. Thank you for being thorough. 

Would you be willing to split out the changes to license headers? It would be easier to review.

nik7273:
Alright, will do.

EricWF:
This LGTM. But I'll defer to another Abseil member.

absl/time/internal/cctz/testdata/zoneinfo/iso3166.tab
```diff
@@ -10,7 +10,7 @@
 #
 # 1.  ISO 3166-1 alpha-2 country code, current as of
 #     ISO 3166-1 N905 (2016-11-15).  See: Updates on ISO 3166-1
-#     http://isotc.iso.org/livelink/livelink/Open/16944257
+#     https://isotc.iso.org/livelink/livelink/Open/16944257
```
derekmauro: This change needs to be reverted and submitted to https://github.com/google/cctz. (This is a common mistake. We should update our documentation.)

derekmauro: The same goes for all files under absl/time/internal/cctz. Sorry about this.

nik7273: Alright, I've done this and will submit to google/cctz soon.

derekmauro (approved):

LICENSE
```diff
@@ -1,7 +1,7 @@
 
                                  Apache License
                            Version 2.0, January 2004
-                        http://www.apache.org/licenses/
+                        https://www.apache.org/licenses/
```
Hixie, line 4: This change (and some others in this PR) affect the text of a license file. This seems unlikely to have been the intent. For example, the link here and below are supposed to point to the canonical version of this very license, and they do, and that canonical version does not use "https".

This causes downstream problems, for example, it means that Flutter's license script considers Abseil's Apache license to be non-canonical, causing us to include two copies of the Apache license in our license screen, one for Abseil files that refer to the canonical license directly in their header (and for everything else in Flutter that's covered by the Apache license), and one for Abseil files that don't have an explicit license statement and therefore end up covered by this file which differs from the canonical version by two bytes (the two "s" characters added by this PR).

derekmauro: We heard from numerous internal Google teams about the same issue. In hindsight, this change was not worth it, and it is great example of [Hyrum's Law](https://www.hyrumslaw.com/) in action.

At this point, this change is over two years old. We won't be changing it back, but we will use it as an example of why we need to avoid unnecessary churn in the future.

Hixie: Thanks for the acknowledgement.

nik7273: This is great feedback -- surprised I decided to do this back then but definitely useful information to know!

PR #274, opened 2019-03-08
implement Span::first and Span::last

### LINKED ISSUE #272: absl::Span::first, absl::Span::last
C++20 has `std::span::first` and `std::span::last`:
* https://en.cppreference.com/w/cpp/container/span/first
* https://en.cppreference.com/w/cpp/container/span/last

I've had couple cases already where those would have come in handy.

Would a PR that added those be welcome?

girtsf:
This implements `first` and `last` methods on `Span` that mimic the ones in `std::span`.

Fixes #272.

### DIFF
--- a/absl/types/span.h
+++ b/absl/types/span.h
@@ -485,6 +485,40 @@ class Span {
                : (base_internal::ThrowStdOutOfRange("pos > size()"), Span());
   }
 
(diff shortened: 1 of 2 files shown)


girtsf:
bot, let's try again.

Issue #127, opened 2018-05-23
thread_safety_analysis::ts_unchecked_read should be in the absl:: namespace


mbxx:
All of the symbols in Abseil were intended to be in the `absl::` namespace, but `thread_safety_analysis::ts_unchecked_read` was accidentally released outside of it.  It ought to be moved into `absl::base_internal::`.

CJ-Johnson:
Closing as the PR is closed

PR #275, opened 2019-03-09
moved thread_safety_analysis::ts_unchecked_read to absl namespace

serejkus:
This solves issue #127

### DIFF
--- a/absl/base/thread_annotations.h
+++ b/absl/base/thread_annotations.h
@@ -245,8 +245,9 @@
 // Disables warnings for a single read operation.  This can be used to avoid
 // warnings when it is known that the read is not actually involved in a race,
 // but the compiler cannot confirm that.
-#define TS_UNCHECKED_READ(x) thread_safety_analysis::ts_unchecked_read(x)
+#define TS_UNCHECKED_READ(x) absl::thread_safety_analysis::ts_unchecked_read(x)
 
+namespace absl {
 
 namespace thread_safety_analysis {
 
@@ -263,5 +264,7 @@ inline T& ts_unchecked_read(T& v) NO_THREAD_SAFETY_ANALYSIS {
 }
 
 }  // namespace thread_safety_analysis
+
+} // namespace absl
 
 #endif  // ABSL_BASE_THREAD_ANNOTATIONS_H_


serejkus:
I've signed CLA after receiving this message, so check should be restarted and label "cla: no" should be removed, I guess.

derekmauro (commented):
This isn't so simple.

From https://abseil.io/about/compatibility
> We will not break API compatibility. If we must, we will ship a tool to automate the upgrade to a preferred API. We will never break that API in a single change - we believe in non-atomic refactoring. We will introduce the new API and the tool, wait some time, and then remove the old.

This change would technically be an API break since `thread_safety_analysis::ts_unchecked_read()` is currently a public API according to our rules. It wasn't intended to be, but that was our mistake. I just verified there is at least one use in Google's codebase calling `thread_safety_analysis::ts_unchecked_read` directly (I will send a fix for that).

In addition, we also made the error of releasing `TS_UNCHECKED_READ` instead of `ABSL_TS_UNCHECKED_READ`. For example, [Chromium also has a `TS_UNCHECKED_READ` macro](https://chromium.googlesource.com/chromium/src/base/+/b7b147ba1083870cea43fe241c7c24e21abb9fe0/thread_annotations.h), and [macro redefinition](https://gcc.gnu.org/onlinedocs/cpp/Undefining-and-Redefining-Macros.html) is a potential problem in any project that may be using both definitions.

Fixing this is going to require more work.

serejkus:
So, what do we do here? Rename `TS_UNCHECKED_READ` to `ABSL_TS_UNCHECKED_READ`?
It is possible to make non-breaking change with something like this:
```c++
#define ABSL_TS_UNCHECKED_READ(x) absl::thread_safety_analysis::ts_unchecked_read(x)
// Deprecated. Use ABSL_TS_UNCHECKED_READ instead
#define TS_UNCHECKED_READ(x) ABSL_TS_UNCHECKED_READ(x)

namespace absl {
namespace thread_safety_analysis {
// ts_unchecked_read_here
}
}

namespace thread_safety_analysis {
  template<typename T>
  ABSL_DEPRECATED("use absl::thread_safety_analysis::ts_unchecked_read") inline const T& ts_unchecked_read(const T& v) NO_THREAD_SAFETY_ANALYSIS {
    return absl::thread_safety_analysis::ts_unchecked(v);
  }
  // ...
}
```

derekmauro:
You can't change the definition of `TS_UNCHECKED_READ` because of the macro redefinition issue. This is something the team is going to have to handle when we have time. We will need to release a tool to update code that depends on this API.

PR #276, opened 2019-03-10
Implement the Parallel Hashmap in Abseil

greg7mdp:
see https://greg7mdp.github.io/parallel-hashmap/ for writeup.

The new absl::parallel_flat_hash_map and absl::parallel_flat_hash_set have the same tests as the base maps and all tests pass on Windows (vs2017) and linux (g++ 5.4 and clang 3.8)


fowles:
Thanks for the writeup and the shoutout at the end of it.  For the sake of transparency, I will warn you that the bar to getting this added to absl will be very high.  Fundamental types require a very strong burden of proof.

In particular, you will need to measure things versus https://github.com/google/hashtable-benchmarks to verify the performance over a very broad range of things.  It is also worth realizing that `find()` performance in a variety of circumstances is one of the most important things for tables, so that should be the focus of much of your analysis.

greg7mdp:
Thanks Matt for the feedback. I believe the overhead for any operation to the `parallel` version vs the base `flat` version will be similar to the one for the insert (pretty small). but I can certainly understand that hard numbers will be more convincing :-).

I did run the https://github.com/google/hashtable-benchmarks during the development, but somehow I got stuck on viewing the results with the notebook. I'll try again.

However, considering the single-threaded `find()` performance kinds of defeats the purpose, as it *will* be slower than for the flat_hash_map. Not sure what conclusion we would be able to reach from that data point? 

Would it be OK to include the parallel_hash_map in Abseil if it is only slightly slower on `find()` than the flat_hash_map, considering it allows for lower peak memory usage and better multithreaded mutable operations?

greg7mdp:
@fowles , an example of an algo that can be made faster with the parallel_hash_map would be to extract the unique values from an unsorted container. 

If there is an algo of special interest to google currently using the flat_hash_map, and which doesn't do only `const` lookups, I'll be happy to see how it could be improved with the flat_hash_map if you can share it with me.

fowles:
The biggest barrier to admission into absl is proving general usefulness.  I suggest making a small repo that contains only this new class and then if usage of that appears common, you will have a strong argument for including it directly.

greg7mdp:
I am not sure I understand your suggestion. The `parallel_hash_map` relies directly on `absl::container_internal::raw_hash_set`, so I cannot make a repository containing only the new class.

I guess what I can do is create a new repository containing what I need from Abseil (mostly stuff from container-internal) + the parallel_hash_map. I'm just not sure how it would show anything useful.

fowles:
You can always create a branch that uses absl internals, the real distinction between "internal" things and external is that we make zero effort to preserve backwards compatibility for "internal" things.  For a prototype implementation like `parallel_hash_map` that is fine.

greg7mdp:
But isn't this what I exactly did? I created a fork and implemented the parallel_hash_map using `absl::container_internal::raw_hash_set`.

fowles:
Sorry, poor phrasing on my part.  Make a tiny repo that depends on absl and provides this class.  If it gets broad usage we can look at rolling it into absl proper.

greg7mdp:
Well, I can't really do that because, in order to avoid computing the hash twice, I had to make some small modifications to `absl::container_internal::raw_hash_set`, which is really the main thing I use from Abseil.

Since it sounds like you guys are not interested to take my PR at this point, I guess I'll have to create a separate repository for the `parallel_hash_map`.  I was thinking it would be a useful addition to Abseil, but clearly it is less obvious to you than it is to me :-). My bad.

fowles:
We are cautious around adding types because the support burden can be quite high

greg7mdp:
OK, no problem, I guess you can close the PR then if you don't want to take it.

bstaletic:
Just an idea, since I didn't look at the `parallel_hash_map` code, but can this container be turned into an adapter instead of being a full-blown container?

This would mean `parallel_hash_map` would be able to use any map with the STL API and could be self-contained in a separate repo.

Think of `std::queue` and the new `std::flat_map` and `std::flat_set`.

greg7mdp:
Yes, it would certainly be possible, and that was my first idea. Unfortunately, the std::unordered_map API doesn't allow to pass the precomputed hash value (except for find() in C++20), and I feel that computing the hash twice would be unacceptable except when the computation is trivial.

bstaletic:
That seems like an obstacle that could be avoided by using `is_invokable` and favoring the reusable hash algorithm if the underlying hash table supports it.

greg7mdp:
@bstaletic except that no hash table supports passing the hash value at this point. Even in c++20 it is available only for find(). I have implemented it for all APIs with some small changes in Google's raw_hash_map.

fowles:
Adding the C++20 apis to SwissTable is something we would very likely consider; however, you should not assume that different instances of `absl::Hash` will produce the same hash code for an item, which makes something like this surprisingly subtle

greg7mdp:
@fowles, indeed the SwissTable already supports `template<class K> iterator find(const K& x, size_t hash);` which is great, but not for insert, erase, emplace, …
> you should not assume that different instances of absl::Hash will produce the same hash code for an item

That is surprising! I saw that there was a random seeding at program startup, but then I would have assumed that different instances would match (actually if they didn't it would break my parallel_hash_map)

fowles:
They won't currently, but we don't guarantee that in perpetuity and we might need it for security reasons.  You probably want an HashRef that delegates to someone elses absl::Hash...

greg7mdp:
Actually, upon reflection, it will not be an issue for my implementation, provided I always pass the computed hash value to the underlying raw_hash_map (which indeed does require the extension of the API to take these precomputed hash values).

But thanks I had not been aware of that potential issue.

greg7mdp:
@fowles, also you may want to mention this fact in the Abseil documentation, since you are committed to API compatibility, and it would be hard to write a "code refactoring program"  that would deal with the fact that different instances of `absl::Hash<>` don't provide the same hash values.

fowles:
Patch coming to the docs soon

vslashg:
I wanted also to add that we appreciate the idea, but I agree it is highly unlikely that this API would clear the bar for Abseil.

We are very concerned with keeping our API surface small.  We want to only release the things that are heavily used.  Since we're are mostly open-sourcing existing code, we can use the existing internal usage as a metric for this.  For external contributions, gauging this is a lot harder.  We tend to be conservative about this.

We've attempted to describe our thinking in [CONTRIBUTING.md](https://github.com/abseil/abseil-cpp/blob/master/CONTRIBUTING.md).  We need to go through our other documentation to make sure we're calling enough attention to this elsewhere.  We tend to say "no" to requests (both internal and external!), but we want to be transparent about our decision-making motives.

> @bstaletic except that no hash table supports passing the hash value at this point. Even in c++20 it is available only for find().

There is a trick you can use here: any container that supports heterogeneous lookup (like the SwissTable containers do) can be used to cache hash values.

The `LazyInt` test case in flat_hash_map_test.cc shows the syntax for making a heterogeneous hash and equality objects.  The idea is you could provide a hasher and eq that worked on both T, and a struct containing a pointer to T and a pre-calculated hash.  This would allow you to implement your class in terms of SwissTable types, using only the publicly supported SwissTable API.  (Indeed, you could implement this in terms of any hash map that supports heterogeneous lookup.)

greg7mdp:
@vslashg thanks for the explanation, I certainly can see your side of things with respect to keeping the API small-ish. 
And many thanks for the lovely trick of bundling the hash value with the key, very cool indeed, I will look into it.

ROBOWORD:
. .

GOGOYAO:
> @vslashg thanks for the explanation, I certainly can see your side of things with respect to keeping the API small-ish. And many thanks for the lovely trick of bundling the hash value with the key, very cool indeed, I will look into it.

`bundling the hash value with the key` I think is not good when `heterogeneous lookup` is available, which means the key will be copied within `find` api. I think the performance is the best with precomputed hash value passing through, and this performance advantage will be achieved by hash functionity of cache too. I think the later may be a better way to implement  Parallel Hashmap. Thanks to @greg7mdp, your work is so wonderful. Thanks for all of you. This discussion is very useful.

greg7mdp:
Thanks you @GOGOYAO , I really appreciate your kind words, and I agree with you that bundling the hash with the key carries a size penalty which may not be acceptable.

GOGOYAO:
> Thanks you @GOGOYAO , I really appreciate your kind words, and I agree with you that bundling the hash with the key carries a size penalty which may not be acceptable.

Yes. Extra size and possible copy overhead is not necessary with precomputed hash or cached hash. Glad to see your work and this discussion.

I prefer cached hash because I prefer use the original absl library. I just want a dedicated implementation which is simple enough.

GOGOYAO:
And I am curious about why abseil did not publish a concurrent hash map. 
It is a popular container and we can see some implementation in folly, boost, tbb.
So why not publish one? After all, this pr has been denied for several years and no more about concurrent hash map has been made.
@vslashg Any idea?

derekmauro:
> And I am curious about why abseil did not publish a concurrent hash map.

Some of this is answered here: https://github.com/abseil/abseil-cpp/blob/master/FAQ.md

> It is a popular container and we can see some implementation in folly, boost, tbb.

I'm not sure why you want an Abseil implementation. It is not our goal to be the one library everyone uses. Why not just use the Folly one or the one by @greg7mdp?

I disagree that it is actually a popular container (and I have real usage data to backup that claim). `absl::flat_hash_map` is a general purpose container with an easy to use interface that works very well for most use cases. We have some private general purpose concurrent hash map implementations, but none of them have high usage, and all have more awkward interfaces.

In my experience, when you have a real world program where `absl::flat_hash_map` is a real bottleneck, there is little to gain by switching to another general purpose container. In situations where you can justify switching to another container, you probably have more to gain with a specialized solution that exploits usage patterns. For example, maybe you have a usage pattern in which an intrusive implementation is better. Or maybe lock-free reads are better for your usage pattern. A great example is the hash table implementation used by Stockfish - try dropping a general purpose concurrent hash map in there and tell me just how big of a performance regression that causes.

GOGOYAO:
Thank you for you explanation. As you saied, Abseil's interface, implementation and dependency is concise interface and is used widely. So I prefer Abseil than boost and folly.

PR #278, opened 2019-03-15
[CMake] Set correct flags for clang-cl

rongjiecomputer:
`clang-cl` produce binaries with MSVC ABI and wants to be as flag-compatible with pure MSVC as possible, so this leads to all sorts of weird cases.

- `clang-cl` alias `/Wall` as `clang`'s `-Weverything` which is way too verbose, so it needs `/W3` like pure MSVC.
- `clang-cl` only understand GCC style warning flags (`-W[no]blah`) and just silent drop MSVC style warning flags (`/wd[num]`).
- `clang-cl` needs MSVC define flags since it is consuming the same header files as pure MSVC.

CMake set `CMAKE_CXX_COMPILER_ID` as `Clang` when `clang-cl` is detected, so need extra `if (MSVC)` to differentiate it.

We are not doing `clang-cl` specialization in Bazel as currently there is no reliable way to detect `clang-cl` in Bazel..

This PR should be NFC for LLVM/GCC users on Unix platforms.

Other changes:

- Add `ABSL_` prefix to variable names to avoid name collision in CMake.


derekmauro:
Merged, thanks!

PR #286, opened 2019-03-24
support OpenBSD and Solaris 2.x (GCC/Clang)

despair86:
some of my projects use Abseil _and_ support various non-Linux systems.

- OpenBSD: clang v6 or later, any recent GCC (GCC v8 recently became available in the ports tree.)
- Solaris: gcc or clang only. (why anyone would still use the vendor compiler is beyond me.)

Is it possible to submit the required CLA _without_ having a google account? (Notice that my github account is tied to a personal mailbox.)

### DIFF
--- a/absl/base/config.h
+++ b/absl/base/config.h
@@ -85,11 +85,14 @@
 #endif
 
 // ABSL_HAVE_TLS is defined to 1 when __thread should be supported.
-// We assume __thread is supported on Linux when compiled with Clang or compiled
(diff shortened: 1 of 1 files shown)


derekmauro:
I'm a simple engineer, these CLAs frighten and confuse me too. I will have to ask someone who knows about licensing, and I will report back when I have an answer.

derekmauro:
We can't take this without a CLA. Sorry about this.

Feel free to re-open if you decide to create a Google account.

PR #298, opened 2019-04-17
Conan tag 20190416

uilianries:
As Conan exports the library list instead of using CMake targets, and Abseil can change the library names, we need to tag a specific version to list all correct   libraries.

When using all libraries together, a specific order must be respected. We have used --start-group solve any order problem, however this is not a good approach.

A new tag can be created monthly or when requested by issue.

/cc @lasote
related issue: https://github.com/conan-io/conan/issues/4928


gennadiycivil:
@uilianries Thank you for this contribution. Could you please provide build/test logs/evidence before/after to prove that this change indeed does what it intends to do. Abseil does not have Conan in the regular CI build matrix so we need to see that this "does not harm" and "works as intended"
Thank you !

uilianries:
Hi @gennadiycivil !

To validate this build using Conan, you could follow these steps:

    git clone https://github.com/uilianries/abseil-cpp.git
    cd abseil-cpp
    git checkout feature/conan-version
    docker run --rm -v ${PWD}:/home/conan/project conanio/gcc8 /bin/bash -c "cd project && conan create . abseil/testing -tf .conan/test_package" 

This Docker image contains Conan 1.17.1 + CMake 3.15 + GCC 8. It will build Abseil by running Conan. The version related in the Conan recipe is the latest commit available.

The folder .conan/test_package is a test to check if the generated Conan package is okay. It simulates an external project which consumes Abseil Conan package and build a minimal application. The test is only check the Conan package, not Abseil project.

You can read the full log here: [build.log](https://github.com/abseil/abseil-cpp/files/3427115/build.log)

suertreus:
I tried to validate this per your directions and got the following failure: [build_fail.log](https://github.com/abseil/abseil-cpp/files/3848903/build_fail.log)

I snuck a `find` command into the docker commandline to see where the missing file wound up; any idea what's wrong here?

suertreus:
This PR hasn't had any activity in many months, so I'm going to close it.  Please reopen or start a new PR as appropriate if further changes are needed.

PR #303, opened 2019-05-09
Support build on RHEL 5 Linux

xor2003:
I have been able to build Abseil on RHEL 5 and tests passing.
It has lower glibc version when specified minimal.
Also Futexes are not fully supported. I put a check to swtich them off.

### DIFF
--- a/absl/base/policy_checks.h
+++ b/absl/base/policy_checks.h
@@ -82,13 +82,10 @@
 // Standard Library Check
 // -----------------------------------------------------------------------------
 
-// We have chosen glibc 2.12 as the minimum as it was tagged for release
-// in May, 2010 and includes some functionality used in Google software
-// (for instance pthread_setname_np):
-// https://sourceware.org/ml/libc-alpha/2010-05/msg00000.html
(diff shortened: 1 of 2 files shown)


derekmauro:
https://abseil.io/about/compatibility

RHEL 5 is outside of our compatibility guidelines. While it may be possible to support, it has an engineering cost to ensure support, so we'd rather simply not support it.

RHEL 5 was released in 2007 and didn't even ship a C++11 compiler. In fact, according to distrowatch, neither did RHEL 6. You've clearly upgraded from the default compiler, maybe you could upgrade the glibc and kernel as well, or maybe you could do the easier thing and just upgrade the whole OS?

xor2003:
I understand your point. Unfortunately in my work I cannot throw away RHEL 5 right now.
My patches does not put any obligation to the library to support. I don't change the compatibility text.
Patches only add precision to futexes when they become useful. I mean no additional support for you and less patch support for me.
I thought Abseil is best library where I can already start to use modern C++ features like string_view and hash_map while our clients will slowly migrate from RHEL 5.

PR #311, opened 2019-05-16
Use the implicitly defaulted constructor

miscco:
Defining an empty default constructor is worse than using the compiler generated constructor.

Note that the compiler generated constructor requires that constexpr as well as exception specifiers are correctly considered. See Chapter 15.1 [class.ctor] paragraph 7

### DIFF
--- a/absl/container/internal/common.h
+++ b/absl/container/internal/common.h
@@ -55,7 +55,7 @@ class node_handle_base {
  public:
   using allocator_type = Alloc;
 
-  constexpr node_handle_base() {}
+  constexpr node_handle_base() = default;
   node_handle_base(node_handle_base&& other) noexcept {
     *this = std::move(other);
   }
@@ -114,7 +114,7 @@ class node_handle : public node_handle_base<PolicyTraits, Alloc> {
  public:
   using value_type = typename PolicyTraits::value_type;
 
-  constexpr node_handle() {}
+  constexpr node_handle() = default;
 
   value_type& value() const { return PolicyTraits::element(this->slot()); }
 
@@ -135,7 +135,7 @@ class node_handle<Policy, PolicyTraits, Alloc,
   using key_type = typename Policy::key_type;
   using mapped_type = typename Policy::mapped_type;
 
-  constexpr node_handle() {}
+  constexpr node_handle() = default;
 
   auto key() const -> decltype(PolicyTraits::key(this->slot())) {
     return PolicyTraits::key(this->slot());
--- a/absl/strings/internal/char_map.h
+++ b/absl/strings/internal/char_map.h
@@ -32,17 +32,17 @@ namespace strings_internal {
 
 class Charmap {
  public:
-  constexpr Charmap() : m_() {}
+  constexpr Charmap() = default;
 
   // Initializes with a given char*.  Note that NUL is not treated as
   // a terminator, but rather a char to be flicked.
-  Charmap(const char* str, int len) : m_() {
+  Charmap(const char* str, int len) {
     while (len--) SetChar(*str++);
   }
 
   // Initializes with a given char*.  NUL is treated as a terminator
   // and will not be in the charmap.
-  explicit Charmap(const char* str) : m_() {
+  explicit Charmap(const char* str) {
     while (*str) SetChar(*str++);
   }
 
@@ -127,7 +127,7 @@ class Charmap {
     m_[c / 64] |= static_cast<uint64_t>(1) << (c % 64);
   }
 
-  uint64_t m_[4];
+  uint64_t m_[4] = {};
 };
 
 // Mirror the char-classifying predicates in <cctype>
--- a/absl/container/flat_hash_map.h
+++ b/absl/container/flat_hash_map.h
@@ -150,7 +150,7 @@ class flat_hash_map : public absl::container_internal::raw_hash_map<
   //
   //   std::vector<std::pair<int, std::string>> v = {{1, "a"}, {2, "b"}};
   //   absl::flat_hash_map<int, std::string> map7(v.begin(), v.end());
-  flat_hash_map() {}
+  flat_hash_map() = default;
   using Base::Base;
 
   // flat_hash_map::begin()
--- a/absl/container/flat_hash_set.h
+++ b/absl/container/flat_hash_set.h
@@ -143,7 +143,7 @@ class flat_hash_set
   //
   //   std::vector<std::string> v = {"a", "b"};
   //   absl::flat_hash_set<std::string> set7(v.begin(), v.end());
-  flat_hash_set() {}
+  flat_hash_set() = default;
   using Base::Base;
 
   // flat_hash_set::begin()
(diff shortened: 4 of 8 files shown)


absl/container/internal/common.h
```diff
@@ -55,7 +55,7 @@ class node_handle_base {
  public:
   using allocator_type = Alloc;
 
-  constexpr node_handle_base() {}
+  node_handle_base() = default;
```
asoffer: I think we want to keep `constexpr` on these, rather than let it be implicitly constexpr for the reasons you mentioned, but I'm unsure. As for the others, I'm also unsure, but @EricWF may have more information about whether this is okay.

absl/strings/internal/char_map.h
```diff
@@ -32,7 +32,7 @@ namespace strings_internal {
 
 class Charmap {
  public:
-  constexpr Charmap() : m_() {}
+  Charmap() = default;
```
asoffer: `m_` is an array whose default construction is trivial, whereas `m_()` will do value-initialization setting it to zero. 

https://godbolt.org/z/zhZjwr

But perhaps `uint64_t m_[4] = {};` would enable this change.

EricWF: Apart from gaining the correct exception specification, (which we could simply add), I'm not sure this change provides any benefit.

absl/container/internal/common.h
EricWF: I agree that we want to keep the `constexpr`.

We should only omit `constexpr` when we want it to be deduced. When we require it, we should write it explicitly. Then the compiler will (hopefully) tell us right away if our expectation is violated.

miscco: I would also agree, for me it is a deficiency in the standard, that `constexpr foo() = default;` is not possible. I will try to write a paper for that

absl/strings/internal/char_map.h
miscco: Quite frankly it depends.

1. You have to do work to get the same behavior 
2. As far as I know virtually everyone and their mother agree that member initialization if preferable to initialization to a default value in the constructor.
3. It is more consitent. There are a lot of places in abseil where implicitely defaulted member functions are used.

@asoffer You are correct. However, I am not sure whether member initialization is valid in C++11 mode?

derekmauro:
I don't know whether `constexpr` should be removed, but if you do end up removing `constexpr`, be sure to add a test that constexpr construction is still possible.

absl/strings/internal/char_map.h
asoffer: Member initialization is valid but I think we still want `constexpr` default cosntructors and so `= default` is probably not what we want. I agree with your sentiment that this is a deficiency in the standard and would love to see your paper.

miscco: Hm I started to write it and idway through it i thought about actually reading the relevant pages of the standard. (Perfect ordering...)

It seems that I get a pass this time because of Section 11.4.2 Explicitly-defaulted functions [dcl.fct.def.default] paragraph 2:

> An explicitly-defaulted function that is not deﬁned as deleted may be declared constexpr only if it would have been implicitly declared as constexpr. If a function is explicitly defaulted on its ﬁrst declaration, it is implicitly considered to be constexpr if the implicit declaration would be.

So according to the standard we should already be able to use `constexpr foo() = default;` I am 95% sure that i tried it and ended up with all hell breaking loose but I will try it again.

miscco:
I kept the constexpr forthe node_handles and added default initialization for the Charmap

miscco:
@asoffer The last kokoro run had some failure could you ran that againnow that `constexpr `is still in place

EricWF:
Here's my grip with `= default` for class templates. As someone who reads and writes a lot of generic code, a user defined constructor effectively communicates its semantics to the reader. For example:

`T() {}` means `T` is default constructible, it may throw, and it is not `constexpr`.
`T() = default;` means `T` may or may not be default constructible. It may or may not throw exceptions. And it may or may not be constexpr.

miscco:
@EricWF There is indeed a certain ambiguity. However I cannot follow you on

> `T() = default;` means `T` may or may not be default constructible.  

If `T` would not be default constructible then `T() {}` would be ill formed?! Is there any possibility that `T() = default;` does not work whereas `T() {}` does? I am aware that this is the case for move assignment/constructor (which feels like a deficiency of the standard) but I am quite sure this is not the case for the default constructor.

For the ambiguity on the exception specification and `constexpr` I understand the frustration but isnt it a strict improvement over `T() {}`?

If it would be `noexcept` you would have missed that opportunity to improve performance. I am not deep enough in the standard to fully evaluate the ramifications but I would guess that things like containers of `T` and aggregate with members of type `T` might suffer some problems as the (wrong?) exception specification propagates up the class hierarchy

The same on `constexpr` albeit with `constexpr T() = default;` I do not see such a problem here. Given the advent of `constexpr` for `std::vector` in C++20 it is quite concievable that `absl::flat_hash_set` might be used in a compile time context in the future. 

Is there any benefit on loosing out on these optimizations.

zhangxy988:
It is unclear to me what the intent of this change is. Is it for improve performance or readability?
From the previous conversation, it seems that `T() {}` is easier to read than `T() = default`.

miscco:
There are some benefits.

1. The exception specification is inferred 
2. `constexpr` is inferred. (or enforced by`constexpr T() = default;`)

Consequently writing `T(){}` is a pessimization that might prevent compiler optimizations.

zhangxy988:
I don't see strong reasons to either change or not change this. Let me take your patch and run it against our internal code base to see if there is any real impact.

miscco:
I would also blame Greg Falcon for this:

https://youtu.be/6ZOygaUjzjQ?t=1515

zhangxy988:
On a second thought, I think I still prefer `T() {} ` more than `T() = default`, because `T() {}` conveys the information very clearly: you don't need to look at the members of `T` to determine whether T is default constructible, nothrow, or constexpr. Later if we want to add a new data member, it won't accidentally change the API. This is even more important when T is a template type, because it's even harder to see through.
However, I do agree that for simple cases like a trivial class, you should probably just write `T() =default` so you don't make it not trivially default constructible.

miscco:
Actually it is the opposite.

1. T will always be non-trivially constructible albeit it could
2. T will never be marked as noexcept albeit it could
3. T will not be constexpr albeit it could

I understand that it is more explicit but at the same time the more explicit version is less correct than the implicit one

zhangxy988:
For 2 and 3, I would prefer `constexpr T() noexcept {}`.
For 1, which types are actually "losing" trivial default constructibility now?

miscco:
For 1 a class with `T(){}` is not `trivially_default_constructible` even if the same class with `T() = default;` would be. See your godbolt example in the corresponding issue

For 2 it is more complex as the exception specification depends on T so you would have to write some elaborate conditionals. In the implicit case the compiler does that for you.

For 3. You can also write `constexpr T() = default;` which was done for those types where constexpr should be enforced.

zhangxy988:
My point is, whether you write `T() = default ` or not is just an implementation technique to achieve the API requirements.  The real question is, for each type you are changing, can it/ should it be trivially default constructible/ noexcept / constexpr? After we agree on the answer to this (design) question,  we can easily figure out the best implementation to achieve that.

gennadiycivil:
@miscco Thank you so much for this proposal and the detailed discussion. Looks like the team did not find enough benefit in the change and therefore we decided not to accept the change. Thank you again and we are looking forward to continued collaboration in the future.

PR #316, opened 2019-05-21
Fix #315 when null ptr is passed to memcpy with size 0 

### LINKED ISSUE #315: passing empty string_view to StrCat results in run-time error with ubsan enabled
The failure is here (str_cat.cc:92): 
```
static char* Append(char* out, const AlphaNum& x) {
  // memcpy is allowed to overwrite arbitrary memory, so doing this after the
  // call would force an extra fetch of x.size().
  char* after = out + x.size();
  memcpy(out, x.data(), x.size());
  return after;
}
```
apparently we can not pass null with zero size to memcpy and friends - it's UB. 
memcpy's declaration:
`extern void *memcpy (void *__restrict __dest, const void *__restrict __src,
		     size_t __n) __THROW __nonnull ((1, 2));`

See here for more info:
https://www.imperialviolet.org/2016/06/26/nonnull.html

romange:
Fix #315 when null ptr is passed to memcpy with size 0

### DIFF
--- a/absl/strings/str_cat.cc
+++ b/absl/strings/str_cat.cc
@@ -89,7 +89,7 @@ static char* Append(char* out, const AlphaNum& x) {
   // memcpy is allowed to overwrite arbitrary memory, so doing this after the
   // call would force an extra fetch of x.size().
   char* after = out + x.size();
-  memcpy(out, x.data(), x.size());
+  if (x.size()) memcpy(out, x.data(), x.size());
   return after;
 }
 
--- a/absl/strings/str_cat.cc
+++ b/absl/strings/str_cat.cc
@@ -89,7 +89,7 @@ static char* Append(char* out, const AlphaNum& x) {
   // memcpy is allowed to overwrite arbitrary memory, so doing this after the
   // call would force an extra fetch of x.size().
   char* after = out + x.size();
-  if (ABSL_PREDICT_TRUE(x.size() != 0)) {
+  if (x.size() != 0) {
     memcpy(out, x.data(), x.size());
   }
   return after;
@@ -148,7 +148,7 @@ std::string CatPieces(std::initializer_list<absl::string_view> pieces) {
   char* out = begin;
   for (const absl::string_view piece : pieces) {
     const size_t this_size = piece.size();
-    if (ABSL_PREDICT_TRUE(this_size != 0)) {
+    if (this_size != 0) {
       memcpy(out, piece.data(), this_size);
       out += this_size;
     }
@@ -180,7 +180,7 @@ void AppendPieces(std::string* dest,
   char* out = begin + old_size;
   for (const absl::string_view piece : pieces) {
     const size_t this_size = piece.size();
-    if (ABSL_PREDICT_TRUE(this_size != 0)) {
+    if (this_size != 0) {
       memcpy(out, piece.data(), this_size);
       out += this_size;
     }
--- a/absl/strings/str_cat.cc
+++ b/absl/strings/str_cat.cc
@@ -89,7 +89,7 @@ static char* Append(char* out, const AlphaNum& x) {
   // memcpy is allowed to overwrite arbitrary memory, so doing this after the
   // call would force an extra fetch of x.size().
   char* after = out + x.size();
-  if (x.size() != 0) {
+  if (ABSL_PREDICT_TRUE(x.size() != 0)) {
     memcpy(out, x.data(), x.size());
   }
   return after;
@@ -148,7 +148,7 @@ std::string CatPieces(std::initializer_list<absl::string_view> pieces) {
   char* out = begin;
   for (const absl::string_view piece : pieces) {
     const size_t this_size = piece.size();
-    if (this_size != 0) {
+    if (ABSL_PREDICT_TRUE(this_size != 0)) {
       memcpy(out, piece.data(), this_size);
       out += this_size;
     }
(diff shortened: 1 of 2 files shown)


absl/strings/str_cat.cc
```diff
@@ -89,7 +89,7 @@ static char* Append(char* out, const AlphaNum& x) {
   // memcpy is allowed to overwrite arbitrary memory, so doing this after the
   // call would force an extra fetch of x.size().
   char* after = out + x.size();
-  memcpy(out, x.data(), x.size());
+  if (x.size()) memcpy(out, x.data(), x.size());
```
derekmauro: Please use !x.empty(), and please use curly braces around the if statement.

derekmauro: Can you add a test that would hit this behavior?

romange: I can easily demonstrate it when compiling with  `-fsanitize=undefined ` flag and in that case ubsan just terminates the process. I am not sure it's possible to uncover the bug within the current capabilities of gtest framework without ubsan.

derekmauro: You don't have to demonstrate that the process gets terminated. In fact, you want the opposite. I'd like to see something that demonstrates that the process doesn't get terminated under the conditions in this bug. If someone ever removes the branch and runs the test under ubsan, I'd like to know something is wrong.

romange: Hey Derek, I am sorry, I do not understand what you mean. The bug is that with specific compiler options UB sanitizer terminates the process. I am not aware of any other manifestation of the bug.
To demonstrate it all you need is to call `StrCat(string_view{})`.

tituswinters:
The bug is that it's UB to invoke memcpy on a null pointer. The
manifestation is that ubsan points that out in some cases.

We'd like to add tests to help demonstrate that we *aren't* hitting that
UB, which means those codepaths need to be exercised.

romange:
Hi @tituswinters , thanks for the reply. I will gladly add the test that covers the UB path. What I am missing  is how to make it fail in case the `if` branch is removed. I do not know of any way to assert externally that memcpy call does receive null argument.

romange:
specifically this test 
```
TEST_F(StrCatTest, UB) {
  StrCat(2, absl::string_view{});
}
```
fails with the error below with ubsan but unfortunately I do not know how I can catch it differently.
```
../abseil-cpp/absl/strings/str_cat.cc:92:9: runtime error: null pointer passed as argument 2, which is declared to never be null
    #0 0x55e01bfa54dd in Append ../abseil-cpp/absl/strings/str_cat.cc:92
    #1 0x55e01bfa56fc in absl::StrCat[abi:cxx11](absl::AlphaNum const&, absl::AlphaNum const&) ../abseil-cpp/absl/strings/str_cat.cc:103
    #2 0x55e01bef0494 in StrCatTest_Empty_Test::TestBody() ../strings/strcat_test.cc:20
```

tituswinters:
You don't need to assert it, you just need to add the test - we'll make
sure ubsan runs regularly enough that if there's UB lurking, it'll get
surfaced. :)

romange:
Oohh :) 
image

absl/strings/str_cat.cc
romange: @derekmauro  Fixed style comments.  please note AlphaNum does not have `.empty()`.

romange:
PTAL

romange:
I verified the test coverage using the following build configuration:
`cmake -L -DCMAKE_BUILD_TYPE=Debug -DCMAKE_CXX_FLAGS_DEBUG="-fsanitize=undefined" -DABSL_USE_GOOGLETEST_HEAD=ON -DABSL_RUN_TESTS=ON`

absl/strings/str_cat_test.cc
```diff
@@ -408,6 +408,18 @@ TEST(StrCat, VectorBoolReferenceTypes) {
   EXPECT_EQ(result, "1010");
 }
 
+TEST(StrCat, UBCheck) {
```
derekmauro: Can you give this test case a more descriptive name than just `UBCheck`? Maybe `AvoidsMemcpyWithNullptr`? You should also put a comment explaining that passing `nullptr` to `memcpy` is undefined behavior and that test is to catch that bug.

romange: Sure, fixed. I also added ABSL_PREDICT_TRUE to mitigate possible performance degradation.

romange:
btw, from looking at `AppendPieces` and `CatPieces`, it's seems that `CatPieces` can be replaced by or implemented using `AppendPieces`.

absl/strings/str_cat.cc
```diff
@@ -89,7 +89,7 @@ static char* Append(char* out, const AlphaNum& x) {
   // memcpy is allowed to overwrite arbitrary memory, so doing this after the
   // call would force an extra fetch of x.size().
   char* after = out + x.size();
-  if (x.size() != 0) {
+  if (ABSL_PREDICT_TRUE(x.size() != 0)) {
```
derekmauro: My gut reaction is that `ABSL_PREDICT_TRUE` is a premature optimization here. Passing an empty string is a legitimate use of `absl::StrCat`. We also have no evidence yet that the branch predictor is getting this wrong. We also use profile-guided optimization in production binaries. This use of `ABSL_PREDICT_TRUE` will prevent the branch predictor from doing its job.

romange:
Passing the empty string is a legitimate case...
This bug has been here for a long time hinting that it's relatively rare
case - I am guessing i am not the first one using ubsan on abseil code.
Anyway, i will revert those changes...

romange:
PTAL

derekmauro (approved):

PR #319, opened 2019-05-24
AlphaNum adds a char32_t constructor to support rune

fcharlie:
When using absl::StrCat, we can add more interesting features, such as using it to connect to rune. When we know the Unicode code point of a character, we can easily connect it to the string.

Code Example:

```cpp
#include <absl/strings/str_cat.h>
#include <absl/strings/str_format.h>

int main() {
  char32_t sparkling_heart = 0x1F496;
  auto s = absl::StrCat("I have a ", sparkling_heart, " !");
  absl::FPrintF(stderr, "Tom say: %s\n", s);
  return 0;
}

```
Snapshot:
2019-05-24 09-40-01屏幕截图

### DIFF
--- a/absl/strings/str_cat.cc
+++ b/absl/strings/str_cat.cc
@@ -21,6 +21,7 @@
 
 #include "absl/strings/ascii.h"
 #include "absl/strings/internal/resize_uninitialized.h"
+#include "absl/strings/internal/utf8.h"
 
 namespace absl {
 
@@ -74,6 +75,11 @@ AlphaNum::AlphaNum(Dec dec) {
   }
 
   piece_ = absl::string_view(writer, end - writer);
+}
+
+AlphaNum::AlphaNum(char32_t rune) {
+  piece_ = absl::string_view(
+      digits_, absl::strings_internal::EncodeUTF8Char(digits_, rune));
 }
 
 // ----------------------------------------------------------------------
--- a/absl/strings/str_cat.h
+++ b/absl/strings/str_cat.h
@@ -235,8 +235,9 @@ class AlphaNum {
   AlphaNum(double f)  // NOLINT(runtime/explicit)
       : piece_(digits_, numbers_internal::SixDigitsToBuffer(f, digits_)) {}
 
-  AlphaNum(Hex hex);  // NOLINT(runtime/explicit)
-  AlphaNum(Dec dec);  // NOLINT(runtime/explicit)
+  AlphaNum(Hex hex);        // NOLINT(runtime/explicit)
+  AlphaNum(Dec dec);        // NOLINT(runtime/explicit)
+  AlphaNum(char32_t rune);  // NOTLINT(runtime/explicit)
 
   template <size_t size>
   AlphaNum(  // NOLINT(runtime/explicit)


fcharlie:
I signed it!

fcharlie:
I signed it!

EricWF:
This seems weird to me. This doesn't work for `char`, Why should it work for `char32_t`. And should it work for `char8_t` or `char16_t`?

For example:
```
char c = 'c';
// AlphaNum(char c)' has been explicitly marked deleted here
cout << StrCat("ab", c);
```

In general I think treating a single character as being convertible to a string of a single character is a bug.  Character types too easily promote to `int`.

fcharlie:
@EricWF  `char` does not take effect because the overload of char is actually deleted: 
https://github.com/abseil/abseil-cpp/blob/4adaf5490921f13028b55018c9f550277de5aebb/absl/strings/str_cat.h#L255

However, in the Arg class of absl::Substitute, there are many contributors to absl, and different people do not know the preference. I also have some doubts here.

https://github.com/abseil/abseil-cpp/blob/4adaf5490921f13028b55018c9f550277de5aebb/absl/strings/substitute.h#L118-L120

`char16_t` A single character may not be converted to UTF-8 (<0xD800), so there is no overload added here, in bela (a library that learns abseil, the pure wchar_t version on Windows overrides char16_t)

https://github.com/fcharlie/bela/blob/c1b83c5922ad1d08ed2d03cae07ffe9c2b608681/include/bela/strcat.hpp#L199-L208

```cpp
  AlphaNum(wchar_t c) : piece_(digits_, 1) { digits_[0] = c; }
  AlphaNum(char16_t ch) : piece_(digits_, 1) {
    digits_[0] = static_cast<wchar_t>(ch);
  }
  AlphaNum(char32_t ch)
      : piece_(digits_,
               bela::char32tochar16(reinterpret_cast<char16_t *>(digits_),
                                    numbers_internal::kFastToBufferSize, ch)) {}
```

tituswinters:
I think that Eric's question was, "We've explicitly *not* done this for
`char` because it will be confusing (do you get the numeric value or the
ASCII value), what is the reasoning for allowing wider chars but not `char`
directly?"

fcharlie:
@tituswinters Whether it's char or char16_t or wchar_t on Windows, a variable of one length can't hold all Unicode characters, but char32_t can hold all Unicode characters for a long time. char32_t and uint32_t have the same width, signedness, and alignment, but they are not the same type, so I think here char32_t should be converted to a UTF-8 string instead of a decimal number.

https://en.cppreference.com/w/cpp/language/types

tituswinters:
Sure, but StrCat is explicitly *not* in the business of supporting printing
characters: it handles strings and numerics so as to avoid questions of
when a character type should be interpreted as a string or as an int. The
fact that we *could* support a single character type (out of many)
unambiguously doesn't mean that it's *right* for the API and ecosystem to
do so.

(I'm not saying it *isn't* either, I'm just saying that this isn't making
the argument clear for why this choice is the right answer, what other
design options would achieve the same thing without changing previous
precedent, etc.)

fcharlie:
@tituswinters Of course, your statement is correct. Adding support for char32_t is not indispensable, it is just a icing on the cake.

mbxx:
Thank you so much for using Abseil, and for caring enough to offer this pull request.

The implementation looks good to me, but as @tituswinters and @EricWF mentioned, StrCat specifically avoids printing character types.   I think it would be very surprising for StrCat to handle some character types but not others.

Abseil also generally tries to avoid the subject of Unicode whenever possible.  There are really good Unicode libraries out there (ICU comes to mind), and we don't want to put out an inferior, partial implementation of their features.

I hate to say this because I really appreciate the pull request and I can see how this change can be useful to some folks, but I'm afraid I don't think this change is right for Abseil.

PR #323, opened 2019-06-10
Fix for CMake 3.6

gosnik:


gennadiycivil:
Thank your for this contribution. Could you please elaborate and provide more information on exactly the issue / problems this is trying to address.
What is the before/after effect this would create and how is this beneficial?

gosnik:
When integrating Abseil as a CMake subproject in to an Android Studio NDK project - it did not build. Android Studio uses CMake 3.6. VERSION_GREATER_EQUAL is not a valid comparison for CMake 3.6. There is only VERSION_LESS up until CMake 3.7
This change performs the same check while also working for CMake projects with CMake < v3.7.

EricWF:
Thanks for the patch. This LGTM. I'll integrate this tomorrow.

Issue #218, opened 2018-11-14
[GCC 8.2.1, CMake] absl::ByChar::Find link error when absl::string_view is std::string_view


chenxiaolong:
It seems that when `absl::string_view` is `std::string_view` (i.e. targeting C++17), then `absl::ByChar::Find` fails to link properly.

```
[  2%] Built target absl_internal_spinlock_wait
[  3%] Built target absl_dynamic_annotations
[ 10%] Built target absl_base
[ 12%] Built target absl_internal_throw_delegate
[ 25%] Built target absl_strings
[ 26%] Building CXX object CMakeFiles/main.dir/main.cpp.o
[ 27%] Linking CXX executable main
/usr/bin/ld: CMakeFiles/main.dir/main.cpp.o: in function `absl::strings_internal::SplitIterator<absl::strings_internal::Splitter<absl::ByChar, absl::AllowEmpty> >::operator++()':
main.cpp:(.text._ZN4absl16strings_internal13SplitIteratorINS0_8SplitterINS_6ByCharENS_10AllowEmptyEEEEppEv[_ZN4absl16strings_internal13SplitIteratorINS0_8SplitterINS_6ByCharENS_10AllowEmptyEEEEppEv]+0x65): undefined reference to `absl::ByChar::Find(std::basic_string_view<char, std::char_traits<char> >, unsigned long) const'
collect2: error: ld returned 1 exit status
make[2]: *** [CMakeFiles/main.dir/build.make:89: main] Error 1
make[1]: *** [CMakeFiles/Makefile2:77: CMakeFiles/main.dir/all] Error 2
make: *** [Makefile:84: all] Error 2
```

If I target C++14 instead, which causes abseil's own string_view implementation to be used, then everything links just fine. I'm using GCC 8.2.1 on Fedora 29 and building with CMake.

---

I created a very simple project to reproduce this: https://github.com/chenxiaolong/abseil_link_error_test

To test:
```
git clone --recursive https://github.com/chenxiaolong/abseil_link_error_test.git
cd abseil_link_error_test
mkdir build
cd build
cmake ..
make
```

JonathanDCohen:
Thanks for the report!  We're currently migrating all of our cmakelists to use absl_cC_library and ape our bazel targets, so we'll get this when we get to strings/

SlavaMelanko:
Hi @JonathanDCohen 
Do you have any updates?

derekmauro:
[...]

PR #330, opened 2019-06-17
Fixes for conan + support for specifying C++ standard in conanfile

kylemacfarlan:
1) The conanfile.py did not define a version, so attempting to create a conan package would fail. I added the version label of "latest" in an attempt to mimic the "live at head" philosophy of abseil (ideally, this would be a CI thing where a new commit would automatically generate a new conan package, but I don't think we're there yet). LTS branches should change this version to match the LTS version number

2) Added an option to specify the C++ standard version when creating the conan package. When using the current conan package to link against a project in C++17 mode, linker errors could occur. #218 is an example of what can happen if abseil is built with C++11 or C++14 support, but attempted to link against a c++ 17 project. 

I know google doesn't use conan, so it's difficult for you all to evaluate conan changes, so I've tagged some of the bincrafters team who have worked on the bincrafters version of an abseil package (I hope the tagging works--I'm not sure I did it right). 
@uilianries 
@SSE4
@danimtb
@solvingj

### DIFF
--- a/conanfile.py
+++ b/conanfile.py
@@ -22,7 +22,10 @@ class AbseilConan(ConanFile):
     exports_sources = ["CMakeLists.txt", "CMake/*", "absl/*"]
     generators = "cmake"
     settings = "os", "arch", "compiler", "build_type"
-
+    options = {"cxx_standard": {11, 14, 17}}
+    default_options = {"cxx_standard": 11}
+    version = "latest"
+    
     def configure(self):
         if self.settings.os == "Windows" and \
            self.settings.compiler == "Visual Studio" and \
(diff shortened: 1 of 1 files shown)


conanfile.py
```diff
@@ -22,17 +22,21 @@ class AbseilConan(ConanFile):
...
            self.settings.compiler == "Visual Studio" and \
            Version(self.settings.compiler.version.value) < "14":
             raise ConanInvalidConfiguration("Abseil does not support MSVC < 14")
 
     def build(self):
-        tools.replace_in_file("CMakeLists.txt", "project(absl)", "project(absl)\ninclude(conanbuildinfo.cmake)\nconan_basic_setup()")
+        tools.replace_in_file("CMakeLists.txt", "project(absl CXX)", "project(absl CXX)\ninclude(conanbuildinfo.cmake)\nconan_basic_setup()")
         cmake = CMake(self)
         cmake.definitions["BUILD_TESTING"] = False
+        cmake.definitions["CMAKE_CXX_STANDARD"] = self.options.cxx_standard
```
uilianries, line 39: you could use https://docs.conan.io/en/latest/howtos/manage_cpp_standard.html

kylemacfarlan:
Ah, I didn't know about that feature. That is the better solution. I'll
delete the pull request then. Thanks!

PR #336, opened 2019-06-20
Improve example AbslUnparseFlag().

shields:
1. Move default case outside switch.  This is a good practice because it allows -Wswitch to warn about missing enum values.

2. Use absl::StrCat() instead of (nonexistent) SimpleItoa().


EricWF (commented):
The change looks good, but the commit message is out of date.

LGTM

shields:
The commit message was correct, but I forgot to actually make the change.

shaindelschwartz (approved):

PR #338, opened 2019-06-24
Candidate fix for issue #337

### LINKED ISSUE #337: UBSan violation in InlineVector when NDEBUG isn't defined
```
    absl::inlined_vector<int, 2> foo;
    absl::inlined_vector<int, 2> bar{1,2,3};
    foo = bar;
```

leads to
```
external/com_google_absl/absl/container/internal/inlined_vector.h:61:15: runtime error: null pointer passed as argument 1, which is declared to never be null
/usr/include/string.h:62:62: note: nonnull attribute specified here
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior external/com_google_absl/absl/container/internal/inlined_vector.h:61:15 in
```

More specifically, 
https://github.com/abseil/abseil-cpp/blob/e9324d926a9189e222741fce6e676f0944661a72/absl/container/internal/inlined_vector.h#L47
gets called with 
```
alloc_ptr = <some address>, 
destroy_first = nullptr,
destroy_size = 0
```

leading to call to memset with `nullptr` as starting address in 

https://github.com/abseil/abseil-cpp/blob/e9324d926a9189e222741fce6e676f0944661a72/absl/container/internal/inlined_vector.h#L61

DarkDimius:
There seem to be cases when `DestroyElements` gets called for `0` elements with `nullptr` as `destroy_first`.
Currently, this triggers ubsan failure in debug builds-only(optimized builds define NDEBUG and thus don't observe ASAN failure).

### DIFF
--- a/absl/container/inlined_vector_test.cc
+++ b/absl/container/inlined_vector_test.cc
@@ -1787,4 +1787,16 @@ TEST(InlinedVectorTest, AbslHashValueWorks) {
   EXPECT_TRUE(absl::VerifyTypeImplementsAbslHashCorrectly(cases));
 }
 
+struct HasVector {
+  absl::InlinedVector<int, 1> foo;
+};
+
+// This code used to trigger memset with `nullptr` as starting address in builds that do not have NDEBUG set.
+// This is undefined behavior with UBSan.
+TEST(CopyAssignmentOperator, AvoidsMemsetWithNullptr) {
+    HasVector foo;
+    HasVector bar;
+    bar.foo.push_back(2);
+    foo = bar;
+}
 }  // anonymous namespace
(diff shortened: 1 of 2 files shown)


DarkDimius:
I've confirmed that added test used to fail with the following message:
```
external/com_google_absl/absl/container/internal/inlined_vector.h:61:15: runtime error: null pointer passed as argument 1, which is declared to never be null
/usr/include/string.h:62:62: note: nonnull attribute specified here
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior external/com_google_absl/absl/container/internal/inlined_vector.h:61:15 in
```

before the patch if compiled with `-fsanitize=undefined -g3` and no `-O.*` flags.

After this patch test succeeds.

CJ-Johnson:
Addressed internally. Will upstream soon...

PR #340, opened 2019-06-25
Fixed ABSL_INTERNAL_MACOS_CXX17_TYPES_UNAVAILABLE check

jtsylve:
The `__MAC_OS_X_VERSION_MIN_REQUIRED__` macro no longer seems to be present on my system (macOS 10.14.5) with the latest xcode command line tools.  This PR fixes the check to allow proper detection of the unavailable C++17 types with Xcode 10.

You can test this by compiling with MACOSX_DEPLOYMENT_TARGET=10.13 environment variable on a 10.14 system.

### DIFF
--- a/absl/base/config.h
+++ b/absl/base/config.h
@@ -375,7 +375,7 @@
 // https://github.com/abseil/abseil-cpp/issues/207 and
 // https://developer.apple.com/documentation/xcode_release_notes/xcode_10_release_notes
 #if defined(__APPLE__) && defined(_LIBCPP_VERSION) && \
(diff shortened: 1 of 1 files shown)


ddovod:
Probably not the best place to mention, but it would be very useful to have similar check for iOS cause iOS 12 acts just like macOS 10.14 in terms of "unavailable" libc++ functionality.

derekmauro:
@mattcalabrese-google - Can you look into merging this? The CI failures are for the new tests I added that weren't available at this commit, but should be fine at head.

jtsylve:
I just rebased this against head, so let me know if that fixes the CI checks

derekmauro:
CI passes now, thanks.

PR #341, opened 2019-06-25
Fix ABSL_HAVE_ALARM check on mingw

jtsylve:
mingw compilers don't support alarm()

### DIFF
--- a/absl/base/config.h
+++ b/absl/base/config.h
@@ -334,6 +334,8 @@
 #define ABSL_HAVE_ALARM 1
 #elif defined(_MSC_VER)
 // feature tests for Microsoft's library
+#elif defined(__MINGW32__)
+// feature tests for mingw
 #elif defined(__EMSCRIPTEN__)
 // emscripten doesn't support signals
 #elif defined(__native_client__)
--- a/absl/base/config.h
+++ b/absl/base/config.h
@@ -335,7 +335,10 @@
 #elif defined(_MSC_VER)
 // feature tests for Microsoft's library
 #elif defined(__MINGW32__)
-// feature tests for mingw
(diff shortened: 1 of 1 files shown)


EricWF:
This looks good to me.
But please be aware that  we dont have test coverage on MinGW (nor is it officially supported).

Regardless, thank you for the patch. I'll merge it once I verify it internally.

jtsylve:
rebased against master

absl/base/config.h
```diff
@@ -334,6 +334,8 @@
 #define ABSL_HAVE_ALARM 1
 #elif defined(_MSC_VER)
 // feature tests for Microsoft's library
+#elif defined(__MINGW32__)
+// feature tests for mingw
```
suertreus: Please explain in this comment why `ABSL_HAVE_ALARM` is not defined for `__MINGW32__`.  I realize you've copied the form of the preceding comments, but they're not very helpful and I'd like to do better.  The `__EMSCRIPTEN__` example below is better.

It looks to me like the right comment is:
```
// mingw32 doesn't provide alarm(2):
// https://osdn.net/projects/mingw/scm/git/mingw-org-wsl/blobs/5.2-trunk/mingwrt/include/unistd.h
// mingw-w64 provides a no-op implementation:
// https://sourceforge.net/p/mingw-w64/mingw-w64/ci/master/tree/mingw-w64-crt/misc/alarm.c
```

suertreus (changes requested):

shaindelschwartz:
@jtsylve Would you like to follow up here? There's an open review comment that should be resolved before we merge this.

suertreus (approved):

PR #342, opened 2019-06-25
Fix ABSL_WAITER_MODE detection for mingw

jtsylve:
The synchronization APIs used in ABSL_WAITER_MODE_WIN32 are only available starting in Windows Vista.  By default, mingw sets `_WIN32_WINNT` to Windows XP causing compilation to fail.  This PR addresses this issue by adding an additional check.

### DIFF
--- a/absl/synchronization/internal/waiter.h
+++ b/absl/synchronization/internal/waiter.h
@@ -19,9 +19,9 @@
 #include "absl/base/config.h"
 
 #ifdef _WIN32
-#include <pthread.h>
-#else
 #include <SdkDdkVer.h>
+#else
+#include <pthread.h>
 #endif
 
 #ifdef ABSL_HAVE_SEMAPHORE_H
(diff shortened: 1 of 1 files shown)


jtsylve:
rebased against master

147901 (approved):

absl/synchronization/internal/waiter.h
```diff
@@ -40,7 +40,7 @@
 
 #if defined(ABSL_FORCE_WAITER_MODE)
 #define ABSL_WAITER_MODE ABSL_FORCE_WAITER_MODE
-#elif defined(_WIN32)
+#elif defined(_WIN32) && _WIN32_WINNT >= 0x0600
```
suertreus: How would you feel about comparing against `_WIN32_WINNT_VISTA` instead of the magic constant?  Looks like you'd have to conditionally `#include <SdkDdkVer.h>`.

If that's not possible or desirable, please add a comment indicating that `0x0600` corresponds to Vista.

suertreus (changes requested):

shaindelschwartz:
@jtsylve Would you like to follow up here? There's an open review comment that should be resolved before we merge this.

suertreus (approved):

Issue #356, opened 2019-07-31
Compilation failed using CMake and C++17 standard


HenryRLee:
This is what I did after pulling the latest code
```
 $  mkdir build
 $  cd build/
 $  cmake .. -DABSL_RUN_TESTS=ON -DABSL_USE_GOOGLETEST_HEAD=ON -DCMAKE_CXX_STANDARD=17
 $  make
```

And the error log is:
```
Scanning dependencies of target absl_variant_exception_safety_test
[ 95%] Building CXX object absl/types/CMakeFiles/absl_variant_exception_safety_test.dir/variant_exception_safety_test.cc.o
In file included from /Users/lee/Studio/git/abseil-cpp/absl/types/variant_exception_safety_test.cc:15:
In file included from /Users/lee/Studio/git/abseil-cpp/absl/types/variant.h:50:
/Library/Developer/CommandLineTools/usr/include/c++/v1/variant:472:5: error: static_assert failed due
      to requirement '__all<is_same_v<ThrowingBool (*)(__value_visitor<equal_to<void> > &&, const
      __base<std::__1::__variant_detail::_Trait::_Available,
      ThrowingValue<testing::TypeSpec::kEverythingThrows>,
      ThrowingValue<testing::TypeSpec::kNoThrowCopy>, ThrowingValue<testing::TypeSpec::kNoThrowMove>,
      vector<ThrowingValue<testing::TypeSpec::kEverythingThrows>,
      ThrowingAllocator<ThrowingValue<testing::TypeSpec::kEverythingThrows>,
      testing::AllocSpec::kEverythingThrows> > > &, const
      __base<std::__1::__variant_detail::_Trait::_Available,
      ThrowingValue<testing::TypeSpec::kEverythingThrows>,
      ThrowingValue<testing::TypeSpec::kNoThrowCopy>, ThrowingValue<testing::TypeSpec::kNoThrowMove>,
      vector<ThrowingValue<testing::TypeSpec::kEverythingThrows>,
      ThrowingAllocator<ThrowingValue<testing::TypeSpec::kEverythingThrows>,
      testing::AllocSpec::kEverythingThrows> > > &), ThrowingBool (*)(__value_visitor<equal_to<void>
      > &&, const __base<std::__1::__variant_detail::_Trait::_Available,
      ThrowingValue<testing::TypeSpec::kEverythingThrows>,
[...]

PR #360, opened 2019-08-02
Ignore the test case variant_exception_safety_test in AppleClang C++17 #356

HenryRLee:


CJ-Johnson:
kokoro:run

zhangxy988:
Thanks for reporting this issue. However, before we adopt this workaround, we'd like to know about the issue. Could you please share more details on the failure?  Specifically, which test cases are failing and what toolchain are you using,  e.g. compiler version, standard library? This way we can reproduce the error and identify the root cause. Thanks!

HenryRLee:
@zhangxy988 Hi, thanks for the reply. I created an issue as #356 and linked this PR to it.

The failing test case is `variant_exception_safety_test`. My compiler version is `Apple LLVM version 10.0.1`, and my standard library is LLVM.

PR #372, opened 2019-08-22
supppress unused variable warning for recent gcc

pizzard:
gcc7.2 complains with: 
```
'stacktraces_enabler' defined but not used [-Wunused-variable] 
static int stacktraces_enabler = []() { 
```


zhangxy988:
LGTM. Thanks for the PR. I will run a quick sanity check internally and merge this.

PR #376, opened 2019-09-10
Add instructions to install Abseil using vcpkg

vicroms:
Provides instructions to build Abseil using [vcpkg](https://github.com/Microsoft/vcpkg) with a single command, provided that vcpkg is already installed in the user's machine.

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,6 +51,16 @@ more information on building Abseil using the Bazel build system.
 If you require CMake support, please check the
 [CMake build instructions](CMake/README.md).
 
+Alternatively, if you're using the [vcpkg](https://github.com/Microsoft/vcpkg) dependency manager, then you can install Abseil with a single command:
+
+```
+vcpkg install abseil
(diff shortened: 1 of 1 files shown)


vicroms:
@googlebot I signed it!

CJ-Johnson:
@vicroms Thank you for signing the CLA!

W/r/t the change, we aren't opposed to it in principle. That said, before we include such an endorsement in our documentation, we'd like to understand why vcpkg has local changes to Abseil. Abseil intends to fully support MSVC, so if there are problems with our code, we'd prefer to upstream the solutions. (Ex: https://github.com/microsoft/vcpkg/blob/master/ports/abseil/fix-config.patch)

I'd be happy to continue this conversation over email or via Github comments.

Thanks! - CJ

derekmauro:
Closing this. We don't recommend pre-compiled libraries and consider vcpkg unofficial.

PR #387, opened 2019-10-07
s/ParseFlag/AbslParseFlag

HunterLarco:
[ Context ]

`absl::Duration` + `absl::Time` are unusable as a `ABSL_FLAG` because
the Parse/Unparse function hooks are named incorrectly. Perhaps this
was caused by naming changes over time? In either case, I've added the
corrent hooks as per https://abseil.io/docs/cpp/guides/flags#custom

Note that I haven't remove or changed the already existing Parse/Unparse
functions so that any existing clients that use these do not break as a
result of this change.

### DIFF
--- a/absl/time/duration.cc
+++ b/absl/time/duration.cc
@@ -910,6 +910,12 @@ bool ParseFlag(const std::string& text, Duration* dst, std::string* ) {
   return ParseDuration(text, dst);
 }
 
+bool AbslParseFlag(absl::string_view text, Duration* dst, std::string* error) {
+  return ParseFlag(std::string(text), dst, error);
+}
+
 std::string UnparseFlag(Duration d) { return FormatDuration(d); }
+
+std::string AbslUnparseFlag(const Duration& d) { return UnparseFlag(d); }
 
 }  // namespace absl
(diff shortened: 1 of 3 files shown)


derekmauro:
@rogeeff - What's going on here? It looks like these functions already exist in the internal code, but you've purposely excluded them from being exported.

HunterLarco:
I don't understand your comment @derekmauro ?

These functions do not currently exist in the abseil repo? This is supported by a quick

`git grep AbslParseFlag | grep Duration`

in addition to the fact that

`ABSL_FLAG(foo, absl::Duration, ...);`

Does not compile.

rogeeff:
I'm looking into this. It appears that we were changing time library flag parsing APIs in stages and final stage was never checked in. We already have the correct code internally, but it is not being exported

rogeeff:
I'm releasing the fix. Should be out tomorrow.

asoffer:
Hunter,

Thank you for bringing this to our attention. the most recent push should address this issue. We've left the old definitions in place for the time being but will be cleaning them up soon. Please re-open if you have any further issues.

PR #393, opened 2019-10-23
Detect appropriate HWAES flags for gcc builds

### LINKED ISSUE #365: Abseil fails to compile on ARM with CMake, because it forces x86_64 flags
`absl/copts/AbseilConfigureCopts.cmake` checks for GCC `"${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU"` and then forcibly adds `msse`: `set(ABSL_RANDOM_RANDEN_COPTS "${ABSL_RANDOM_HWAES_X64_FLAGS}")`, even though it does not compile on x86_64.
That last line adds this:
```
list(APPEND ABSL_RANDOM_HWAES_X64_FLAGS
    "-maes"
    "-msse4.1"
)
```

Note that the ARM flags(from the generated file with compiler flags) also makes some assumptions.
`-mfpu=neon` is buggy with the version of the compiler we have and the hardware we run the software on.

amrox:
Fixes: #365

If the compiler was detected as gcc, the CMake scripts would
automatically apply the hardware acceleration flags for x86_64. This
merely checks the CMAKE_SYSTEM_PROCESSOR variable and applies the
appropriate flags for the target system.


asoffer:
Thanks for this, looks good! We have a change that landed internally a few days ago and will be pushed  with our next release. I believe that what you have here will be an added improvement on top of that, so once that gets pushed to GitHub, I'll pull this change in as well.

lilianmoraru:
For some reason, "CMAKE_SYSTEM_PROCESSOR" for me is always unset(x86_64 or cross-compiling to ARM with a Toolchain file).

amrox:
@lilianmoraru Interesting. It seems to be set when I compile natively on macOS, and it's also set by the Emscripten and arm64 toolchains I use. What toolchain are you using for cross compiling?

amrox:
@asoffer Happy to help, and open to feedback. I also noticed that the `ABSL_RANDOM_HWAES_X64_FLAGS` is set explicitly here too https://github.com/abseil/abseil-cpp/blob/19b021cb3ff23048dfbe236a4e611925d8930831/absl/copts/AbseilConfigureCopts.cmake#L23 so that may need to be updated.

amrox:
It looks like a more robust fixed landed in commit https://github.com/abseil/abseil-cpp/commit/078b89b3c046d230ef3ad39494e5852184eb528b#diff-34f9b8e4d402381f2a234c311b7fbca6R8

Its working for me, so I'm going to go ahead and close this PR.

PR #576, opened 2019-12-30
Adding small implementation in ascii.h and escaping.cc under the string folder

jameskim0987:
I added a minuscule implementation in ascii.h and escaping.cc. 
Thanks.

### DIFF
--- a/absl/strings/ascii.h
+++ b/absl/strings/ascii.h
@@ -153,7 +153,7 @@ inline bool ascii_islower(unsigned char c) { return c >= 'a' && c <= 'z'; }
 // ascii_isascii()
 //
 // Determines whether the given character is ASCII.
-inline bool ascii_isascii(unsigned char c) { return c < 128; }
+inline bool ascii_isascii(unsigned char c) { return c >= 0 && c < 128; }
 
 // ascii_tolower()
 //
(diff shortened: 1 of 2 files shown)


absl/strings/ascii.h
```diff
@@ -153,7 +153,7 @@ inline bool ascii_islower(unsigned char c) { return c >= 'a' && c <= 'z'; }
 // ascii_isascii()
 //
 // Determines whether the given character is ASCII.
-inline bool ascii_isascii(unsigned char c) { return c < 128; }
+inline bool ascii_isascii(unsigned char c) { return c >= 0 && c < 128; }
```
derekmauro, line 156: What is the purpose of this change? `c >= 0` is trivially true for an unsigned variable.

absl/strings/escaping.cc
```diff
@@ -41,6 +41,11 @@ constexpr bool kUnescapeNulls = false;
 
 inline bool is_octal_digit(char c) { return ('0' <= c) && (c <= '7'); }
 
+inline bool is_octal_number(int num) {
```
derekmauro, line 44: It is meaningless to ask if an integer is an octal number. All integers have an octal representation.

This is also a change to a `.cc` file that isn't exposed in any header, so this new function wouldn't be usable anyway.

derekmauro (changes requested):
I'm not sure what you are trying to do, but neither of these changes are acceptable.

PR #586, opened 2020-01-05
Use https links.

nlewycky:


absl/base/macros.h
```diff
@@ -112,7 +112,7 @@ ABSL_NAMESPACE_END
 // when  performing switch labels fall-through diagnostic
 // (`-Wimplicit-fallthrough`). See clang documentation on language extensions
 // for details:
-// http://clang.llvm.org/docs/AttributeReference.html#fallthrough-clang-fallthrough
+// https://clang.llvm.org/docs/AttributeReference.html#fallthrough-clang-fallthrough
```
derekmauro, line 115: This appears to be a valid webpage, but it is devoid of any meaningful content (for both HTTP and HTTPS). I wonder if LLVM accidentally broke their document export process. Thankfully I know a few people I can ask.

nlewycky: I noticed that too and filed https://llvm.org/PR44463. LLVM regenerates docs nightly, so I figured it would just fix itself automatically at some point. I decided to go ahead with the PR since it doesn't change the fact that the links should be HTTPS.

abc52090241:
I also seem to have the question of http ask how to change it (Sorry I'm still a novice)

absl/base/macros.h
derekmauro: The docs look fixed now. I will merge this soon.

derekmauro (approved):

PR #594, opened 2020-01-13
UWP doesn't allow reading regkeys

### LINKED ISSUE #593: BuildBreak: sysinfo.cc uses unsupported APIs on UWP apps
UWP doesn't allow reading regkeys. Unfortunatelly, UWP also doesn't offer an API for returning nominal processor frequency at this moment. Other options would require apps depending on abseil-cpp to be packaged with extra manifest data or libraries for bridging platforms.

LoadLibrary:
UWP doesn't allow reading regkeys. Unfortunately, UWP also doesn't offer an API for returning nominal processor frequency at this moment. Other options would require apps depending on abseil-cpp to be packaged with extra manifest data or libraries for bridging platforms.

This change pushes the unsupported APIs accessing the registry behind a define guard. This define guard makes GetNominalCPUFrequency to compile and run as usual on desktop, but it will return the value 1.0 on UWP Apps (Store).

Fixes: #593

### DIFF
--- a/absl/base/internal/sysinfo.cc
+++ b/absl/base/internal/sysinfo.cc
@@ -76,7 +76,12 @@ static int GetNumCPUs() {
 #if defined(_WIN32)
 
 static double GetNominalCPUFrequency() {
-#if defined(WINAPI_FAMILY) && (WINAPI_FAMILY != WINAPI_FAMILY_APP)
+// UWP apps don't have access to the registry and currently don't provide an
+// API informing about CPU nominal frequency.
+#if WINAPI_FAMILY_PARTITION(WINAPI_PARTITION_APP) && \
+    !WINAPI_FAMILY_PARTITION(WINAPI_PARTITION_DESKTOP)
+  return 1.0;
+#else
 #pragma comment(lib, "advapi32.lib")  // For Reg* functions.
   HKEY key;
   // Use the Reg* functions rather than the SH functions because shlwapi.dll
@@ -95,8 +100,9 @@ static double GetNominalCPUFrequency() {
       return data * 1e6;  // Value is MHz.
     }
   }
-#endif
   return 1.0;
+#endif /* WINAPI_FAMILY_PARTITION(WINAPI_PARTITION_APP) && \
+          !WINAPI_FAMILY_PARTITION(WINAPI_PARTITION_DESKTOP) */
 }
 
 #elif defined(CTL_HW) && defined(HW_CPU_FREQ)
--- a/absl/base/internal/sysinfo.cc
+++ b/absl/base/internal/sysinfo.cc
@@ -76,6 +76,7 @@ static int GetNumCPUs() {
 #if defined(_WIN32)
 
 static double GetNominalCPUFrequency() {
+#if defined(WINAPI_FAMILY) && (WINAPI_FAMILY != WINAPI_FAMILY_APP)
 #pragma comment(lib, "advapi32.lib")  // For Reg* functions.
   HKEY key;
   // Use the Reg* functions rather than the SH functions because shlwapi.dll
@@ -94,6 +95,7 @@ static double GetNominalCPUFrequency() {
       return data * 1e6;  // Value is MHz.
     }
   }
+#endif
   return 1.0;
 }
 
--- a/absl/base/internal/sysinfo.cc
+++ b/absl/base/internal/sysinfo.cc
@@ -101,8 +101,7 @@ static double GetNominalCPUFrequency() {
     }
   }
   return 1.0;
-#endif /* WINAPI_FAMILY_PARTITION(WINAPI_PARTITION_APP) && \
-          !WINAPI_FAMILY_PARTITION(WINAPI_PARTITION_DESKTOP) */
+#endif // WINAPI_PARTITION_APP && !WINAPI_PARTITION_DESKTOP
 }
 
 #elif defined(CTL_HW) && defined(HW_CPU_FREQ)


absl/base/internal/sysinfo.cc
```diff
@@ -76,6 +76,7 @@ static int GetNumCPUs() {
 #if defined(_WIN32)
 
 static double GetNominalCPUFrequency() {
+#if defined(WINAPI_FAMILY) && (WINAPI_FAMILY != WINAPI_FAMILY_APP)
```
vslashg: I'm curious about the thinking here.  In the case where WINAPI_FAMILY is not defined, is it correct to assume registry access is not available?  Might this cause problems on older toolchains where this macro is not yet defined?

Is it possible that this wants to be
```
#if !defined(WINAPI_FAMILY) || (WINAPI_FAMILY != WINAPI_FAMILY_APP)
```
?  (This is not intended to be a leading question; I'm really asking.)

LoadLibrary: Hi vslashg, the following was extracted from the winapifamily.h public header in the Windows SDK.

`WINAPI_FAMILY` might be undefined, or defined as:
* `WINAPI_FAMILY_PC_APP` - Windows Store Applications
* `WINAPI_FAMILY_PHONE_APP` - Windows Phone Applications
* `WINAPI_FAMILY_SYSTEM` - Windows Drivers and Tools
* `WINAPI_FAMILY_SERVER` - Windows Server Applications
* `WINAPI_FAMILY_DESKTOP_APP` - Windows Desktop Applications

If no WINAPI_FAMILY value is specified, then all APIs available to Windows desktop applications are exposed.
```c
#ifndef WINAPI_FAMILY
#define WINAPI_FAMILY WINAPI_FAMILY_DESKTOP_APP
#endif
```
So, not having WINAPI_FAMILY defined should work fine for code written for desktop apps.

Did you have any problem with this code in an old toolchain?

LoadLibrary: BTW, `#include <winapifamily.h>` is literally the first line of Windows.h.

LoadLibrary: Hi @vslashg , I went to abseil compatibility page (https://abseil.io/docs/cpp/platforms/platforms) and noticed that abseil supports down to "MVC 2015 Update 3, MSVC 2017", but no specific Windows version or SDK version is mentioned.
I think you might be on something here. I'm not sure if the SDK for Windows 7 has winapifamily.h.
Can you confirm what is the minimal Windows/SDK version abseil supports?

zhangxy988: In general, we support a version until 5 years after its replacement is available. In this case, I think we can probably ignore windows 7.
@vslashg WDYT?

LoadLibrary: I've read that Google is going to support Win7 until at least July 15, 2021 (https://cloud.google.com/blog/products/chrome-enterprise/how-chrome-is-helping-enterprises-still-using-windows-7). That's why I've created a Win7 VM + VS 2008 to check my change.
What is not clear to me is which toolchain abseil uses on Win7. Abseil's website (https://abseil.io/docs/cpp/platforms/platforms) informs that MVC 2015 Update 3 is used, but it doesn't say which SDK version is used.

absl/base/internal/sysinfo.cc
```diff
@@ -76,6 +76,12 @@ static int GetNumCPUs() {
 #if defined(_WIN32)
 
 static double GetNominalCPUFrequency() {
+// UWP apps don't have access to the registry and currently don't provide an
+// API informing about CPU nominal frequency.
+#if WINAPI_FAMILY_PARTITION(WINAPI_PARTITION_APP) && \
+    !WINAPI_FAMILY_PARTITION(WINAPI_PARTITION_DESKTOP)
```
zhangxy988, line 82: Do you know since which version of windows SDK is `WINAPI_FAMILY_PARTITION` defined? My only concern is that if it is not defined this will fail to compile.

zhangxy988:
I will run this through our internal testing system.

absl/base/internal/sysinfo.cc
```diff
@@ -95,6 +101,8 @@ static double GetNominalCPUFrequency() {
     }
   }
   return 1.0;
+#endif /* WINAPI_FAMILY_PARTITION(WINAPI_PARTITION_APP) && \
+          !WINAPI_FAMILY_PARTITION(WINAPI_PARTITION_DESKTOP) */
```
zhangxy988: LINT warning: 
Complex multi-line /*...*/-style comment found. Lint may give bogus warnings.  Consider replacing these with //-style comments, with #if 0...#endif, or with more clearly structured multi-line comments.

absl/base/internal/sysinfo.cc
LoadLibrary: WINAPI_FAMILY_PARTITION has been defined in the Windows 8 SDK. This change has been successfully tested on Windows 7 with Win7 SDK using VS2008. This change works on Win7 because WINAPI_FAMILY_PARTITION is not defined in the Win7 SDK.

absl/base/internal/sysinfo.cc
LoadLibrary: I'm not familiar with the linting rules used by abseil-cpp.
What is the recommended way to have that comment?  Would the following acceptable?
```c
#endif // WINAPI_FAMILY_PARTITION(WINAPI_PARTITION_APP) && \
       // !WINAPI_FAMILY_PARTITION(WINAPI_PARTITION_DESKTOP)

Comment (LoadLibrary) [reply absl/base/internal/sysinfo.cc]: Fixed: #endif // WINAPI_PARTITION_APP && !WINAPI_PARTITION_DESKTOP

**Comment by JackBoosY** [comment]
Hi guys, I have another issue:
In _SOURCE/absl/debugging/symbolize_win32.inc_ function `InitializeSymbolizer` and `Symbolize`, there are functions `SymSetOptions` and` SymFromAddr` which are not supported under uwp[1][2].
So I simply disabled them: return directly or return false[3].
Is this acceptable?

[1]. https://docs.microsoft.com/en-us/windows/win32/api/dbghelp/nf-dbghelp-symfromaddr
[2]. `DbgHelp.h` Line 30:
```
#if WINAPI_FAMILY_PARTITION(WINAPI_PARTITION_DESKTOP | WINAPI_PARTITION_PKG_WER)
```
[3]. [#10301](https://github.com/microsoft/vcpkg/pull/10301)

LoadLibrary:
@JackBoosY The purpose of this PR is to fix the abseil build used by WebRTC for UWP. WINAPI_PARTITION_PKG_WER is not the right way to gate code based on UWP/Desktop partitions. Feel free to file another bug and post a different PR to fix SOURCE/absl/debugging/symbolize_win32.inc.

JackBoosY:
@LoadLibrary Okay, but I'm not sure my changes is correct.

PR #605, opened 2020-01-17
Removing duplicated base filenames for Apple's libtool.

### LINKED ISSUE #604: BuildBreak: libtool on MacOS doesn't allow duplicated filenames
When building the absl_strings static library, libtool emits a warning, which is in fact something that breaks later eventually: `/Library/Developer/CommandLineTools/usr/bin/libtool: warning same member name (escaping.o) in output file used for input files`

The gist is that you can't have two files in a MacOS static library with the same base filename, as one will override the other. Basically, Apple's libtool crops the directory name away, only to keep the base filename.

This breaks eventually deep into gRPC's build system, thanks to sanity checks provided by NodeJS' build system: grpc/grpc#21702

Suggest adding a suffix to this duplicated filename to make Apple's libtool happy: https://github.com/abseil/abseil-cpp/blob/b3aaac8a37c467a1125c794196caa90d0957bdc3/absl/strings/CMakeLists.txt#L41

nicolasnoble:
Fixes #604


zhangxy988:
Please see my comment (https://github.com/abseil/abseil-cpp/issues/604#issuecomment-576753127).

nicolasnoble:
Supersed by internal changes that'll be pushed here eventually.

PR #608, opened 2020-01-20
fix: Add support for more ARM processors detection

### LINKED ISSUE #603: AbseilConfigureCopts.cmake does not handle architecture aarch64
On an nvidia Xavier the cmake variable `CMAKE_SYSTEM_PROCCESSOR` returns `aarch64`. This makes ARM platform detection fail here
https://github.com/abseil/abseil-cpp/blob/29235139149790f5afc430c11cec8f1eb1677607/absl/copts/AbseilConfigureCopts.cmake#L14-L22

which in turn causes compilation failures with absl::random. I can get around the problem by doing
```diff
diff --git a/absl/copts/AbseilConfigureCopts.cmake b/absl/copts/AbseilConfigureCopts.cmake
index b442646..4e55754 100644
--- a/absl/copts/AbseilConfigureCopts.cmake
+++ b/absl/copts/AbseilConfigureCopts.cmake
@@ -11,7 +11,7 @@ if("${CMAKE_SYSTEM_PROCESSOR}" STREQUAL "x86_64")
   else()
     set(ABSL_RANDOM_RANDEN_COPTS "${ABSL_RANDOM_HWAES_X64_FLAGS}")
   endif()
-elseif("${CMAKE_SYSTEM_PROCESSOR}" STREQUAL "arm")
+elseif("${CMAKE_SYSTEM_PROCESSOR}" STREQUAL "arm" OR "${CMAKE_SYSTEM_PROCESSOR}" STREQUAL "aarch64" )
   if ("${CMAKE_SIZEOF_VOID_P}" STREQUAL "8")
     set(ABSL_RANDOM_RANDEN_COPTS "${ABSL_RANDOM_HWAES_ARM64_FLAGS}")
   elseif("${CMAKE_SIZEOF_VOID_P}" STREQUAL "4")
```

but I am not entirely sure if it's the best way of going about this. Anyone else get this issue?

andre-nguyen:
`aarch64` is the processor string returned by an Nvidia TX2 and an Nvidia Xavier.
`arm.*` should be able to match processors such as `armv7l` (Raspberry pi 3, Odroid) 

Fixes #603


absl/copts/AbseilConfigureCopts.cmake
```diff
@@ -11,7 +11,7 @@ if("${CMAKE_SYSTEM_PROCESSOR}" STREQUAL "x86_64" OR "${CMAKE_SYSTEM_PROCESSOR}"
   else()
     set(ABSL_RANDOM_RANDEN_COPTS "${ABSL_RANDOM_HWAES_X64_FLAGS}")
   endif()
-elseif("${CMAKE_SYSTEM_PROCESSOR}" STREQUAL "arm")
+elseif("${CMAKE_SYSTEM_PROCESSOR}" MATCHES "arm.*|aarch64")
```
zhangxy988, line 14: Is there any specific reason to use regex matches rather than just `STREQUAL arm OR STREQUAL(aarch64)`?

zhangxy988:
Also, since we don't have test coverage for ARM yet, could you please test and verify that your change works?

absl/copts/AbseilConfigureCopts.cmake
andre-nguyen: Good question, yes there is. 

On unix systems `CMAKE_SYSTEM_PROCESSOR` is set to the output of `uname -p` which, after some googling through some common ARM based boards (rpi3, odroid) I see the arm instruction set version appended at the end, e.g. `armv7l`, `armv6l`. I'm also guessing the original author had a system where `CMAKE_SYSTEM_PROCESSOR` was set to `arm` so the regex `arm.*` should match all these possibilities whereas `STREQAL arm` wouldn't due to it not being an exact match.

andre-nguyen:
Here is the build/execution log on an nvidia Xavier
[log.txt](https://github.com/abseil/abseil-cpp/files/4093475/log.txt)

Looks like absl_sysinfo_test failed. Running the test manually I get
```
nvidia2@nvidia2-desktop:~/Documents/abseil-cpp/build/absl/base$ ./absl_sysinfo_test
Running main() from /home/nvidia2/Documents/abseil-cpp/build/googletest-src/googletest/src/gtest_main.cc
[==========] Running 4 tests from 1 test suite.
[----------] Global test environment set-up.
[----------] 4 tests from SysinfoTest
[ RUN      ] SysinfoTest.NumCPUs
[       OK ] SysinfoTest.NumCPUs (0 ms)
[ RUN      ] SysinfoTest.NominalCPUFrequency
/home/nvidia2/Documents/abseil-cpp/absl/base/internal/sysinfo_test.cc:47: Failure
Expected equality of these values:
  NominalCPUFrequency()
    Which is: 2.2656e+09
  1.0
    Which is: 1
CPU frequency detection was fixed! Please update unittest.
[  FAILED  ] SysinfoTest.NominalCPUFrequency (0 ms)
[ RUN      ] SysinfoTest.GetTID
[       OK ] SysinfoTest.GetTID (409 ms)
[ RUN      ] SysinfoTest.LinuxGetTID
[       OK ] SysinfoTest.LinuxGetTID (0 ms)
[----------] 4 tests from SysinfoTest (410 ms total)

[----------] Global test environment tear-down
[==========] 4 tests from 1 test suite ran. (410 ms total)
[  PASSED  ] 3 tests.
[  FAILED  ] 1 test, listed below:
[  FAILED  ] SysinfoTest.NominalCPUFrequency

 1 FAILED TEST
```

Which seems it might be a normal problem according to the comment??

https://github.com/abseil/abseil-cpp/blob/a2e6adecc294dc4cd98cc285a9134ce58e0f2ad0/absl/base/internal/sysinfo_test.cc#L45-L48

Misamgamshadzehi (commented):
بروزرسانی میخواهم بکنم چکارکنم

PR #616, opened 2020-02-10
Fix build on FreeBSD/powerpc

kgotlinux:
Only glibc has __ppc_get_timebase(), __ppc_get_timebase_freq() and sys/platform/ppc.h. For TBR and TBR frequency, FreeBSD doesn't have equivalent functions.

### DIFF
--- a/absl/base/internal/unscaledcycleclock.cc
+++ b/absl/base/internal/unscaledcycleclock.cc
@@ -12,7 +12,6 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#include "absl/base/call_once.h"
 #include "absl/base/internal/unscaledcycleclock.h"
 
 #if ABSL_USE_UNSCALED_CYCLECLOCK
@@ -25,8 +24,8 @@
 #ifdef __GLIBC__
 #include <sys/platform/ppc.h>
 #elif defined(__FreeBSD__)
-#include <sys/types.h>
 #include <sys/sysctl.h>
+#include <sys/types.h>
 #endif
 #endif
 
@@ -65,24 +64,22 @@ double UnscaledCycleClock::Frequency() {
 int64_t UnscaledCycleClock::Now() {
 #ifdef __GLIBC__
   return __ppc_get_timebase();
-#elif defined(__FreeBSD__)
+#else
 #ifdef __powerpc64__
   int64_t tbr;
-  asm volatile("mfspr %0, 268" : "=r" (tbr));
+  asm volatile("mfspr %0, 268" : "=r"(tbr));
   return tbr;
 #else
- int32_t tbu, tbl, tmp;
-    asm volatile(
-    "0:\n"
-    "mftbu %[hi32]\n"
-    "mftb %[lo32]\n"
-    "mftbu %[tmp]\n"
-    "cmpw %[tmp],%[hi32]\n"
-    "bne 0b\n"
-    : [hi32] "=r"(tbu), [lo32] "=r"(tbl),
-    [tmp] "=r"(tmp)
-  );
-  return ((int64_t) tbu << 32) | tbl);
+  int32_t tbu, tbl, tmp;
+  asm volatile(
+      "0:\n"
+      "mftbu %[hi32]\n"
+      "mftb %[lo32]\n"
+      "mftbu %[tmp]\n"
+      "cmpw %[tmp],%[hi32]\n"
+      "bne 0b\n"
+      : [ hi32 ] "=r"(tbu), [ lo32 ] "=r"(tbl), [ tmp ] "=r"(tmp));
+  return (static_cast<int64_t>(tbu) << 32) | tbl;
 #endif
 #endif
 }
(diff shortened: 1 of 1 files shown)


absl/base/internal/unscaledcycleclock.cc
```diff
@@ -57,11 +62,34 @@ double UnscaledCycleClock::Frequency() {
 #elif defined(__powerpc__) || defined(__ppc__)
 
 int64_t UnscaledCycleClock::Now() {
+#ifdef __GLIBC__
   return __ppc_get_timebase();
+#elif defined(__FreeBSD__)
+  union { long long complete; unsigned int part[2]; } ticks;
+  unsigned int tmp;
+  asm volatile(
```
derekmauro, line 74: I'm not sure this code is optimal. It looks like, for example, ppc64 has a way of getting the value in one instruction.

https://github.com/bminor/glibc/blob/5f72f9800b250410cad3abfeeb09469ef12b2438/sysdeps/powerpc/sys/platform/ppc.h#L26-L51

absl/base/internal/unscaledcycleclock.cc
```diff
@@ -57,11 +62,34 @@ double UnscaledCycleClock::Frequency() {
 #elif defined(__powerpc__) || defined(__ppc__)
 
 int64_t UnscaledCycleClock::Now() {
+#ifdef __GLIBC__
   return __ppc_get_timebase();
+#elif defined(__FreeBSD__)
+  union { long long complete; unsigned int part[2]; } ticks;
```
derekmauro: Though you sometimes see it, type-punning through a union is undefined behavior in C++. See https://adriann.github.io/undefined_behavior.html for example.

absl/base/internal/unscaledcycleclock.cc
```diff
@@ -57,11 +62,34 @@ double UnscaledCycleClock::Frequency() {
...
+#endif
 }
 
 double UnscaledCycleClock::Frequency() {
+#ifdef __GLIBC__
   return __ppc_get_timebase_freq();
+ #elif defined(__FreeBSD__)
+  double timebaseFrequency = 0;
+  size_t length = sizeof(timebaseFrequency);
+  sysctlbyname("kern.timecounter.tc.timebase.frequency", &timebaseFrequency, &length, NULL, 0);
```
derekmauro: Can you cache this value using some sort of "once" initialization? I don't think Abseil calls the code at all, but other (internal) code calls it very often.

For various reasons you'll need to use `absl::base_internal::LowLevelCallOnce` in this context.

https://github.com/abseil/abseil-cpp/blob/bf78e977309c4cb946914b456404141ddac1c302/absl/base/call_once.h#L199-L208

absl/base/internal/unscaledcycleclock.cc
```diff
@@ -57,11 +62,34 @@ double UnscaledCycleClock::Frequency() {
...
+  );
+  return ticks.complete;
+#endif
 }
 
 double UnscaledCycleClock::Frequency() {
+#ifdef __GLIBC__
   return __ppc_get_timebase_freq();
+ #elif defined(__FreeBSD__)
+  double timebaseFrequency = 0;
```
derekmauro: https://google.github.io/styleguide/cppguide.html#Variable_Names

Variable names are lower-case with underscore in Google C++ style.

absl/base/internal/unscaledcycleclock.cc
derekmauro: https://google.github.io/styleguide/cppguide.html#Integer_Types

This is a case where you should use fixed-width integer types like `int64_t`.

derekmauro (changes requested):

kgotlinux:
Please check if it's ok now.

absl/base/internal/unscaledcycleclock.cc
```diff
@@ -57,11 +63,40 @@ double UnscaledCycleClock::Frequency() {
 #elif defined(__powerpc__) || defined(__ppc__)
 
 int64_t UnscaledCycleClock::Now() {
+#ifdef __GLIBC__
   return __ppc_get_timebase();
+#elif defined(__FreeBSD__)
```
derekmauro: Is this code portable? I think it will work on non-FreeBSD ppc platforms as well. Should this just be `#else`?

absl/base/internal/unscaledcycleclock.cc
```diff
@@ -57,11 +63,40 @@ double UnscaledCycleClock::Frequency() {
...
+#endif
 }
 
 double UnscaledCycleClock::Frequency() {
+#ifdef __GLIBC__
   return __ppc_get_timebase_freq();
+ #elif defined(__FreeBSD__)
+  static once_flag init_timebase_frequency_once;
+  static double timebase_frequency = 0;
+  size_t length = sizeof(timebase_frequency);
```
derekmauro: This can go inside the lambda.

absl/base/internal/unscaledcycleclock.cc
```diff
@@ -57,11 +63,40 @@ double UnscaledCycleClock::Frequency() {
...
+  base_internal::LowLevelCallOnce( &init_timebase_frequency_once, [&]() { sysctlbyname("kern.timecounter.tc.timebase.frequency", &timebase_frequency, &length, NULL, 0); });
+  return timebase_frequency;
+#endif
```
derekmauro, line 101: We probably want an `#else` `#error` here.

derekmauro (changes requested):

derekmauro:
I pushed changes to your branch to resolve the comments.

PR #622, opened 2020-02-20
Add forgotten ifdef conditionals

franksinankaya:


CMakeLists.txt
```diff
@@ -86,6 +86,9 @@ option(ABSL_USE_GOOGLETEST_HEAD
 
 option(ABSL_RUN_TESTS "If ON, Abseil tests will be run." OFF)
 
+# link fails on arm64 and x86-64 without this
+set(CMAKE_POSITION_INDEPENDENT_CODE ON)
```
derekmauro: I talked with @asoffer about this and we don't think adding this unconditionally is the right thing to do since it can have performance implications. We think that a recent change (37dd2562ec830d547a1524bb306be313ac3f2556) actually broke adding this when `-DBUILD_SHARED_LIBS=ON` is specified.

I suggest we revert this change try to figure it out separately from the VDSO support change.

franksinankaya: ok, I'll drop this

PR #628, opened 2020-02-26
Support testing with a local googletest checkout

bbarenblat:
Debian likes doing dependency management manually, so add support for testing against a local googletest checkout rather than one downloaded from GitHub. Add an `ABSL_LOCAL_GOOGLETEST_DIR` variable that can be used in conjunction with `-DABSL_RUN_TESTS=ON -DABSL_USE_GOOGLETEST_HEAD=OFF` to specify the googletest location manually, and do a bit of related cleanup.


derekmauro:
As per https://github.com/abseil/abseil-cpp/blob/master/CONTRIBUTING.md#guidelines-for-pull-requests, Googlers should send out Piper CLs, they are much easier for us to handle. Sorry about this, we should try to make this more prominent.

PR #638, opened 2020-03-13
Fix abs/strings/numbers compilation on Win/MSVS.

mkolomeytsev:
There is no x86intrin.h on Win/MSVS.
Use intrin.h instead.

### DIFF
--- a/absl/strings/numbers.h
+++ b/absl/strings/numbers.h
@@ -25,7 +25,11 @@
 #define ABSL_STRINGS_NUMBERS_H_
 
 #ifdef __SSE4_2__
+#ifdef _MSC_VER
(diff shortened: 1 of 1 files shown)


mkolomeytsev:
@googlebot I signed it!

absl/strings/numbers.h
```diff
@@ -25,8 +25,12 @@
 #define ABSL_STRINGS_NUMBERS_H_
 
 #ifdef __SSE4_2__
```
derekmauro, line 27: It doesn't look like `__SSE4_2__` is defined by MSVC, so I don't think this is the correct fix. Are you trying to define `__SSE4_2__` yourself under MSVC?

PR #639, opened 2020-03-13
Fix abs/strings/numbers compilation on Win/MSVS.

mkolomeytsev:
There is no x86intrin.h on Win/MSVS.
Use intrin.h instead.

### DIFF
--- a/absl/strings/numbers.h
+++ b/absl/strings/numbers.h
@@ -25,7 +25,11 @@
 #define ABSL_STRINGS_NUMBERS_H_
 
 #ifdef __SSE4_2__
+#ifdef _MSC_VER
+#include <intrin.h>
+#else
 #include <x86intrin.h>
+#endif
 #endif
 
 #include <cstddef>


absl/strings/numbers.h
```diff
@@ -25,8 +25,12 @@
 #define ABSL_STRINGS_NUMBERS_H_
 
 #ifdef __SSE4_2__
```
derekmauro, line 27: It doesn't look like `__SSE4_2__` is defined by MSVC, so I don't think this is the correct fix. Are you trying to define `__SSE4_2__` yourself under MSVC?

rongjiecomputer: I think what Eigen3 and Tensorflow do is they assume SSE4.2 is enabled if AVX is enabled since MSVC does define `__AVX__` when `/arch:AVX` flag is used. Not the best fix, but it is safer than no guard at all.

vslashg: I'm investigating whether we can get away with using immintrin.h instead, which is portable to MSVC and I hope exposes all the intrinsics we use here.

vslashg:
I haven't completed my testing, but initial testing seems to indicate that replacing "x86intrin.h" with "immintrin.h" will work on our end.  Could you check if this simpler change would fix the build in your configuration?

IAMGoogleGo (approved):

rebel2270 (approved):

IAMGoogleGo (approved):

absl/strings/numbers.h
MBkkt: > It doesn't look like __SSE4_2__ is defined by MSVC

Don't know why, but its defined

Dronplane: > It doesn't look like `__SSE4_2__` is defined by MSVC,

But there are no "standart" saying __SSE4_2__ will be never defined by MSVC.

PR #647, opened 2020-03-26
Simplify use of abseil in add_subdirectory mode

adrianimboden:
We use abseil with CMake FetchContent and add_subdirectory. We also use googletest in the same way.
This produces a conflict because googletest is now included twice. This option would solves this problem.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -81,6 +81,9 @@ endif()
 ## pthread
 find_package(Threads REQUIRED)
 
+option(ABSL_USE_EXTERNAL_GOOGLETEST
+  "If ON, abseil will assume that the targets for googletest are already provided by the including project folder. This makes sense when abseil is used with add_subproject." OFF)
+
 option(ABSL_USE_GOOGLETEST_HEAD
   "If ON, abseil will download HEAD from googletest at config time." OFF)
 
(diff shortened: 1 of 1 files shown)


adrianimboden:
@googlebot I signed it!

CMakeLists.txt
```diff
@@ -81,6 +81,9 @@ endif()
 ## pthread
 find_package(Threads REQUIRED)
 
+option(ABSL_IGNORE_GOOGLETEST
+  "If ON, abseil will use the googletest targets which are provided somehow (useful when abseil is a used with add_subdirectory)." OFF)
```
derekmauro: > use the googletest targets which are provided somehow

Can you use more precise, formal language here? "provided somehow" doesn't really tell me much.

Also, there was a commit yesterday that may have fixed `add_subdirectory` support: https://github.com/abseil/abseil-cpp/commit/79e0dc11514df035a8d07a356f9ee1800fb2160c

I'm not a CMake expert, but can you explain why `ABSL_LOCAL_GOOGLETEST_DIR` is not sufficient for your use case?

derekmauro (changes requested):

CMakeLists.txt
adrianimboden: I use it like this:
```
my_app/CMakeLists.txt
my_app/deps/abseil-cpp/CMakeLists.txt
my_app/deps/googletest/CMakeLists.txt
```
my_app/CMakeLists.txt contains:
- add_subdirectory(deps/abseil-cpp)
- add_subdirectory(deps/googletest)

The cmake targets for googletest are already available in the project when abseil calls add_subdirectory. ABSL_LOCAL_GOOGLETEST_DIR would allow me to specify where my googletest folder is (my_app/deps/googletest in this case), but it would include the googletest cmake project a second time, which must not happen as it causes duplicate targets. That is the reason why ABSL_LOCAL_GOOGLETEST_DIR is not sufficient.

I did not find any changes in 79e0dc1 which would solve this problem.

samdespera2:
Hello and Have a great time

CMakeLists.txt
derekmauro: I understand, thanks.

In that case, I think something like `ABSL_USE_EXTERNAL_GOOGLETEST` is a better name. Please update the docstring as well and possibly add a comment that explains it further.

adrianimboden: Thanks, I updated the pull request.

Does this look good?

derekmauro (approved):

Issue #659, opened 2020-04-08
status_payload_printer.cc is not compiled with Clang 10 on Windows


samolisov:
The following code in the `absl/status/status_payload_printer.cc` raises a compilation error on Windows when Clang 10 (via clang-cl.exe) is used to compile Abseil:

```c++
namespace {
// Tried constant initialized global variable but it doesn't work with Lexan
// (MSVC's `std::atomic` has trouble constant initializing).
std::atomic<StatusPayloadPrinter>& GetStatusPayloadPrinterStorage() {
  ABSL_CONST_INIT static std::atomic<StatusPayloadPrinter> instance{nullptr};
  return instance;
}
}  // namespace
```

The error:

```
..\absl\status\status_payload_printer.cc(28,60): error: variable does not have a constant initializer
  ABSL_CONST_INIT static std::atomic<StatusPayloadPrinter> instance{nullptr};
                                                           ^~~~~~~~~~~~~~~~~
..\absl\status\status_payload_printer.cc(28,3): note: required by 'require_constant_initialization' attribute here
  ABSL_CONST_INIT static std::atomic<StatusPayloadPrinter> instance{nullptr};
  ^~~~~~~~~~~~~~~
..\absl/base/attributes.h(618,27): note: expanded from macro 'ABSL_CONST_INIT'
#define ABSL_CONST_INIT [[clang::require_constant_initialization]]
                          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
C:\Program Files (x86)\Microsoft Visual Studio\2017\BuildTools\VC\Tools\MSVC\14.15.26726\include\xxatomic(207,21): note: cast that performs the conversions of a reinterpret_cast is not allowed in a constant expression
                : _Atomic_address{(_ATOMIC_UINT)_Right}
                                  ^
..\absl\status\status_payload_printer.cc(28,60): note: in call to 'atomic(nullptr)'
  ABSL_CONST_INIT static std::atomic<StatusPayloadPrinter> instance{nullptr};
                                                           ^
1 error generated.
```

I see the following solution: to add a check the compiler is not MSVC (since clang-cl mimics MSVC) into the `ABSL_CONST_INIT` macro but it won't protect from the error if `clang++` is used instead of `clang-cl`. Maybe someone knows a better solution?

zhangxy988:
[...]

PR #661, opened 2020-04-10
Use base_internal::AtomicHook instead of std::atomic

samolisov:
std::atomic has a broken implementation on the Windows platform and it
is not conform to the ABSL_CONST_INIT macro when clang-cl is used as a
compiler: the macro is expanded to the [[clang::require_constant_initialization]]
attribute and the attribute cannot be applied to the broken std::atomic.

Therefore, std::atomic has been replaced with absl::base_internal::AtomicHook
to fix the compilation error (thank Derek Mauro for the suggestion).

Issue: #659 

Signed-off-by: Pavel Samolysov <samolisov@gmail.com>

### DIFF
--- a/absl/status/status_payload_printer.cc
+++ b/absl/status/status_payload_printer.cc
@@ -16,26 +16,21 @@
 #include <atomic>
 
 #include "absl/base/attributes.h"
+#include "absl/base/internal/atomic_hook.h"
 
 namespace absl {
 ABSL_NAMESPACE_BEGIN
 namespace status_internal {
 
-namespace {
-// Tried constant initialized global variable but it doesn't work with Lexan
-// (MSVC's `std::atomic` has trouble constant initializing).
-std::atomic<StatusPayloadPrinter>& GetStatusPayloadPrinterStorage() {
-  ABSL_CONST_INIT static std::atomic<StatusPayloadPrinter> instance{nullptr};
-  return instance;
-}
-}  // namespace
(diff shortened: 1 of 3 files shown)


absl/status/status_payload_printer.cc
```diff
@@ -16,26 +16,27 @@
...
 namespace status_internal {
 
 namespace {
-// Tried constant initialized global variable but it doesn't work with Lexan
-// (MSVC's `std::atomic` has trouble constant initializing).
-std::atomic<StatusPayloadPrinter>& GetStatusPayloadPrinterStorage() {
-  ABSL_CONST_INIT static std::atomic<StatusPayloadPrinter> instance{nullptr};
+absl::base_internal::AtomicHook<StatusPayloadPrinter>&
+GetStatusPayloadPrinterStorage() {
+  ABSL_CONST_INIT static absl::base_internal::AtomicHook<StatusPayloadPrinter>
```
derekmauro: This actually needs to be a global, so you can remove the function that wraps it. It should also be `ABSL_INTERNAL_ATOMIC_HOOK_ATTRIBUTES`, not `ABSL_CONST_INIT`. Don't initialize it to `nullptr` either.

samolisov: @derekmauro Thank you for the comment. I've changed the commit to introduce the global variable and use `ABSL_INTERNAL_ATOMIC_HOOK_ATTRIBUTES` instead of `ABSL_CONST_INIT`.

derekmauro (approved):

PR #664, opened 2020-04-14
Fix ABSL_RANDOM_RANDEN_COPTS setting on FreeBSD

sunpoet:
The error message is as follows:
```
CMake Warning at absl/copts/AbseilConfigureCopts.cmake:30 (message):
  Value of CMAKE_SYSTEM_PROCESSOR (amd64) is unknown and cannot be used to
  set ABSL_RANDOM_RANDEN_COPTS
  CMake/AbseilHelpers.cmake:18 (include)
  CMakeLists.txt:57 (include)
```


mbxx (approved):
Thanks!

sunpoet:
Thanks!

Issue #668, opened 2020-04-21
Incorrect use of FindThreads in cmake package config file


K-ballo:
The generated cmake package config file does a direct inclusion of the `FindThread` module:

https://github.com/abseil/abseil-cpp/blob/1a02b7a2054c24f900dab796edb812f9260b51a6/CMake/abslConfig.cmake.in#L3

This results in the following cmake developer warning:

```
 CMake Warning (dev) at /usr/local/lib/python3.6/dist-packages/cmake/data/share/cmake-3.17/Modules/FindPackageHandleStandardArgs.cmake:272 (message):
  The package name passed to `find_package_handle_standard_args` (Threads)
  does not match the name of the calling package (absl).  This can lead to
  problems in calling code that expects `find_package` result variables
  (e.g., `_FOUND`) to follow a certain pattern.
````

The proper way to consume find modules is via `find_package(<NAME>)`, which interacts with the module by setting and reading various variables prefixed with `<NAME>` (hence the above warning).

Furthermore, the [`find_dependency`](https://cmake.org/cmake/help/latest/module/CMakeFindDependencyMacro.html) macro was created to wrap `find_package` calls in package configuration files to correctly handle `QUIET` and `REQUIRED` arguments given to to the original call.

Summing up, rather than directly including the `FindThreads` module, the cmake package file should instead do:
```
include(CMakeFindDependencyMacro)
find_dependency(Threads)
```

PR #671, opened 2020-04-22
CMake Thread dependency change

zach2good:
Issue: https://github.com/abseil/abseil-cpp/issues/668
Implement the proposed suggestion from the issue above issue


derekmauro:
Thanks!

PR #675, opened 2020-04-29
Fix build on riscv32

kraj:
Identify PC register from signal context

Signed-off-by: Khem Raj <raj.khem@gmail.com>

### DIFF
--- a/absl/base/internal/direct_mmap.h
+++ b/absl/base/internal/direct_mmap.h
@@ -26,10 +26,6 @@
 
 #ifdef __linux__
 
-#if !defined(__NR_mmap) && defined(__riscv) && __riscv_xlen == 32
-# define __NR_mmap __NR_mmap2
-#endif
-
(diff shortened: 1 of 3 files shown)


absl/base/internal/direct_mmap.h
```diff
@@ -26,6 +26,10 @@
 
 #ifdef __linux__
 
+#if !defined(__NR_mmap) && defined(__riscv) && __riscv_xlen == 32
+# define __NR_mmap __NR_mmap2
```
vslashg: Why is this needed in your change?  Is this only intended for use with  the __BIONIC__ branch later below?  If so, it might be better to define SYS_mmap appropriately in the __BIONIC__ setup stanza.

kraj: rv32 does not define mmap syscall only mmap2, so its really not libc specific

PR #678, opened 2020-05-04
fix MSVC warning 4245: conversion signed => unsigned during initialization

christoph-cullmann:
MSVC bails out for me with -Werror (like we enforce for our own code at AbsInt), if I use the flat hash tables because of this sign/unsigned issue in the code (warning 4245).

With this trivial fix, all is fine and it should have zero impact on the generated (binary) code.


christoph-cullmann:
Thanks for taking care!

PR #684, opened 2020-05-13
Fix public target name of the random library

nicholasbishop:
The name seems to be `random_random` rather than just `random`:

https://github.com/abseil/abseil-cpp/blob/master/absl/random/CMakeLists.txt#L19

### DIFF
--- a/CMake/README.md
+++ b/CMake/README.md
@@ -93,7 +93,7 @@ absl::flags
 absl::memory
 absl::meta
 absl::numeric
-absl::random
+absl::random_random
 absl::strings
 absl::synchronization
 absl::time


nicholasbishop:
I can't sign the CLA because https://cla.developers.google.com/clas loads with my work account, and there's no account picker. I don't have a good understanding of why every different Google site implements the account picker differently (or not at all), but any chance this can be fixed? Thanks :)

suertreus:
I've filed an internal bug requesting an account picker UI for the site.  As a workaround, you should be able to switch accounts by adding `?authuser=n` to the URL, e.g. https://cla.developers.google.com/clas?authuser=1 for your second logged-in account.  If that doesn't work, try starting an incognito or private browsing session and logging in with *only* your personal account.

nicholasbishop:
Thanks a bunch for filing that internal bug, I think I ran into this same problem (in reverse) five years ago when signing the agreement for my work account. Would be great to see that fixed :)

Your suggestion of appending `?authuser=1` seemed to work. I also had to append that (as `&authuser=1`) after following the "Only Yourself" link. However, after submitting the agreement, if I attempt to go back to https://cla.developers.google.com/clas?authuser=1 to verify it worked, I get a 500 error. So that's weird, but hopefully it worked anyway.

nicholasbishop:
@googlebot I signed it!

nicholasbishop:
Good, looked like it did work. As an aside, might be a good idea to add some sort of help or feedback link to the CLA page, so that people have a more direct way of informing the relevant team of problems. Thanks again for your help.

suertreus:
Thanks for the fix!

PR #691, opened 2020-05-24
CMake: don't attempt to use googletest unless ABSL_RUN_TESTS==true.

### LINKED ISSUE #690: CMake: build requires googletest even if ABSL_RUN_TESTS is not enabled
I am using ABSL as a dependency in https://github.com/google/bloaty, but I'm not yet using ABSL's CMake files.

When I tried using ABSL's CMakeLists.txt, I get this error:

```
CMake Error at /usr/share/cmake-3.16/Modules/ExternalProject.cmake:1706 (file):
  file problem creating directory: /usr/src/googletest
  /usr/share/cmake-3.16/Modules/ExternalProject.cmake:3188 (_ep_set_directories)
  CMakeLists.txt:18 (ExternalProject_Add)


CMake Error at /usr/share/cmake-3.16/Modules/ExternalProject.cmake:1708 (message):
  dir '/usr/src/googletest' does not exist after file(MAKE_DIRECTORY)
  /usr/share/cmake-3.16/Modules/ExternalProject.cmake:3188 (_ep_set_directories)
  CMakeLists.txt:18 (ExternalProject_Add)
```

I didn't enable `ABSL_RUN_TESTS`. So it is surprising that ABSL is attempting to download googletest to `/usr/src`. There should be no reason to fail just because googletest is not present.

This should be as simple as changing `BUILD_TESTING` to `ABSL_RUN_TESTS` here:
https://github.com/abseil/abseil-cpp/blob/73ea9a95720c4f4faff9b1bcbad48d23f2cb2a46/CMakeLists.txt#L105

This is very similar to this previously-reported issue: https://github.com/abseil/abseil-cpp/issues/162

haberman:
Fixes #690.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -99,10 +99,8 @@ if(${ABSL_RUN_TESTS})
   # on the command line
   include(CTest)
   enable_testing()
-endif()
 
-## check targets
-if(BUILD_TESTING)
+  ## check targets
   if (NOT ABSL_USE_EXTERNAL_GOOGLETEST)
     set(absl_gtest_build_dir ${CMAKE_BINARY_DIR}/googletest-build)
     if(${ABSL_USE_GOOGLETEST_HEAD})


derekmauro:
As per https://github.com/abseil/abseil-cpp/blob/master/CONTRIBUTING.md#guidelines-for-pull-requests, can you please use the internal code review process? Thanks.

PR #692, opened 2020-05-26
Add C++20 alternative for result_of_t

stijnherfst:
Abseil doesn't compile in the newest MSVC version as std::result_of has been removed from the C++20 standard. Instead std::invoke_result should be used. By using a the standard [__cpp_lib_is_invocable](https://en.cppreference.com/w/cpp/feature_test) define we can easily test if std::invoke_result exists.


stijnherfst:
@googlebot I signed it!

derekmauro:
This doesn't build. The syntax required by `std::invoke_result` is slightly different than the syntax required by `std::result_of`.

https://en.cppreference.com/w/cpp/types/result_of

PR #697, opened 2020-05-30
Exclude empty directories

sunpoet:
install(DIRECTORY...) creates all directories. If there is no matching
files, it becomes an empty directory. The three empty directories are
absl/copts, absl/strings/testdata and absl/time/internal/cctz/testdata.
It is a workaround as cmake does not have an option to exclude empty
directories.

Reference:	https://gitlab.kitware.com/cmake/cmake/-/issues/19189


tsp186:
Maybe for the meantime or just empty to aware all of us

PR #707, opened 2020-06-08
Update ABSL_CONST_INIT doc to cover extern variables.

MichaelSheely:
This fell out of a discussion with Eric Fiselier from 2019.
More details about the reasoning can be found at: https://gist.github.com/EricWF/128781c188b1a4fca7581e7ea943d58b#applying-constinit-to-declarations-or-just-definitions

Apologies for the delay between commit and PR.  Looks like I got side tracked before sending the request.  Let me know if this looks OK or if we want to modify the documentation update or leave it as is.


absl/base/attributes.h
```diff
@@ -602,7 +602,9 @@
 // duration, because it guarantees that they will not suffer from the so-called
 // "static init order fiasco".  Prefer to put this attribute on the most visible
 // declaration of the variable, if there's more than one, because code that
-// accesses the variable can then use the attribute for optimization.
+// accesses the variable can then use the attribute for optimization. In the
+// case of extern variables, the annotation should also be placed on the
```
suertreus: How about "If that declaration is not visible at the definition (i.e. an `extern` variable), then the definition must also be marked `ABSL_CONST_INIT` so the compiler knows to enforce constant initialization." ?

I think what you're trying to add to the documentation is correct and important, but I want to avoid confusing Googlers, who need to use ABSL_CONST_INIT fairly frequently (due to style guide requirements) but will almost never encounter this particular situation (due to style guide constraints).

MichaelSheely: Thanks, that sounds much clearer.  Updated the PR.

suertreus (approved):

suertreus:
This is obsolete after c27ab06897f330267bed99061ed3e523e2606bf1, which gives new guidance in light of `constinit`.

PR #714, opened 2020-06-16
Added neccessary headers for FreeBSD

kgotlinux:
1. Changed order of sys/sysctl.h, because sysctl.h needs types.h.
2. Threads.h is neccessary for once_flag.
3. absl/base/call_once.h is neccessary for LowLevelCallOnce.

### DIFF
--- a/absl/base/internal/unscaledcycleclock.cc
+++ b/absl/base/internal/unscaledcycleclock.cc
@@ -24,8 +24,10 @@
 #ifdef __GLIBC__
 #include <sys/platform/ppc.h>
 #elif defined(__FreeBSD__)
-#include <sys/sysctl.h>
+#include "absl/base/call_once.h"
 #include <sys/types.h>
+#include <sys/sysctl.h>
+#include <threads.h>
 #endif
 #endif
 


absl/base/internal/unscaledcycleclock.cc
```diff
@@ -24,8 +24,10 @@
 #ifdef __GLIBC__
 #include <sys/platform/ppc.h>
 #elif defined(__FreeBSD__)
-#include <sys/sysctl.h>
+#include "absl/base/call_once.h"
 #include <sys/types.h>
+#include <sys/sysctl.h>
+#include <threads.h>
```
suertreus, line 30: If <threads.h> is needed for once_flag, shouldn't we be including it in call_once.h instead/also?

Mizux:
May I know which FreeBSD version did you use as well as the compiler ?

My 2 cents, currently I was able to build abseil-cpp on FreeBSD 12 and 11.4 using the default clang installed IF tests are disabled.
my VagrantFile without comments (cat Vagrantfile | sed "/^ *#/d" | sed "/^ *$/d"):
```ruby
Vagrant.configure("2") do |config|
  config.vm.guest = :freebsd
  config.vm.box = "generic/freebsd12"
  config.ssh.shell = "sh"
  config.vm.synced_folder ".", "/vagrant", id: "vagrant-root", disabled: true
  config.vm.provision "env", type: "shell", inline:<<-SHELL
  set -x
  pkg update -f
  pkg install -y git cmake vim
  SHELL
  config.vm.provision "devel", type: "shell", inline:<<-SHELL
  set -x
  git clone --depth 1 --branch 20200225.2 https://github.com/abseil/abseil-cpp.git
  cd abseil-cpp
  SHELL
  config.vm.provision "configure", type: "shell", inline:<<-SHELL
  set -x
  cd abseil-cpp
  cmake -S. -Bbuild
  SHELL
  config.vm.provision "build", type: "shell", inline:<<-SHELL
  set -x
  cd abseil-cpp
  cmake --build build -v
  SHELL
end
```
and running it using `vagrant up`.

@Googlers:
* Do you have a kokoro job to test FreeBSD ?
* if not, could you have a vagrant + virtualbox runner available on kokoro to add a new job... (maybe I should directly ask to the kokoro team ?).

PR #723, opened 2020-06-25
Fix CMake path for INSTALL_INCLUDEDIR

igor-egorov:
The raw literal `{ABSL_SUBDIR}` was used as a part of the path instead of variable `${ABSL_SUBDIR}`.


igor-egorov:
@googlebot I signed it!

Issue #135, opened 2018-06-28
Build failure on s390x


ajaygtm71:
I'm trying to use Abseil-cpp which is dependency of Istio-proxy on big endian by following https://abseil.io/docs/cpp/quickstart but its failing with the following error: 

```
ERROR: /home/ajay/.cache/bazel/_bazel_ajay/05774821e5b6d73b3dae6e2f79bcee31/external/com_google_absl/absl/base/BUILD.bazel:80:1: C++ compilation of rule '//absl/base:malloc_internal' failed (Exit 1)
In file included from absl/base/internal/low_level_alloc.cc:26:0:
./absl/base/internal/direct_mmap.h: In function 'void* absl::base_internal::DirectMmap(void*, size_t, int, int, int, off64_t)':
./absl/base/internal/direct_mmap.h:96:39: error: cast from 'void*' to 'uint32_t {aka unsigned int}' loses precision [-fpermissive]
       reinterpret_cast<uint32_t>(start), static_cast<uint32_t>(length),
                                       ^
./absl/base/internal/direct_mmap.h:99:10: error: 'reintrepret_cast' was not declared in this scope
   return reintrepret_cast<void*>(syscall(SYS_mmap, buf));
          ^
./absl/base/internal/direct_mmap.h:99:27: error: expected primary-expression before 'void'
   return reintrepret_cast<void*>(syscall(SYS_mmap, buf));
                           ^
./absl/base/internal/direct_mmap.h:99:27: error: expected ';' before 'void'
./absl/base/internal/direct_mmap.h:99:32: error: expected unqualified-id before '>' token
   return reintrepret_cast<void*>(syscall(SYS_mmap, buf));
                                ^
./absl/base/internal/direct_mmap.h:99:32: error: expected initializer before '>' token
./absl/base/internal/direct_mmap.h:95:12: warning: unused variable 'buf' [-Wunused-variable]
   uint32_t buf[6] = {
            ^
INFO: Elapsed time: 72.279s, Critical Path: 2.16s
INFO: 14 processes, linux-sandbox.
FAILED: Build did NOT complete successfully
//absl/algorithm:algorithm_benchmark                                  NO STATUS
//absl/algorithm:algorithm_test                                       NO STATUS
//absl/algorithm:container_test                                       NO STATUS
//absl/base:atomic_hook_test                                          NO STATUS
//absl/base:bit_cast_test                                             NO STATUS
//absl/base:call_once_test                                            NO STATUS
[...]

PR #728, opened 2020-06-30
Add support for s390x cpu frequency

Sidong-Wei:
As could be seen from [this](https://github.com/abseil/abseil-cpp/issues/135#issuecomment-405463796) issue, the unit test `SysinfoTest.NominalCPUFrequency` has always been failing on s390x machines. The reason for the failure is because there is no mechanism in abseil code to retrieve CPU time and frequency for s390x machines. I am adding the methods for s390x by referencing a similar method committed to another project, see [here](https://github.com/google/benchmark/pull/540/commits/94e51b0280b48823e9f4ae31b6e4df413b179d44).

### DIFF
--- a/absl/base/internal/unscaledcycleclock.h
+++ b/absl/base/internal/unscaledcycleclock.h
@@ -47,7 +47,7 @@
 // The following platforms have an implementation of a hardware counter.
 #if defined(__i386__) || defined(__x86_64__) || defined(__aarch64__) || \
   defined(__powerpc__) || defined(__ppc__) || \
-  defined(_M_IX86) || defined(_M_X64)
+  defined(_M_IX86) || defined(_M_X64) || defined(__s390x__)
 #define ABSL_HAVE_UNSCALED_CYCLECLOCK_IMPLEMENTATION 1
 #else
 #define ABSL_HAVE_UNSCALED_CYCLECLOCK_IMPLEMENTATION 0
@@ -81,7 +81,7 @@
 // This macro can be used to test if UnscaledCycleClock::Frequency()
 // is NominalCPUFrequency() on a particular platform.
 #if  (defined(__i386__) || defined(__x86_64__) || \
-      defined(_M_IX86) || defined(_M_X64))
+      defined(_M_IX86) || defined(_M_X64)) || defined(__s390x__)
 #define ABSL_INTERNAL_UNSCALED_CYCLECLOCK_FREQUENCY_IS_CPU_FREQUENCY
 #endif
 
--- a/absl/base/internal/unscaledcycleclock.cc
+++ b/absl/base/internal/unscaledcycleclock.cc
@@ -101,6 +101,18 @@ double UnscaledCycleClock::Frequency() {
 #endif
 }
 
+#elif defined(__s390x__)
+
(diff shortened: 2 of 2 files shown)


absl/base/internal/unscaledcycleclock.h
```diff
@@ -81,7 +81,7 @@
 // This macro can be used to test if UnscaledCycleClock::Frequency()
 // is NominalCPUFrequency() on a particular platform.
 #if  (defined(__i386__) || defined(__x86_64__) || \
-      defined(_M_IX86) || defined(_M_X64))
+      defined(_M_IX86) || defined(_M_X64)) || defined(__s390x__)
```
derekmauro, line 84: https://groups.google.com/forum/#!topic/bit.listserv.ibm-main/mYYcbXg0lGY makes me think that the counter retreived by the `stck` instruction is a clock, but not one that runs at the CPU frequency, since it seems constant across chips. Can you point me to documentation that shows otherwise?

derekmauro (commented):
Are you trying to solve some other problem besides a failing unit test? The failing unit test is for a method that not called by Abseil code at all, so it shouldn't be an issue that needs fixing.

Sidong-Wei:
I am also not sure how likely this API will be used by s390x users, but the reason I would like to fix this is that Tensorflow uses this API in its code, see [here](https://github.com/tensorflow/tensorflow/blob/0d37c8f114a0ef4228cee2f371f858d701d7f093/tensorflow/core/platform/default/port.cc#L348), and that also leads to some test case failures with Tensorflow.

absl/base/internal/unscaledcycleclock.h
Sidong-Wei: Thanks for the notes. Honestly speaking, I'm not very familiar with instruction set or cpu counter, so I guess what you are suggesting is that `stck` does not return the correct cpu time so the frequency calculated based on this will not be accurate, am I understanding it correctly? If this is the case, could you please give me a hint of what counter do we need to retrieve here so that I could look into it further? Thank you.

derekmauro:
TensorFlow should not be doing that.

From https://abseil.io/about/compatibility:
> Do not depend upon internal details. This should go without saying: if something is in a namespace or filename/path that includes the word “internal”, you are not allowed to depend upon it. It’s an implementation detail. You cannot friend it, you cannot include it, you cannot mention it or refer to it in any way.

Sidong-Wei:
Thanks for pointing out that, I did feel a little bit strange when seeing them using this API rather than another Tensorflow-native API to retrieve CPU frequency. I may raise an issue in the Tensorflow community about this.

As for the fix itself, I think I will still look into it a little bit to see if I could fix it. If it is too much trouble, I guess I will then close this PR since it is not a critical bug as you mentioned. Will that be okay if I leave it open now?

rogeeff:
Do we need to keep this PR in open state?

Sidong-Wei:
Hi, I have raised the issue with Tensorflow and they have not responded yet, so I will follow up on it. Currently, I think it may not worth investing more time in fixing the cpu frequency issue since it is not actually used (or not supposed to be used) anywhere, so I will close this PR for now and I will consider reopening it if things have changed. Thank you.

PR #730, opened 2020-07-04
Added flag in absl/container/internal/have_see.h to see if AVX512 is …

goldsteinn:
Modified absl/container/internal/have_sse.h so that it would set a flag if AVX512 was supported. Modified absl/container/internal/raw_hash_set.h so that if AVX512 is supported it will use AVX512 instructions in a few places increasing the efficiency of the group instructions.

### DIFF
--- a/absl/container/internal/have_sse.h
+++ b/absl/container/internal/have_sse.h
@@ -34,6 +34,19 @@
 #endif
 #endif
 
+#ifndef ABSL_INTERNAL_RAW_HASH_SET_HAVE_AVX512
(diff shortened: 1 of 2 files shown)


goldsteinn:
@googlebot I fixed it.

goldsteinn:
@googlebot I fixed it.

rogeeff:
Can you provide some benchmarking results demonstrating this change leads to positive change?

goldsteinn:
Seems to only be for a a meaningful improvement in a few use cases so I'll close (reasonable speedup queries, reasonable slow down insert, no effect iterator). Thank you for your time.

PR #735, opened 2020-07-09
CMake: improve super build usage

### LINKED ISSUE #703: Can't export nor install CMake project depending on absl
Depending on absl's through its CMake buildsystem means that one cannot export a depending project's local buildsystem nor install it because abseil doesn't do either. This prevents further downstream projects from, for example, using `find_package(mylib)` by setting `mylib_DIR` to a build directory of mylib, or from installing mylib (and thus its dependencies) to some local path and using that.

Mizux:
Up to now when integrating abseil-cpp as subproject it is impossible:
- to change option value due to lack of CMP0077
- reuse the absl install rules since there are disabled and we can't change this (use of set() in stead of option())

note: These are blocking google/or-tools so currently we have to maintain a patch.
note: May fix #703

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -30,6 +30,9 @@ cmake_policy(SET CMP0057 NEW)
 # Project version variables are the empty string if version is unspecified
 cmake_policy(SET CMP0048 NEW)
 
+# option() honor variables
+cmake_policy(SET CMP0077 NEW)
+
 project(absl CXX)
 
 # Output directory is correct by default for most build setups. However, when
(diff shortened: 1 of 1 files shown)


rogeeff:
Will this restrict usage of our CMake build to new versions of CMake?

Mizux:
TLDR: no

it's more: if you have a recent version of CMake AND you want to use abseil-cpp as subdirectory AND you want to modify option field then you must enable this policy
So here you'll only give more configuration option to user while keeping compatibility with old user cmake

EDIT: this policy has been introduced in CMake 3.13
ref: https://cmake.org/cmake/help/git-stage/manual/cmake-policies.7.html#policies-introduced-by-cmake-3-13
For Googlers: contrary to only have one g3 HEAD, outhere we use versioning scheme so each time CMake want to modify its build system default behaviour (mostly due to mistake) they have to create a policy to introduce the change.

rogeeff:
It appears install is broken now:

+ cmake --build . --target install -- -j
make: *** No rule to make target 'install'.  Stop.

Mizux:
I miss the description field, my bad. I just amend the patch, should be ok now !

vslashg:
Sorry, it looks like our normal pull process fell down somehow.  Your change was incorporated at f624790b7f76ab92fed5ae966abb99a0d455c96f last week, but it lost its link with this PR.  (Our internal source repo serves as the source of truth for Abseil, but our process is supposed to accept PRs from github properly.  That appears not to have happened here.)

I'm closing this PR, since there's nothing left to merge.  Thanks for your contribution, and I'm sorry for the snafu here.

PR #743, opened 2020-07-20
cmake: Check if CMP0077 is available (Fix #742)

### LINKED ISSUE #742: CMake error due to "CMP0077"
From July 15, one of gRPC tests with the HEAD of Abseil started failing with the error message;

```
CMake Error at third_party/abseil-cpp/CMakeLists.txt:34 (cmake_policy):
  Policy "CMP0077" is not known to this version of CMake.
```

This policy is available from CMake 3.13 ([ref](https://cmake.org/cmake/help/git-stage/policy/CMP0077.html)) and we use CMake 3.6.2 which is higher than 3.5 which Abseil claims as the minimum version to support.

Mizux:
Since CMP0077 was introduced in CMake 3.13, I should have first detect if CMP0077 is available before set it to NEW.

FYI only policies >= 66 need this since they have been introduced after CMake 3.5 and may not be available, see https://cmake.org/cmake/help/latest/manual/cmake-policies.7.html

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -31,7 +31,9 @@ cmake_policy(SET CMP0057 NEW)
 cmake_policy(SET CMP0048 NEW)
 
 # option() honor variables
-cmake_policy(SET CMP0077 NEW)
+if(POLICY CMP0077)
+  cmake_policy(SET CMP0077 NEW)
+endif()
 
 project(absl CXX)
 


derekmauro:
I already submitted a fix for this. It will be exported soon.

Mizux:
@derekmauro Thanks ! commented your CL...

PR #746, opened 2020-07-25
absl/debugging: use <execinfo.h> only when available

tpetazzoni:
Instead of relying on __GLIBC__ or other unreliable detection
mechanism, simply detect if <execinfo.h> is available before using the
stacktrace_generic-inl.inc implementation.

Signed-off-by: Thomas Petazzoni <thomas.petazzoni@bootlin.com>

### DIFF
--- a/absl/debugging/CMakeLists.txt
+++ b/absl/debugging/CMakeLists.txt
@@ -14,6 +14,13 @@
 # limitations under the License.
 #
 
+include(CheckIncludeFileCXX)
+
+check_include_file_cxx(execinfo.h HAVE_EXECINFO_H)
+if(HAVE_EXECINFO_H)
+    add_definitions(-DHAVE_EXECINFO_H)
+endif()
+
 absl_cc_library(
   NAME
     stacktrace
--- a/absl/debugging/internal/stacktrace_config.h
+++ b/absl/debugging/internal/stacktrace_config.h
@@ -61,7 +61,7 @@
 # elif defined(__aarch64__)
 #define ABSL_STACKTRACE_INL_HEADER \
     "absl/debugging/internal/stacktrace_aarch64-inl.inc"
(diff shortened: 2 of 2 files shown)


tpetazzoni:
@googlebot I signed it!

derekmauro:
The proposed change only works with CMake. We need something that works with Bazel as well. While we understand that the only reliable thing to do is check for the presence of the header, there is a reason we currently do it the way we do.

What platform are you trying to fix with this change?

tpetazzoni:
The uClibc-ng C library defines __GLIBC__, but its execinfo.h support is optional, so some uClibc-ng toolchains might have execinfo.h, some might not. The musl C library does not implement execinfo.h, and does not define __GLIBC__. So this is handled today on ARM by the check on __GLIBC__ in your code, but it is not handled for any other CPU architecture.

Checking the execinfo.h header is really the only reliable thing to do here. I am not familiar with Bazel, but checking whether a system header is available or not is a pretty standard functionality for build systems, so certainly there should be a way to do that as well ?

derekmauro:
Bazel has different design goals from many traditional build systems, so it doesn't actually have functionality for checking for the existence of a header.

You might be able to do something with `__has_include` to check directly during compilation.

tpetazzoni:
Isn't ``has_include`` quite new (C++17 maybe ?) and therefore only available in very recent compilers?

Also, wouldn't it be possible for the Bazel build system to always define HAVE_EXECINFO_H, which would pretty much preserve the same behavior as today ?

derekmauro:
I see support for `__has_include` going back all the way to gcc 4.9.2 (https://godbolt.org/z/GWdM9E), which is good enough for me. We should prefer mechanisms that are build-system neutral given that Bazel is very important to us, and maintaining two build systems is a burden.

derekmauro:
I'm closing this issue as I'm working on a different change to address this issue.

PR #751, opened 2020-07-28
Use lpthread instead of pthread when linking

guille2718:
In clang-12 linking with -pthread produces a warning:
```
clang-12: warning: argument unused during compilation: '-pthread' [-Wunused-command-line-argument]
```


guille2718:
@googlebot I signed it!

derekmauro:
My testing finds platforms where this fails to compile:
```
ld.lld: error: unable to find library -lpthread
```

Since what you are seeing is only a warning, we can't make this change. I don't know if there is an easy way to fix both the warning and the link error you would get by not linking pthread.

PR #754, opened 2020-08-04
use insertion hint when merging into btree container

ddcc:


derekmauro:
How did you determine that this is improvement? I could imagine that it helps in some cases and hurts in others, depending on the values.

ddcc:
I have an application that tracks an internal key-value map of pointer-value pairs. Among other things, it occasionally extracts a source range of entries into a temporary map, clears a destination range of entries in the original map, and merges back the temporary map, since the two ranges may overlap. When run with a suite of 54 benchmark applications, which have a max working set size of anywhere from zero to a few million entries, I observe a net median speedup of 1 - 2%, so the overall effect isn't large.

rogeeff:
We are looking into an alternative approach for addressing this. I'll close this for now. Please try again in a month or so and if you still observe performance difference, please report.

Issue #734, opened 2020-07-08
[Feature Request] Need basic_string_view


hcoona:
I want basic_string_view for other char types. On windows, wchar is quite useful talking with Windows API. It would be great to have absl::basic_string_view to support it.

rogeeff:
Long story short, we do not have an incentive to support this. This is pretty massive effort, including all the other string interfaces Abseil supports and Google first party users did not request it. In a few years we'll all migrate to C++17's std::string_view, which does everything you want already. Can you use that?

hcoona:
For who still interested to this topic, please go #124 for a more detailed discussion.

Issue #124, opened 2018-05-17
Feature request: absl::wstring_view


rongjiecomputer:
As a Windows user, I use `std::wstring` to manipulate Windows file path. I do like to use `{std,absl}::wstring_view` for efficiency.

`absl/strings/string_view` currently implements `absl::string_view` class directly rather than implementing `absl::basic_string_view<CharT>` then `using string_view = basic_string_view<char>`.

The addition of `absl::u16string_view` and `absl::u32string_view` would be nice too, but I don't need them yet.

If Abseil team fear that people might do silly thing with `basic_string_view`, then just put it in `internal` namespace, I doubt if anyone actually wants to use `basic_string_view` directly.

rongjiecomputer:
Also: https://groups.google.com/a/chromium.org/d/msg/cxx/pxn9Iu1iMX0/-c1KmL6vCAAJ

mbxx:
Thank you for the feature request!  This seems like a completely understandable thing to want, but I'm worried that this may prove difficult for us to implement and that even if we did, it wouldn't be as production-tested as the rest of Abseil.

Abseil has 250M lines of code that depends on it inside of Google.  Part of refactoring absl::string_view into an alias for absl::basic_string_view<char> would be making sure the change doesn't break those 250M lines of code.  This should mostly just work, but having absl::string_view be a class isn't quite the same thing as having absl::string_view be an alias for a template.  We'd have to manually fix any problems that might arise from the change and I'm nervous about how much work that might be.  It's something we could handle, but that's effort that would have to be diverted away from getting other pieces of code ready for release.

Even if we did that refactoring, I'd feel a little weird about releasing an absl::wstring_view.  Right now, all of the code that's a part of Abseil is thoroughly tested in Google production.  absl::wstring_view would not be.  That might be something we can get past, but it's something we'd have to talk about as a team because it would be a departure from what we currently offer.

It's very interesting that Chromium is interested in picking up Abseil as a dependency and we'd certainly love to chat with them about it, but if they're close to having C++17 support, then I suspect that they may rather pick up std::basic_string_view.

I think implementing an absl::wstring_view is something that we'd do with unlimited time and resources, but I also think that the things that we're working on now (like releasing our hash table implementation) are in higher demand.  I'm going to leave this feature request open to gauge interest absl::wstring_view, though, because I'd like to know if I'm wrong about that.

rongjiecomputer:
[...]

PR #757, opened 2020-08-05
[string_view] Use member-types instead of hardcoding char type

Yannic:
This change replaces uses of `char` with `string_view::value_type`.
This *may* allow making `string_view` a template and support different
value types.

Updates #124, #734


rogeeff:
We do not plan to make string_view a template and making these changes sends wrong message. C++ 17's string_view is a template if you need one.

Issue #771, opened 2020-08-20
signed/unsigned compare in btree.h


faceprint:
**Describe the bug**

Since the enum containing kNodeValues was changed to be explicitly uint32_t, there is now a signed/unsigned comparison:

external/abseil/absl/container/internal/btree.h:2439:19: error: comparison of integer expressions of different signedness: 'int' and 'unsigned int' [-Werror=sign-compare]
 2439 |     if (max_count < kNodeValues) {
      |         ~~~~~~~~~~^~~~~~~~~~~~~

**Steps to reproduce the bug**

Compile revision c6b3f2cf583d8fec124f9d70a172b513363506fe with sign comparison warnings turned on

**What version of Abseil are you using?**
HEAD (c6b3f2cf583d8fec124f9d70a172b513363506fe and later)

**What operating system and version are you using**
Linux (RHEL 6)

**What compiler and version are you using?**
gcc 9.1.0
clang 8.0.1
clang 9.0.0

**What build system are you using?**
cmake 3.15.0

**Additional context**

asraa:
I also ran into this problem building Abseil from HEAD and using clang-12

derekmauro:
A fix for this should be exported to GitHub today.

faceprint:
Unfortunately, there are more issues, my build system just stopped after the first failure:

```
external/abseil/absl/container/internal/btree.h:2252:45: fatal error: comparison of integers of different signs: 'int' and 
'absl::container_internal::btree<absl::container_internal::set_params<StringType, StringCmp, std::__1::allocator<StringType >, 256, false> >::(anonymous enum at <external/abseil/absl/container/internal/btree.h>:1052:3)' [-Wsign-compare]
                      (1 + (insert_position < kNodeValues));
```

I attempted to fix so I could provide all the spots that need fixing, but the int here is set from iterator->position, which is purposefully signed, as it uses -1 as a special value for invalid positions:

```
  1006   // The position within the node of the tree the iterator is pointing at.                                                                                                                                               
  1007   // NOTE: this is an int rather than a field_type because iterators can point                                                                                                                                           
  1008   // to invalid positions (such as -1) in certain circumstances.                                                                                                                                                         
  1009   int position;   
```

I'm not sure what your preferred solution here would be.

termizil:
Open

termizil:
Signed

termizil:
#[https://github.com/abseil/abseil-cpp/issues/771#issue-682705311]()

[...]

PR #773, opened 2020-08-27
Fix signed/unsigned comparison in btree

LevyCory:
This is a fix for issue #771


derekmauro:
Thanks for the patch, but an internal change as already been accepted to fix this, and it should be exported today. Sorry for the confusion.

PR #780, opened 2020-09-12
Added missing asserts for seq.index() < capacity_ and unified their u…

ethician:
In a highly specific situation the call to find() was found to be stuck in an infinite cycle as it never checked whether `probe_seq::index_` as accessed by `seq.index()` exceeded `capacity_`.

This situation arose after a call to a faulty implementation of `std::memset` (using uClibc's v0.9.32.1 memset, ARM platform) in `reset_ctrl()` which ended up setting the control block to `0xffffff80 0xffffff80 ...` as opposed to `0x80808080 0x80808080 ...`.

The reordering of calls in `find_first_non_full()` is simply to prevent one extra, unnecessary traversal.

### DIFF
--- a/absl/container/internal/raw_hash_set.h
+++ b/absl/container/internal/raw_hash_set.h
@@ -1327,6 +1327,7 @@ class raw_hash_set {
       }
       if (ABSL_PREDICT_TRUE(g.MatchEmpty())) return end();
       seq.next();
+      assert(seq.index() < capacity_ && "full table!");
     }
   }
   template <class K = key_type>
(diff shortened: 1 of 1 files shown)


ethician:
@googlebot I fixed it.

absl/container/internal/raw_hash_set.h
```diff
@@ -1857,6 +1859,7 @@ struct HashtableDebugAccess<Set, absl::void_t<typename Set::raw_hash_set>> {
       }
       if (g.MatchEmpty()) return num_probes;
       seq.next();
+      assert(seq.index() < capacity_ && "full table!");
```
derekmauro: This doesn't build, so how was this tested?

ethician: Not sure how I snuck that line in there, this one shouldn't be there. I will close this and resubmit.

PR #781, opened 2020-09-13
Added missing asserts for seq.index() < capacity_ and unified their usage based on has_element().

ethician:
In a highly specific situation the call to find() was found to be stuck in an infinite cycle as it never checked whether `probe_seq::index_` as accessed by `seq.index()` exceeded `capacity_`.

This situation arose after a call to a faulty implementation of `std::memset` (using uClibc's v0.9.32.1 memset, ARM platform) in `reset_ctrl()` which ended up setting the control block to `0xffffff80 0xffffff80 ...` as opposed to `0x80808080 0x80808080 ...`.

The reordering of calls in `find_first_non_full()` is simply to prevent one extra, unnecessary traversal.

### DIFF
--- a/absl/container/internal/raw_hash_set.h
+++ b/absl/container/internal/raw_hash_set.h
@@ -1327,6 +1327,7 @@ class raw_hash_set {
       }
       if (ABSL_PREDICT_TRUE(g.MatchEmpty())) return end();
       seq.next();
+      assert(seq.index() < capacity_ && "full table!");
     }
   }
   template <class K = key_type>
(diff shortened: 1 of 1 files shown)


fowles:
LGTM for the code.  Derek can you pull this in and to test run?

derekmauro:
Thanks for reviewing. It is actually @rogeeff's turn to pull this in.

PK-Patomsiri (approved):

derekmauro (approved):

PR #797, opened 2020-09-29
fix build dll

### LINKED ISSUE #796: abseil-cpp\CMake\AbseilHelpers.cmake
**Describe the bug**

When I use Abseil as a third-party library with source code integration, I want to compile a dynamic library under Windows. At this time, the problem arises. The target in abseil will also be compiled as dll_dep, and I have set BUILD_SHARED_LIBS=ON. Through debugging, I found that it turned out to be the reason for the setting of _NAME. When abseil is used as a third-party library, _NAME will be set to "absl_${ABSL_CC_LIB_NAME}", which is judged to be not part of the dll in the subsequent absl_internal_dll_contains function. These targets are thus identified as dll_dep, and then compiled with the ABSL_CONSUME_DLL macro definition attached, resulting in compilation errors.

ulinzeng:


ulinzeng:
@googlebot I signed it!

suertreus:
Please amend the commit description to describe what was broken, how this change fixes it, and how to validate that the fix works.  If the answers are long, you can put a summary in the commit message and open an issue with more details.

suertreus:
Never mind; I found and linked the related issue.

@asoffer can you take a look at this PR?

rogeeff (approved):

PR #815, opened 2020-10-13
fix compile fails with asan and -Wredundant-decls

hellowangsai:
If compile with asan, unaligned_access.h redeclares the same sanitizer
function with sanitizer/common_interface_defs.h.

This patch removes these declarations and includes sanitizer header
file to avoid this warning.


derekmauro:
Thanks for the patch, but this has already been fixed at head.

PR #861, opened 2020-12-12
container: fix introduction for node_hash_set.h

daohu527:
container: fix introduction for node_hash_set.h

Fixed the wrong comment description "node_hash_map" to "node_hash_set"


mbxx:
Thank you!  Looks good to me, merging.

mbxx:
It turns out there were a number of other copy-and-paste mistakes in the file; we've gone ahead and fixed those, too.  Thank you so much for catching this!

PR #863, opened 2020-12-13
Include the status library into the main README.

ipoletaev:
It is a bit confusing that it is not listed here although already publically released.


sajjadanari (approved):

mbxx:
Thanks!  Yeah, we seem to have forgotten to update this file.  I'll get this added.

PR #884, opened 2021-01-11
Don't assume Windows means MSVC in the bazel BUILD files.

pmuetschard:
When selecting the compiler flags, use the compiler setting, rather than
the platform setting, to detect if MSVC flags are required.

### DIFF
--- a/absl/base/BUILD.bazel
+++ b/absl/base/BUILD.bazel
@@ -160,7 +160,7 @@ cc_library(
     ],
     copts = ABSL_DEFAULT_COPTS,
     linkopts = select({
-        "//absl:windows": [],
+        "//absl:msvc_compat_compiler": [],
         "//absl:wasm": [],
         "//conditions:default": ["-pthread"],
     }) + ABSL_DEFAULT_LINKOPTS,
@@ -220,7 +220,7 @@ cc_library(
     ],
     copts = ABSL_DEFAULT_COPTS,
     linkopts = select({
-        "//absl:windows": [
+        "//absl:msvc_compat_compiler": [
             "-DEFAULTLIB:advapi32.lib",
         ],
         "//absl:wasm": [],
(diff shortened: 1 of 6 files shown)


pmuetschard:
Ooops. The correct value of the flag is `msvc-cl`. Sorry about that. Please take another look.

absl/base/BUILD.bazel
```diff
@@ -160,7 +160,7 @@ cc_library(
     ],
     copts = ABSL_DEFAULT_COPTS,
     linkopts = select({
-        "//absl:windows": [],
+        "//absl:msvc_compiler": [],
         "//absl:wasm": [],
         "//conditions:default": ["-pthread"],
```
suertreus, line 165: Seems like `windows` might have been better here - it doesn't make sense to pass `-pthread` on any Windows toolchain, does it?

suertreus (commented):
I worry that as written, this will break clang-cl builds that are currently (correctly) sharing behavior with MSVC builds.

What's your motivation for this change?  Are you trying to build for Windows using some other compiler and having problems?  It seems like some of these things ought to be switched on the OS and some on the compiler, rather than all one way or all the other.

pmuetschard:
> I worry that as written, this will break clang-cl builds that are currently (correctly) sharing behavior with MSVC builds.

I've changed so that the rules now work for both msvc and clang-cl.

> What's your motivation for this change? Are you trying to build for Windows using some other compiler and having problems? It seems like some of these things ought to be switched on the OS and some on the compiler, rather than all one way or all the other.

We are using gRPC, which now depends on abseil, but we are also currently forced into using the MinGW gcc toolchain. With this change, the parts of abseil that gRPC uses build ok with MinGW gcc.

suertreus:
Apologies, I didn't notice you are a Googler.  Can you please write this change in Piper instead?  You can send it to me for review.

kristiii12 (approved):

PR #889, opened 2021-01-26
Add support for Elbrus 2000 (e2k)

SSE4:


absl/base/attributes.h
```diff
@@ -123,8 +123,12 @@
 #define ABSL_ATTRIBUTE_NO_TAIL_CALL __attribute__((disable_tail_calls))
 #elif defined(__GNUC__) && !defined(__clang__)
 #define ABSL_HAVE_ATTRIBUTE_NO_TAIL_CALL 1
+#if defined(__e2k__)
+#define ABSL_ATTRIBUTE_NO_TAIL_CALL
+#else
 #define ABSL_ATTRIBUTE_NO_TAIL_CALL \
   __attribute__((optimize("no-optimize-sibling-calls")))
+#endif
```
fowles: instead of this, please add `&& !defined(__e2k__)` on line 124

fowles (commented):
We are willing to accept this patch (after it is cleaned up), but we want to make clear that we are only supporting e2k on a best effort basis.

absl/base/attributes.h
SSE4: changed

akagamy55: Close:(Atributos.h)

akagamy55: Close:(puertos e2k: e2k)

PR #891, opened 2021-01-29
Do not set mvsc linker flags for clang-cl (fixes #874)

### LINKED ISSUE #874: all cmake exports are wrong for abseil (windows + clang)
https://github.com/abseil/abseil-cpp/blob/4fd9a1ec5077daac14eeee05df931d658ec0b7b8/CMake/AbseilHelpers.cmake#L284

```cmake
 target_link_libraries(${_NAME}
      INTERFACE
        ${ABSL_CC_LIB_DEPS}
        ${ABSL_CC_LIB_LINKOPTS}
        ${ABSL_DEFAULT_LINKOPTS}
    )
```

cause: https://github.com/microsoft/vcpkg/issues/13945

You should never export in a public interface a link options.

Why ? 

Because user can perfectly install `abseil` using MSVC

and link it using `clang` then: https://github.com/abseil/abseil-cpp/blob/e9b9e38f67a008d66133535a72ada843bd66013f/absl/copts/GENERATED_AbseilCopts.cmake#L137 this one is wrong too

```cpp
int main() { return 0; }
```

```bash
$ clang -ignore:4221 -o foo foo.cpp
clang: error: unknown argument: '-ignore:4221'
```

In `abslTargets.cmake` after installation we can see:

```
set_target_properties(absl::log_severity PROPERTIES
  INTERFACE_INCLUDE_DIRECTORIES "${_IMPORT_PREFIX}/include"
  INTERFACE_LINK_LIBRARIES "absl::core_headers;\$<LINK_ONLY:-ignore:4221>"
)
```

Which is wrong

Does that mean that you don't have any Continuous Integration which tests a lib compiled with MSVC linked on Clang on windows ?

to reproduce:

- Install Abseil with MSVC on a computer + CMake
- Try to use Abseil with clang on windows + CMake

Chronial:
`lld-link` does not support all of the comand line flags of link.exe. Most importantly it doesn't support the `-ignore:4221` flag set by abseil.

This causes CMake clang-cl builds to fail with `clang++: error: unknown argument: '-ignore:4221'`.


Chronial:
@googlebot I signed it!

PR #915, opened 2021-03-16
Fixed compilation on windows

kalman5:
This change allows the compilation on windows/cygwin


derekmauro:
We need the CLA to accept this, but I have a different fix in mind, so I'm going to close this. I'll submit a fix shortly.

kalman5:
Was this fixed in another way then? @derekmauro

PR #916, opened 2021-03-18
Fix C++/CLI build problem

Chronial:
This fixes a build problem when using `btree_set` with C++/CLI.
IMO it's also slightly better style.


joe198:
ok upgraded

joe198:
ok upgraded  all fix there problems

joe198:
OK

PR #925, opened 2021-03-30
Add missing `add_subdirectory()` call for "cleanup"

cblichmann:
Since `absl::Cleanup` is now public, it should also be included
in the `absl/CMakeLists.txt` file.

Signed-off-by: Christian Blichmann <cblichmann@google.com>


derekmauro:
I will also backport this fix into the 20210324 LTS branch.

PR #926, opened 2021-03-30
Set minimum cpp version in CMakeLists.txt.

innerNULL:
Set minimum cpp version in 'CMakeLists.txt'.


innerNULL:
> @googlebot I signed it!

@googlebot I signed it!

PR #927, opened 2021-03-30
Set minimum cpp version in CMakeLists.txt.

innerNULL:
Set minimum cpp version in CMakeLists.txt.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -21,6 +21,8 @@
 # and then issuing `yum install cmake3` on the command line.
 cmake_minimum_required(VERSION 3.5)
 
+set(CMAKE_CXX_STANDARD 11)
+
 # Compiler id for Apple Clang is now AppleClang.
 if (POLICY CMP0025)
   cmake_policy(SET CMP0025 NEW)


derekmauro:
I don't think we actually want to do this. It is important that all projects using Abseil use the same C++ version, and this would be pinning Abseil to C++11, right?

What problem are you trying to solve?

innerNULL:
> I don't think we actually want to do this. It is important that all projects using Abseil use the same C++ version, and this would be pinning Abseil to C++11, right?
> 
> What problem are you trying to solve?

@derekmauro Hi~I think this is necessary, since for example, in one of my cmake script, I will download abseil and build abseil lib with its CMakeLists.txt, without `set(CMAKE_CXX_STANDARD 11)`, I will get following error in 'absl/base/policy_checks.h:77:2:':
`#error "C++ versions less than C++11 are not supported."`
So I think adding the requirment of cpp version will be helpful for integration of abseil

Thanks

derekmauro:
If we were to add `set(CMAKE_CXX_STANDARD 11)`, how would we build Abseil in, for instance, C++14 mode?

innerNULL:
> If we were to add `set(CMAKE_CXX_STANDARD 11)`, how would we build Abseil in, for instance, C++14 mode?

Oh my bad, just hang in my mind about this point~I can pass this parameter with cmake flag, sorry for waste your time.

Issue #933, opened 2021-04-07
Failure signal handler should call writerfn with null buffer at the end


yasushi-saito:
**Describe the bug**

The documentation for FailureSignalHandlerOptions in absl/debugging/failure_signal_handller.h says that writerfn may be called with null buffer to let the callee flush buffered data. But that never happens. It will help us if absl explicitly calls writerfn(nullptr, 0) just before crashing. 

Our use case is that we want to caputure the crash message as one long string and log it specially marked.

I can create a PR if you think this is a good idea.

**What version of Abseil are you using?**

7f4d0ba

**What operating system and version are you using**

Ubuntu 20.04, x86-64.

**What compiler and version are you using?**

clang-10.0

**What build system are you using?**

bazel 3.5

derekmauro:
I vaguely remember writing that comment in the documentation, but I have no idea why it seems like it was never implemented. I would take this as a PR.

derekmauro:
The fix has been merged.

PR #938, opened 2021-04-12
Call FailureSignalHandlerOptions.writenfn with nullptr at the end

yasushi-saito:
This behavior has already been alluded to in the document, but it hasn't been
implemented.

This PR calls changes the failure signal handler to call writerfn(nullptr) at
the end of the message block. It also clarifies the documentation.

https://github.com/abseil/abseil-cpp/issues/933


derekmauro:
Merged, thanks!

Issue #722, opened 2020-06-25
Remove use of deprecated allocator members and types


DmitryTsyganov:
There are uses of allocator members and types that were deprecated in C++17 and removed in C++20. Examples:

Members:
https://github.com/abseil/abseil-cpp/blob/10cb35e459f5ecca5b2ff107635da0bfa41011b4/absl/memory/memory.h#L576
[https://en.cppreference.com/w/cpp/memory/allocator/construct](https://en.cppreference.com/w/cpp/memory/allocator/construct)

https://github.com/abseil/abseil-cpp/blob/10cb35e459f5ecca5b2ff107635da0bfa41011b4/absl/memory/memory.h#L587
[https://en.cppreference.com/w/cpp/memory/allocator/destroy](https://en.cppreference.com/w/cpp/memory/allocator/destroy)

Types:
https://github.com/abseil/abseil-cpp/blob/10cb35e459f5ecca5b2ff107635da0bfa41011b4/absl/memory/memory.h#L355
[rebind in **Member types**](https://en.cppreference.com/w/cpp/memory/allocator)

This makes abseil fail to compile in C++20 mode and in C++17 mode if warnings are treated as errors.

Can be replaced with their substitutes from [https://en.cppreference.com/w/cpp/memory/allocator_traits](https://en.cppreference.com/w/cpp/memory/allocator_traits)

sbenzaquen:
All of those are guarded. They are only used if present, as it is specified in the contract for `allocator_traits`.
Is there a repro of this actually failing in C++20?
The traits should just fallback gracefully.

DmitryTsyganov:
It does work, my bad.
The problem is only with  C++17 if warnings are treated as errors.

sbenzaquen:
This warning should go away after c03c18e7fef8d06eb6c445f5607af533e69cadfe.

BenFrantzDale:
> It does work, my bad.
> The problem is only with C++17 if warnings are treated as errors.

Yeah, I'm on C++17 with warnings treated as errors and this is a pain. I'd rather not have to `#define _SILENCE_CXX17_OLD_ALLOCATOR_MEMBERS_DEPRECATION_WARNING`.

GindaChen:
> Is there a repro of this actually failing in C++20?

@sbenzaquen We happen to encounter a C++20 build failure [in this repo issue](https://github.com/UWHustle/hustle/issues/97). Right now we are trying to separate a small enough example as a repro.

[...]

PR #946, opened 2021-04-23
Avoid directly accessing allocator members

bstaletic:
C++11 introduced allocator_traits and since C++17 some allocator members have been deprecated.

These few lines produced thousands of lines of warnings and explanations on MSVC when compiled with `/std:c++17`.


derekmauro:
This file contains a polyfill of `std::allocator_traits`. Changing the implementation to call `std::allocator_traits` would defeat the purpose. It seems that MSVC is warning about a type not owned by the standard library.

Can you take the discussion to #722 with a repro?

PR #947, opened 2021-04-23
Fix MSVC compiler error when address sanitizers are enabled

jeremyong:
The `ABSL_ADDRESS_SANITIZER_REDZONE` macro specifies the alignment of the `name` struct with a non-portable alignment attribute. When address sanitizers on MSVC are enabled, compiles will fail on this line as a result. Since `alignas` has been supported since C++11, I believe this should be a straightforward substitution.


derekmauro:
Thank you. I'm just trying to verify that no one is using this from C code, which might take a day. Assuming that is the case I'll merge this as is. Otherwise we might need a small tweak.

alfredesigns (approved):

derekmauro (approved):

jeremyong:
Thanks!

PR #963, opened 2021-05-22
Use CMAKE_INSTALL_FULL_{LIBDIR,INCLUDEDIR}.

gebner:
The `CMAKE_INSTALL_LIBDIR` and `CMAKE_INSTALL_INCLUDEDIR` variable do not need to be relative paths.  If they are absolute, then the concatenation `${prefix}/${CMAKE_INSTALL_LIBDIR}` is incorrect.  (This is the case on NixOS.)

This PR replaces this concatenation with `${CMAKE_INSTALL_FULL_LIBDIR}`, which always refers to an absolute path.  https://github.com/jtojnar/cmake-snips#assuming-cmake_install_dir-is-relative-path


gebner:
@googlebot I signed it!

derekmauro:
Thank you! Merged!

PR #974, opened 2021-06-16
Update blocking_counter.h : Fix for building with GCC-11

samcom12:
Fix for building with GCC 11.1.0

### DIFF
--- a/absl/synchronization/blocking_counter.h
+++ b/absl/synchronization/blocking_counter.h
@@ -21,6 +21,7 @@
 #define ABSL_SYNCHRONIZATION_BLOCKING_COUNTER_H_
 
 #include <atomic>
+#include <limits>
 
 #include "absl/base/thread_annotations.h"
 #include "absl/synchronization/mutex.h"


absl/synchronization/blocking_counter.h
```diff
@@ -21,6 +21,7 @@
 #define ABSL_SYNCHRONIZATION_BLOCKING_COUNTER_H_
 
 #include <atomic>
+#include <limits>
```
derekmauro, line 24: Can you share the compiler error message? At first glance I do not see any symbol from `<limits>` that is used in this file.

samcom12: [spack-build-out.txt](https://github.com/abseil/abseil-cpp/files/6670624/spack-build-out.txt)


image

derekmauro:
The compiler error above for the missing symbol occurs in `graphcycles.cc`, not in `blocking_counter.h`, so the include of `<limits>` belongs in `graphcycles.cc`, not `blocking_counter.h`.

A fix for this was committed in 5bf048b8425cc0a342e4647932de19e25ffd6ad7, about 8 months ago.

Further, this fix has already been backported to the LTS release you are using, you just need to upgrade to https://github.com/abseil/abseil-cpp/releases/tag/20200923.3 if you really need an LTS version, or just use the latest release.

PR #977, opened 2021-06-19
Fix shadow member warnings in randen_hwaes.cc

mtrofin:
The warning is due to the name of the Vector128 ctor parameter. Using `v` instead, which I see used elsewhere (e.g. line 290)


derekmauro:
Closing because the internal CL has been approved.

PR #986, opened 2021-07-15
CMake: option to use cxx_std_11 (minimum) that propagates.

Lindurion:
See https://github.com/abseil/abseil-cpp/issues/259 for context (and more commentary on PUBLIC/INTERFACE propagation).

This change introduces the `ABSL_PROPAGATE_CXX_STD` option (default to OFF for now, though it prints a warning in CMake 2.8+ builds that a future Abseil release will default to ON). When enabled, all Abseil CMake targets will set the `cxx_std_11` target meta feature (which will then propagate to targets that depend upon Abseil) rather than setting the target-level `CXX_STANDARD` property to `CMAKE_CXX_STANDARD` (which is the default value anyway), which doesn't propagate.

Updates README documentation to clarify behavior and with a different example recommendation for library projects (which should generally leave `CMAKE_CXX_STANDARD` to the root application project).

See https://crascit.com/2015/03/28/enabling-cxx11-in-cmake/ for a useful overview of these different CMake features surrounding C++ standard version configuration.

### DIFF
--- a/CMake/AbseilHelpers.cmake
+++ b/CMake/AbseilHelpers.cmake
@@ -254,9 +254,23 @@ Cflags: -I\${includedir}${PC_CFLAGS}\n")
       set_property(TARGET ${_NAME} PROPERTY FOLDER ${ABSL_IDE_FOLDER}/internal)
     endif()
 
-    # INTERFACE libraries can't have the CXX_STANDARD property set
-    set_property(TARGET ${_NAME} PROPERTY CXX_STANDARD ${ABSL_CXX_STANDARD})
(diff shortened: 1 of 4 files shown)


derekmauro:
Thank you for sending this pull request, and sorry I haven't got to it yet. I've wanted something like this for a while, but it may take me a little while longer to evaluate it.

Lindurion:
@derekmauro No problem -- not directly blocking anything here, and this is as much about clarifying CMake behavior and best practices in documentation as the small build change itself.

derekmauro (approved):

derekmauro:
Very happy with this, thanks!

PR #998, opened 2021-08-07
cmake: add ABSL_BUILD_TESTONLY_LIBS option (fixes #997)

### LINKED ISSUE #997: CMake: a TESTONLY library is not built if BUILD_TESTING is disabled.
**Describe the bug**

In my CMake project, I depend on Abseil, but have `BUILD_TESTING` disabled, so Abseil tests are not built.

I have my own tests. In my tests I use GoogleTest/GoogleMock and Abseil.  In particular, one of my tests depends on `absl::random_mocking_bit_gen` target.  This target is a TESTONLY library that is meant to be used by any tests, including mine.

But as I keep `BUILD_TESTING` disabled, TESTONLY libraries are not build and I'm unable to depend on Abseil test support code in my tests.

Looks like building test support libraries should be separated from building full-fledged tests.

**What build system are you using?**

CMake

anpol:
This option is meant to enable building TESTONLY libraries without
building test executables.

An external project may depend on an Abseil TESTONLY library like
`absl::random_mocking_bit_gen`.  A TESTONLY library may provide valuable
test support classes for writing tests within that external project.

Building a TESTONLY library requires a dependency on GoogleTest.
However, there may be no need to build full-fledged Abseil tests.

The behavior of a BUILD_TESTING option is not changed, it still enables
building both TESTONLY libraries and test executables.

### DIFF
--- a/CMake/AbseilHelpers.cmake
+++ b/CMake/AbseilHelpers.cmake
@@ -82,7 +82,7 @@ function(absl_cc_library)
     ${ARGN}
   )
 
-  if(ABSL_CC_LIB_TESTONLY AND NOT BUILD_TESTING)
+  if(ABSL_CC_LIB_TESTONLY AND NOT (BUILD_TESTING OR ABSL_BUILD_TESTONLY_LIBS))
     return()
   endif()
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -130,7 +130,10 @@ set(ABSL_LOCAL_GOOGLETEST_DIR "/usr/src/googletest" CACHE PATH
   "If ABSL_USE_GOOGLETEST_HEAD is OFF and ABSL_GOOGLETEST_URL is not set, specifies the directory of a local GoogleTest checkout."
   )
 
-if(BUILD_TESTING)
+option(ABSL_BUILD_TESTONLY_LIBS
+  "If ON, Abseil will build TESTONLY libraries even if Abseil tests are not built." OFF)
+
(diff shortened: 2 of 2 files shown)


anpol:
Issue #997

bdr2486 (approved):

bdr2486 (approved):

derekmauro:
Thanks for alerting us to this issue.

The `TESTONLY` attribute is meant to be a copy of [Bazel's `testonly` attribute](https://docs.bazel.build/versions/main/be/common-definitions.html). Unfortunately somehow it got a different meaning in the CMake build.

With Bazel, if a target has `testonly=1`, only other test targets are allowed to depend on it.

The CMake meaning appears to be "this target will only be built if `BUILD_TESTING=ON`".

For this reason I don't think it makes sense to have `TESTONLY` in CMake at all. The CMake build system doesn't enforce the Bazel meaning at all, and it is confusing to have this thing that has a different meaning exist at all.

I think what we probably want instead is either something like `PRIVATE_TEST_LIB` that has a similar meaning but isn't applied to something like `mocking_bit_gen`, which is meant to be public. Or possible some native CMake mechanism to exclude libraries that aren't meant to be installed.

anpol:
To summarize,

* In Bazel, If TESTONLY is True, only testonly targets (such as tests and test support libraries) can depend on this target.
* In CMake it could be implemented by an imaginable module that would enforce such kind of relationship between targets.

Closing this in favor of #1084 for now.

Issue #1000, opened 2021-08-12
Supporting incompatible_use_platforms_repo_for_constraints?


elliottt:
It looks like the eventual direction for specifying platform constraints is to explicitly depend on https://github.com/bazelbuild/platforms, instead of using the `@bazel_tools//platforms` targets. To support migration towards this future state bazel supports the `--incompatible_use_platforms_repo_for_constraints` flag, tracked in https://github.com/bazelbuild/bazel/issues/8622. Turning this flag on breaks builds that depend on abseil, as it currently relies on `@bazel_tools//platforms:osx` and `@bazel_tools//platforms:ios` in a few places.

Patching those uses in `http_archive` is a workable solution, but it would be great to avoid that. Does the abseil project have plans to migrate to a build configuration that would rely on `@platforms//os:osx` and `@platforms//os:ios` instead?

PR #1001, opened 2021-08-12
Switch to depending on `@platforms` for os constraints

elliottt:
Following on from issue #1000, this is the change that I needed to make locally to switch abseil over to building with `--incompatible_use_platforms_repo_for_constraints`.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -42,3 +42,13 @@ http_archive(
     strip_prefix = "rules_cc-daf6ace7cfeacd6a83e9ff2ed659f416537b6c74",
     urls = ["https://github.com/bazelbuild/rules_cc/archive/daf6ace7cfeacd6a83e9ff2ed659f416537b6c74.zip"],
 )
+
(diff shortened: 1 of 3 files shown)


derekmauro:
Thank you. I didn't know about this change to Bazel.

You need to sign the CLA in order for us to accept this pull request. If you can't sign it, I can make the change.

By the way, I think you also need to add a dependency on the repo in the WORKSPACE file.

elliottt:
> Thank you. I didn't know about this change to Bazel.
> 
> You need to sign the CLA in order for us to accept this pull request. If you can't sign it, I can make the change.

Will do!
> 
> By the way, I think you also need to add a dependency on the repo in the WORKSPACE file.

I'll add that as well, thanks for pointing that out!

derekmauro:
By policy I have to close this because you haven't signed the CLA. However, a change has been pushed to use the platforms repository.

elliottt:
That works for me. I was waiting on Stripe's legal department to approve me, but this is much easier :)

PR #1002, opened 2021-08-16
Add missing ABSL_DLL for a few functions

zcbenz:
The `throw_bad_optional_access` and `ThrowBadVariantAccess` are used in a few inlined functions and should have the `ABSL_DLL` prefix, otherwise linking with `absl.dll` will fail with errors like:

```
error LNK2019: unresolved external symbol "void __cdecl absl::
optional_internal::throw_bad_optional_access(void)" (?throw_bad_optional_access@opt
ional_internal@absl@@YAXXZ) referenced in function "public: class base::TimeDelta &
 __cdecl absl::optional<class base::TimeDelta>::value(void)& " (?value@?$optional@V
TimeDelta@base@@@absl@@QEGAAAEAVTimeDelta@base@@XZ)\r\n'
```

### DIFF
--- a/absl/types/bad_optional_access.h
+++ b/absl/types/bad_optional_access.h
@@ -67,7 +67,7 @@ class bad_optional_access : public std::exception {
 namespace optional_internal {
 
 // throw delegator
-[[noreturn]] void throw_bad_optional_access();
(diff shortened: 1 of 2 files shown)


derekmauro:
Thank you. This should be merged soon.

PR #1003, opened 2021-08-17
Use full include paths in time/internal/cctz

zcbenz:
This makes the coding style of `time/internal/cctz` align with other files, and without using full paths in includes it will be harder for other projects to build together with abseil-cpp.

### DIFF
--- a/absl/time/internal/cctz/src/cctz_benchmark.cc
+++ b/absl/time/internal/cctz/src/cctz_benchmark.cc
@@ -23,7 +23,7 @@
 #include "benchmark/benchmark.h"
 #include "absl/time/internal/cctz/include/cctz/civil_time.h"
 #include "absl/time/internal/cctz/include/cctz/time_zone.h"
-#include "time_zone_impl.h"
+#include "absl/time/internal/cctz/src/time_zone_impl.h"
 
 namespace {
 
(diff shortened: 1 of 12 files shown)


derekmauro:
> without using full paths in includes it will be harder for other projects to build together with abseil-cpp.

Can you explain further? Is this causing a problem?

zcbenz:
I'm merging all abseil's `.cc` code into one file for compilation and distribution (which is similar to Chromium's jumbo build (which had been removed)), so searching headers from the source code file's directory does not work.

I know this is not something that abseil means to support, but it does reduce compilation time to 1/10 for me :). Also this change makes the code complies to Google C++ Style Guide so I think it might be good to upstream:
https://google.github.io/styleguide/cppguide.html#Names_and_Order_of_Includes
> All of a project's header files should be listed as descendants of the project's source directory

derekmauro:
The CCTZ code is pulled into Abseil from https://github.com/google/cctz. We have a script that edits the code to make it work in Abseil. I'd have to go modify the script to "import" this pull request. I understand why you want this, but I have some work to do first. It's not high priority for me at this point, but I may get to it soon.

zcbenz:
That makes sense, thanks for taking care of this!

PR #1013, opened 2021-09-08
Disable abseil tests if abseil is a subprojects.

mb1248:
If abseil is a CMAKE subproject the abseil tests are not longer build when BUILD_TESTING is on for the parent project.

If someone uses abseil as a subproject, they probably do not want abseil tests to build.
In the rare case that someone really does want to enable testing on both packages, the new option ABSL_BUILD_TESTING can be used to override.


mb1248:
@googlebot I fixed it.

PR #1015, opened 2021-09-09
Include immintrin.h instead of wmmintrin.h

stanhu:
`immintrin.h` is the de-factor standard header for clang and GCC to include Intel intrinsics. Using this header avoids requiring the compiler to use the `-maes` and `-msse4.1` compiler options on systems that may not have AES or SSE instruction support.

clang: As seen in https://github.com/llvm-mirror/clang/blob/master/lib/Headers/immintrin.h,
specific intrinsic header files are conditionally included depending on whether the feature is available.

gcc: As seen in https://github.com/gcc-mirror/gcc/blob/master/gcc/config/i386/immintrin.h, gcc includes all intrinsic header files, but each individual file guards against the feature not being available.

This came out of an investigation in https://github.com/grpc/grpc/pull/27121.


derekmauro:
Thank you for the investigation and improvement!

PR #1021, opened 2021-09-21
Initialize support of AIX

miladfarca:
Google V8 has recently started using Abseil as one of its dependencies:
https://chromium-review.googlesource.com/c/v8/v8/+/2504257

We support V8 on IBM platforms which includes Big-Endian architectures (Linux on PowerPC and Z, AIX).
This is an initial PR to only fix compilation issues of abseil on V8 main branch on AIX.
Subsequent PRs will be  provided to add functionalities as well as fix abseil test failures.

### DIFF
--- a/absl/base/attributes.h
+++ b/absl/base/attributes.h
@@ -319,6 +319,10 @@
 // This functionality is supported by GNU linker.
 #ifndef ABSL_ATTRIBUTE_SECTION_VARIABLE
 #ifdef _AIX
+// __attribute__((section(#name))) on AIX is achived by using the `.csect` psudo
+// op which includes an additional integer as part of its syntax indcating
+// alignment. If data fall under different alignments then you might get a
+// compilation error indicating a `Section type conflict`.
 #define ABSL_ATTRIBUTE_SECTION_VARIABLE(name)
 #else
 #define ABSL_ATTRIBUTE_SECTION_VARIABLE(name) __attribute__((section(#name)))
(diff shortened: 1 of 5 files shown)


miladfarca:
I also wanted to clarify how changes are fetched into V8. It seems like chromium has its own repository for abseil?
https://chromium.googlesource.com/chromium/src/third_party/abseil-cpp/

Do changes first land on this github repository then get fetched into chromiums repo? and then subsequently into V8?
I.e. all changes need to be made in this repository and not chromiums?

derekmauro:
Hi and thanks for improving the portability of Abseil. I'll take a look at this soon and run in through our tests.

Changes to Abseil only need to happen here. I believe the Chromium periodically copies this repo into the Chromium repo.

miladfarca:
Hi Derek, thank you so much for clarifying.

miladfarca:
@derekmauro Hi Derek, I was wondering if we could prioritize code review of this pr as it will take a while until it's fetched all the way back into V8 main branch and with V8 release cycles we could get into backporting complications.

derekmauro:
I'll do my best, but please realize that this isn't my top priority, and merging this is more than just a matter of code review. With these low level changes, there is a cycle of tests that need to run to make sure the change doesn't affect less common platforms that are harder to test. These tests frequently take a day or more.

miladfarca:
Not a problem, thank you for clarifying.

PR #1022, opened 2021-09-25
Fixed typo `constuct` to `construct` in 3 places.

blais:
I keep noticing this. Fixing upstream.
(I left last Friday, so I can't fix internally.)


derekmauro:
Thanks!

PR #1028, opened 2021-09-29
Fix hashing on big endian platforms

miladfarca:
`std::hash` does not work correctly on big endian platforms using the following key types:
- std::hash<std::bitset>
- std::hash<std::vector<bool>>

This PR creates two AbslHashValue functions as a workaround to avoid using std::hash, 
more details on this bug can be seen here:
https://gcc.gnu.org/bugzilla/show_bug.cgi?id=102531
https://gcc.gnu.org/bugzilla/show_bug.cgi?id=98731

The kGolden values under `absl_low_level_hash_test` need to also be changed on BE platforms
as these load operations return different values depending on the machine endianness:
https://github.com/abseil/abseil-cpp/blob/master/absl/hash/internal/low_level_hash.cc#L10


absl/hash/internal/hash.h
```diff
@@ -500,6 +501,26 @@ AbslHashValue(H hash_state, const std::vector<T, Allocator>& vector) {
                     vector.size());
 }
 
+#if defined(ABSL_IS_BIG_ENDIAN)
```
derekmauro: This bug is specific to libstdc++'s implementation of std::hash, right? If so, can the comments say that this is a workaround for a libstdc++ bug and can the condition be limited to libstdc++?

absl/hash/internal/hash.h
```diff
@@ -500,6 +501,26 @@ AbslHashValue(H hash_state, const std::vector<T, Allocator>& vector) {
...
+// AbslHashValue for hashing std::vector<bool>
+//
+// std::hash does not work correctly with vector<bool> on Big Endian platforms
+// therefore we need to implement a custom AbslHashValue for it.
+// More details on the bug: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=102531
+template <typename H, typename T, typename Allocator>
+typename std::enable_if<is_hashable<T>::value && std::is_same<T, bool>::value,
+                        H>::type
+AbslHashValue(H hash_state, const std::vector<T, Allocator>& vector) {
+  std::vector<unsigned char> v;
```
derekmauro: We can call `reserve()` here.

absl/hash/internal/hash.h
```diff
@@ -592,9 +613,27 @@ AbslHashValue(H hash_state, const absl::variant<T...>& v) {
...
-// and a fallback to std::hash<> is most likely faster.
+// AbslHashValue for hashing std::bitset is not defined on Little Endian
+// platforms, for the same reason as for vector<bool> (see std::vector above):
+// It does not expose the raw bytes, and a fallback to std::hash<> is most
+// likely faster.
+
+#if defined(ABSL_IS_BIG_ENDIAN)
+// AbslHashValue for hashing std::bitset
+//
+// std::hash does not work correctly with std::bitset on Big Ednain platforms.
```
derekmauro: Ednian -> Endian

absl/hash/internal/hash.h
```diff
@@ -592,9 +613,27 @@ AbslHashValue(H hash_state, const absl::variant<T...>& v) {
...
+// See vector<bool> for more details.
+template <typename H, size_t N>
+H AbslHashValue(H hash_state, const std::bitset<N>& set) {
+  std::vector<unsigned char> v;
```
derekmauro: We can call `reserve()` here.

absl/hash/internal/low_level_hash_test.cc
```diff
@@ -404,6 +404,55 @@ TEST(LowLevelHashTest, VerifyGolden) {
...
+      0x76876721ff49f969, 0xbe7a796cfba15bf5, 0xa4c8bd54586f5488,
+      0xb390a325275501ab, 0x893f11317427ccf1, 0x92f2bb57da5695b9,
+      0x30985b90da88269f, 0x2c690e268e086de8, 0x1c02df6097997196,
+      0x1f9778f8bbdf6455, 0x7d57378c7bf8416d, 0xba8582a5f8d84d38,
+      0xe8ca43b85050be4e, 0x5048cf6bed8a5d9f, 0xfbc5ba80917d0ea4,
+      0x8011026525bf1691, 0x26b8dc6aed9fb50d, 0x191f5bfee77c1fe3,
+      0xdd497891465a2cc1, 0x6f1fe8c57a33072e, 0x2c9f4ec078c460c0,
+      0x9a725bde8f6a1437, 0x6ce545fa3ef61e4d,
+  };
+#else
```
derekmauro: If we make this `#elif defined(__aarch64__)` we can get rid of one of the `#endif`s at the end of the block.

absl/hash/internal/hash.h
```diff
@@ -592,9 +613,27 @@ AbslHashValue(H hash_state, const absl::variant<T...>& v) {
...
+// AbslHashValue for hashing std::bitset is not defined on Little Endian
+// platforms, for the same reason as for vector<bool> (see std::vector above):
+// It does not expose the raw bytes, and a fallback to std::hash<> is most
+// likely faster.
+
+#if defined(ABSL_IS_BIG_ENDIAN)
+// AbslHashValue for hashing std::bitset
+//
+// std::hash does not work correctly with std::bitset on Big Ednain platforms.
+// See vector<bool> for more details.
```
derekmauro: Please copy the details here.

derekmauro (changes requested):
That's quite a bug!

absl/hash/internal/hash.h
derekmauro: Actually, now that I think about it, we shouldn't use `std::vector` here because it since it allocates, it may throw.

I think you can probably use a stack array along with `PiecewiseCombiner` (see above or how it is used in `absl::Cord`) to hash this in pieces without allocating.

miladfarca: Thanks for information, I have refactored it to use PiecewiseCombiner. Not sure what you meant by using a stack array as vector size is not a constant and can't use VLAs in C++, I'm simply passing the address of the input 1 byte at a time, please let me know if this approach is acceptable.

absl/hash/internal/hash.h
miladfarca: Done.

absl/hash/internal/hash.h
miladfarca: I'm also using the `PiecewiseCombiner` approach here.

absl/hash/internal/low_level_hash_test.cc
miladfarca: Done.

absl/hash/internal/hash.h
miladfarca: Done.

absl/hash/internal/hash.h
```diff
@@ -489,8 +490,8 @@ typename std::enable_if<is_hashable<T>::value, H>::type AbslHashValue(
 
 // AbslHashValue for hashing std::vector
 //
-// Do not use this for vector<bool>. It does not have a .data(), and a fallback
-// for std::hash<> is most likely faster.
+// Do not use this for vector<bool> on Little Endian platforms. It does not have
```
derekmauro: "Little Endian platforms" -> "platforms that have a working implementation of std::hash"

absl/hash/internal/hash.h
```diff
@@ -500,6 +501,27 @@ AbslHashValue(H hash_state, const std::vector<T, Allocator>& vector) {
...
+// it. More details on the bug:
+// https://gcc.gnu.org/bugzilla/show_bug.cgi?id=102531
+template <typename H, typename T, typename Allocator>
+typename std::enable_if<is_hashable<T>::value && std::is_same<T, bool>::value,
+                        H>::type
+AbslHashValue(H hash_state, const std::vector<T, Allocator>& vector) {
+  typename H::AbslInternalPiecewiseCombiner combiner;
+  for (const auto& i : vector) {
+    unsigned char c = static_cast<unsigned char>(i);
+    hash_state = combiner.add_buffer(std::move(hash_state), &c, sizeof(c));
```
derekmauro, line 520: Sorry for not being more clear here. I wasn't suggesting using a variable sized array. I was suggesting repeatedly filling a smaller fixed-sized stack array so that we can call add_buffer fewer times. To be honest though, I'm not sure if that is any better than your approach.

derekmauro (changes requested):

absl/hash/internal/hash.h
miladfarca: Done.

absl/hash/internal/hash.h
miladfarca: Thank you for clarifying, that would be a good approach too yes although as you mentioned there will be a secondary loop to fill a buffer which may not make the overall operation much different. Please do let me know if you prefer it done that way and I will make the change.

LordBigSky (commented):
Anything else

PR #1031, opened 2021-10-05
Fix Randen and PCG on Big Endian platforms

miladfarca:
Randen expects seed and state buffers to be in LE ordering even on BE machines, and PCG is not, but they both use a shared seed implementation which is currently favouring Randen.

This patch fixes randen_slow to behave the same as  RandenHwAes and reverse the byte order when necessary and
changes seed generator to be in native byte order.


derekmauro:
I'd like a review from @laramiel on this one.

LordBigSky (approved):

LordBigSky:
Is well

miladfarca:
Hi @laramiel, Just wanted to follow up with you on reviewing this PR, thanks.

miladfarca:
I have also removed `little_endian::FromHost` from `absl/random/internal/explicit_seed_seq.h` as it was causing the following two test failures:
```
absl_random_internal_explicit_seed_seq_test
absl_random_internal_pcg_engine_test
```
The values didn't need to be byte swapped as the final returned values match LE ordering as is.

As a result some of the hex test results under `randen_engine_test.cc` also  had to be modified as they were using the same function.

laramiel:
A few things before I look at this:

RandenSlow and RandenHwAes Generate and Absorb methods *should* generate exactly identical transformations from the same byte-wise input. That should probably be better tested. Any variation in the byte-wise output between endian modes would be a bug.

It is desirable that randen_engine generates the same output across endian architectures.

The use of uint128* in RandenSlow is usually done merely to indicate that we're operating on 128-bit blocks; It could be (maybe should be) replaced with memcpy and simple increments.


What platform are you using? I don't have a BE machine handy, so I'd like to spin up a vm to verify.

miladfarca:
I'm using Linux on an S390x machine to run and verify the tests.

I don't think we are using RandenHwAes at the moment and only falling back to RandenSlow.
Using a memcpy will still be an issue if 128 bit values are written and 64 bit values are read back.

miladfarca:
To elaborate more on the issue I will use `RandenSlowTest` as an example, `absl::little_endian::FromHost64(elem)` isn't really needed as the actual element is on the opposite 64bit lane. It can be read in native byte order, only the index differs.

An alternative fix will be to also write all the values in LE order (which makes BE ops slower), or load the opposite array index when comparing the test result, or read __int128 values altogether instead of 64bits.

laramiel:
I think that there is an issue here, but it's not what you described.

`RandenSlow::Generate` has several code paths.

`prev_inner` reads a uint128 in platform native order, computes the xor, then stores, again in platform order. Converting this has no effect on the state of the buffer as the xor of swapped bytes is idempotent with the xor of non-swapped bytes.

The other codepath, though, may need to be investigated.

In `RandenHwAes::Generate` the PPC (and by extension, the S309x) codepath  potentially swaps the order of the `state` buffer before performing the `Permute` operation.  IIRC, the vector instructions of PPC always operated in big endian mode, where the rest of the cpu could be switched between big endian and little endian. 

```
  SwapEndian(state);
  Permute(state, keys);
  SwapEndian(state);
```

A similar change may be necessary for the S390x version of RandenSlow. 

What should be done to verify this? Start by changing RandenSlowTest and RandeHwAesTest to use an array of unsigned char[] for kGolden. They should be identical in each test and across architectures.

laramiel:
Looks like our comments crossed paths. The description of `RandenSlow::Generate` is what I'd expect on BE machines, and the fix should be to swap the state buffer on input and output, as you've described.

miladfarca:
Thank you for the suggestions, adding a `SwapEndian(state); ` under randen_slow can help pass the randen tests. I think the difference in vector/simd instructions on PPC could be related to lane/bit numbering on IBM machines instead of endianness. i.e lane/bit 0 is on the left side instead of right regardless of endianness.

The other concern is related to `seed_seq`. This snippet is shared between pcg_engine_test, explicit_seed_seq_test and randen_engine_test:
https://github.com/abseil/abseil-cpp/blob/master/absl/random/internal/explicit_seed_seq.h#L77

It works fine on `randen_engine_test` but fails the other two tests. removing `little_endian::FromHost32` from it passes the other two but then fails randen_engine_test. Is there a way to fix pcg_engine similar to randen_engine?

miladfarca:
I have pushed a new patch as you have suggested. To fix pcg tests I have removed `little_endian::FromHost3` from the shared seq generator and instead now reversing the byte order only under `randen_engine.h`.

laramiel:
That looks more like what I'd expect.

laramiel (commented):
I'd rather that we convert the kGolden to something like:

constexpr uint8_t kGolden[] = {
  0xee, 0xd3, ... 
};

And then compare the buffers directly using memcmp.

laramiel:
Also, if you add this test case to the bottom of explicit_seed_seq_test.cc, does it pass?

```
TEST(ExplicitSeedSeq, StdURBGGoldenTests) {

  // Verify that for std::- URBG instances the results are stable across platforms.
  {
    ExplicitSeedSeq seed_sequence{12, 34, 56};
    std::minstd_rand rng(seed_sequence);

    std::minstd_rand::result_type values[] = {rng(), rng(), rng(), rng()};
    EXPECT_THAT(values, testing::ElementsAre(579252, 43785881, 464353103, 1501811174 ));
  }

  {
    ExplicitSeedSeq seed_sequence{12, 34, 56};
    std::mt19937 rng(seed_sequence);

    std::mt19937::result_type values[] = {rng(), rng(), rng(), rng()};
    EXPECT_THAT(values, testing::ElementsAre(138416803, 151130212, 33817739, 138416803 ));
  }

  {
    ExplicitSeedSeq seed_sequence{12, 34, 56};
    std::mt19937_64 rng(seed_sequence);

    std::mt19937_64::result_type values[] = {rng(), rng(), rng(), rng()};
    EXPECT_THAT(values, testing::ElementsAre(19738651785169348, 1464811352364190456, 18054685302720800, 19738651785169348 ));
  }
}
```

miladfarca:
I'm not sure if using a char array would help the 2 failing tests as they are not just byte swapped, the entire returned value is different. For example here is the first two values from `TEST(PCG642018EngineTest, VerifyGoldenSeeded)` without my patch:
```
expected:
{0xb03988f1e39691ee, 0xbd2a1eb5ac31e97a, ...

engine():
{0x3be59729f5e0707a, 0x776f264a228d1471, ...
```
I have added your new test cases and they pass with my patch but fail without them, here is the result for the first one:
```
 Value of: values
 Expected: has 4 elements where
 element #0 is equal to 579252,
 element #1 is equal to 43785881,
 element #2 is equal to 464353103,
 element #3 is equal to 1501811174
   Actual: { 872419757, 419772477, 1329027822, 1723008931 } (of type unsigned long [4]), whose element #0 doesn't match
```

laramiel:
To verify: If we apply just the explicit_seed_seq.h changes then the tests in pcg_engine_test.cc and explicit_seed_seq_test.cc (the new ones) pass, correct?

If you change randen_slow_test.cc to this, does it pass?  (I will likely submit such a change)

```c++

TEST(RandenSlowTest, Default) {
  constexpr uint8_t kGolden[] = {
      0xee, 0xd3, 0xe6, 0x0e, 0x09, 0x34, 0x65, 0x6c, 0xc6, 0x33, 0x53, 0x9d,
      0x9b, 0x2b, 0x4e, 0x04, 0x77, 0x39, 0x43, 0x4e, 0x13, 0x4f, 0xc1, 0xc3,
      0xee, 0x10, 0x04, 0xd9, 0x7c, 0xf4, 0xa9, 0xdd, 0x10, 0xca, 0xd8, 0x7f,
      0x08, 0xf3, 0x7b, 0x88, 0x12, 0x29, 0xc7, 0x45, 0xf5, 0x80, 0xb7, 0xf0,
      0x9f, 0x59, 0x96, 0x76, 0xd3, 0xb1, 0xdb, 0x15, 0x59, 0x6d, 0x3c, 0xff,
      0xba, 0x63, 0xec, 0x30, 0xa6, 0x20, 0x7f, 0x6f, 0x60, 0x73, 0x9f, 0xb2,
      0x4c, 0xa5, 0x49, 0x6f, 0x31, 0x8a, 0x80, 0x02, 0x0e, 0xe5, 0xc8, 0xd5,
      0xf9, 0xea, 0x8f, 0x3b, 0x8a, 0xde, 0xd9, 0x3f, 0x5e, 0x60, 0xbf, 0x9c,
      0xbb, 0x3b, 0x18, 0x78, 0x1a, 0xae, 0x70, 0xc9, 0xd5, 0x1e, 0x30, 0x56,
      0xd3, 0xff, 0xb2, 0xd8, 0x37, 0x3c, 0xc7, 0x0f, 0xfe, 0x27, 0xb3, 0xf4,
      0x19, 0x9a, 0x8f, 0xeb, 0x76, 0x8d, 0xfd, 0xcd, 0x9d, 0x0c, 0x42, 0x91,
      0xeb, 0x06, 0xa5, 0xc3, 0x56, 0x95, 0xff, 0x3e, 0xdd, 0x05, 0xaf, 0xd5,
      0xa1, 0xc4, 0x83, 0x8f, 0xb7, 0x1b, 0xdb, 0x48, 0x8c, 0xfe, 0x6b, 0x0d,
      0x0e, 0x92, 0x23, 0x70, 0x42, 0x6d, 0x95, 0x34, 0x58, 0x57, 0xd3, 0x58,
      0x40, 0xb8, 0x87, 0x6b, 0xc2, 0xf4, 0x1e, 0xed, 0xf3, 0x2d, 0x0b, 0x3e,
      0xa2, 0x32, 0xef, 0x8e, 0xfc, 0x54, 0x11, 0x43, 0xf3, 0xab, 0x7c, 0x49,
      0x8b, 0x9a, 0x02, 0x70, 0x05, 0x37, 0x24, 0x4e, 0xea, 0xe5, 0x90, 0xf0,
      0x49, 0x57, 0x8b, 0xd8, 0x2f, 0x69, 0x70, 0xa9, 0x82, 0xa5, 0x51, 0xc6,
      0xf5, 0x42, 0x63, 0xbb, 0x2c, 0xec, 0xfc, 0x78, 0xdb, 0x55, 0x2f, 0x61,
      0x45, 0xb7, 0x3c, 0x46, 0xe3, 0xaf, 0x16, 0x18, 0xad, 0xe4, 0x2e, 0x35,
      0x7e, 0xda, 0x01, 0xc1, 0x74, 0xf3, 0x6f, 0x02, 0x51, 0xe8, 0x3d, 0x1c,
      0x82, 0xf0, 0x1e, 0x81,
  };

  alignas(16)   uint8_t state[RandenTraits::kStateBytes];
  std::memset(state, 0, sizeof(state));
  RandenSlow::Generate(RandenSlow::GetKeys(), state);
  EXPECT_EQ(0, std::memcmp(state, kGolden, sizeof(state)));
}
```

miladfarca:
That's right, applying only the explicit_seed_seq.h changes (and removing randen_engine.h changes) passes tests in `pcg_engine_test.cc` and `explicit_seed_seq_test.cc` but then fails the `Seeded` tests under randen_engine_test. The values are not just byte swapped anymore, they are completely different, so using a `uint8_t` may not help:
#### TEST(RandenTest, VerifyGoldenRanden64Seeded)
```
kGolden[kNumGoldenOutputs] = 
{ 0x83a9e58f94d3dcd5, 0x70bbdff3d97949fb, ... }

engine():
{ 0x2686906d366f5db0, 0x15ac16bae0bcba0 , ... }
```
`RandenSlowTest` is passing regardless as it's not seeded. This looks like a seeding issue as randen is using LE ordering even on BE machines, and pcg is not, but they both use a shared seed implementation which is currently favouring randen.

absl/random/internal/randen_engine.h
```diff
@@ -121,6 +121,12 @@ class alignas(16) randen_engine {
       const size_t requested_entropy = (entropy_size == 0) ? 8u : entropy_size;
       std::fill(std::begin(buffer) + requested_entropy, std::end(buffer), 0);
       seq.generate(std::begin(buffer), std::begin(buffer) + requested_entropy);
+#ifdef ABSL_IS_BIG_ENDIAN
+      // Reverse the seed byte order on Big Endian platforms.
+      for (int32_t i = 0; i < kBufferSize; ++i) {
+        buffer[i] = absl::little_endian::ToHost(buffer[i]);
```
laramiel: I think that this should use `FromHost`, and it's perhaps a bit more idiomatic to do this now:
```
// Randen expects the seed buffer to be in Little Endian; reverse it on Big Endian platforms.
for (sequence_result_type& e : buffer) {
  e = absl::little_endian::FromHost(e);
}
```

absl/random/internal/randen_test.cc
```diff
@@ -63,7 +64,7 @@ TEST(RandenTest, Default) {
 
   auto id = std::begin(state);
   for (const auto& elem : kGolden) {
-    EXPECT_EQ(elem, *id++);
+    EXPECT_EQ(absl::little_endian::FromHost64(elem), *id++);
```
laramiel: ... This test should use the same uint8_t constant as randen_slow_test and randen_hwaes_test. They are intended to be parallel, and I didn't convert this one.

laramiel:
Ok, yup. The randen_slow and explicit_seed_seq changes look good.

 ... it may also be useful to add a note to the comments that the seed and state buffers are expected to be in LE format.

absl/random/internal/randen_test.cc
miladfarca: I'm not sure which uint8_t constant you are referring to? randen_slow_test is also using a `constexpr uint64_t`?

absl/random/internal/randen_engine.h
miladfarca: Done.

absl/random/internal/randen_test.cc
laramiel: randen_test.cc  should be this, similar to the above. I have internal changes pending; I'm not sure when it gets pushed.  If you replace the test body with the following it will be acceptable.

```
TEST(RandenTest, Default) {
  constexpr uint8_t kGolden[] = {
      0xee, 0xd3, 0xe6, 0x0e, 0x09, 0x34, 0x65, 0x6c, 0xc6, 0x33, 0x53, 0x9d,
      0x9b, 0x2b, 0x4e, 0x04, 0x77, 0x39, 0x43, 0x4e, 0x13, 0x4f, 0xc1, 0xc3,
      0xee, 0x10, 0x04, 0xd9, 0x7c, 0xf4, 0xa9, 0xdd, 0x10, 0xca, 0xd8, 0x7f,
      0x08, 0xf3, 0x7b, 0x88, 0x12, 0x29, 0xc7, 0x45, 0xf5, 0x80, 0xb7, 0xf0,
      0x9f, 0x59, 0x96, 0x76, 0xd3, 0xb1, 0xdb, 0x15, 0x59, 0x6d, 0x3c, 0xff,
      0xba, 0x63, 0xec, 0x30, 0xa6, 0x20, 0x7f, 0x6f, 0x60, 0x73, 0x9f, 0xb2,
      0x4c, 0xa5, 0x49, 0x6f, 0x31, 0x8a, 0x80, 0x02, 0x0e, 0xe5, 0xc8, 0xd5,
      0xf9, 0xea, 0x8f, 0x3b, 0x8a, 0xde, 0xd9, 0x3f, 0x5e, 0x60, 0xbf, 0x9c,
      0xbb, 0x3b, 0x18, 0x78, 0x1a, 0xae, 0x70, 0xc9, 0xd5, 0x1e, 0x30, 0x56,
      0xd3, 0xff, 0xb2, 0xd8, 0x37, 0x3c, 0xc7, 0x0f, 0xfe, 0x27, 0xb3, 0xf4,
      0x19, 0x9a, 0x8f, 0xeb, 0x76, 0x8d, 0xfd, 0xcd, 0x9d, 0x0c, 0x42, 0x91,
      0xeb, 0x06, 0xa5, 0xc3, 0x56, 0x95, 0xff, 0x3e, 0xdd, 0x05, 0xaf, 0xd5,
      0xa1, 0xc4, 0x83, 0x8f, 0xb7, 0x1b, 0xdb, 0x48, 0x8c, 0xfe, 0x6b, 0x0d,
      0x0e, 0x92, 0x23, 0x70, 0x42, 0x6d, 0x95, 0x34, 0x58, 0x57, 0xd3, 0x58,
      0x40, 0xb8, 0x87, 0x6b, 0xc2, 0xf4, 0x1e, 0xed, 0xf3, 0x2d, 0x0b, 0x3e,
      0xa2, 0x32, 0xef, 0x8e, 0xfc, 0x54, 0x11, 0x43, 0xf3, 0xab, 0x7c, 0x49,
      0x8b, 0x9a, 0x02, 0x70, 0x05, 0x37, 0x24, 0x4e, 0xea, 0xe5, 0x90, 0xf0,
      0x49, 0x57, 0x8b, 0xd8, 0x2f, 0x69, 0x70, 0xa9, 0x82, 0xa5, 0x51, 0xc6,
      0xf5, 0x42, 0x63, 0xbb, 0x2c, 0xec, 0xfc, 0x78, 0xdb, 0x55, 0x2f, 0x61,
      0x45, 0xb7, 0x3c, 0x46, 0xe3, 0xaf, 0x16, 0x18, 0xad, 0xe4, 0x2e, 0x35,
      0x7e, 0xda, 0x01, 0xc1, 0x74, 0xf3, 0x6f, 0x02, 0x51, 0xe8, 0x3d, 0x1c,
      0x82, 0xf0, 0x1e, 0x81,
  };

  alignas(16) uint8_t state[Randen::kStateBytes];
  std::memset(state, 0, sizeof(state));

  Randen r;
  r.Generate(state);

  EXPECT_EQ(0, std::memcmp(state, kGolden, sizeof(state)));
}
```

miladfarca: Thank you for clarifying, I've made the change as you have suggested.

laramiel (approved):
This version looks good.

miladfarca:
@laramiel Thank you for your time.

absl/random/internal/randen_slow.cc
```diff
@@ -395,6 +395,20 @@ inline ABSL_RANDOM_INTERNAL_ATTRIBUTE_ALWAYS_INLINE void Permute(
   }
 }
 
+// Enables native loads in the round loop by pre-swapping.
+inline ABSL_RANDOM_INTERNAL_ATTRIBUTE_ALWAYS_INLINE void SwapEndian(
+    absl::uint128* state) {
```
derekmauro, line 400:
```
absl/random/internal/randen_slow.cc:400:20: error: unused parameter 'state' [-Werror,-Wunused-parameter]
    absl::uint128* state) {
```

A rather annoying warning. I think instead omit the variable name, maybe something like
```
#ifdef ABSL_IS_BIG_ENDIAN
inline ABSL_RANDOM_INTERNAL_ATTRIBUTE_ALWAYS_INLINE void SwapEndian(absl::uint128* state) {
...
}
#else
inline ABSL_RANDOM_INTERNAL_ATTRIBUTE_ALWAYS_INLINE void SwapEndian(absl::uint128*) {}
#endif
```

absl/random/internal/randen_test.cc
```diff
@@ -41,30 +42,38 @@ TEST(RandenTest, CopyAndMove) {
...
-      0x15dbb1d37696599f, 0x30ec63baff3c6d59, 0xb29f73606f7f20a6,
-      0x02808a316f49a54c, 0x3b8feaf9d5c8e50e, 0x9cbf605e3fd9de8a,
-      0xc970ae1a78183bbb, 0xd8b2ffd356301ed5, 0xf4b327fe0fc73c37,
-      0xcdfd8d76eb8f9a19, 0xc3a506eb91420c9d, 0xd5af05dd3eff9556,
-      0x48db1bb78f83c4a1, 0x7023920e0d6bfe8c, 0x58d3575834956d42,
-      0xed1ef4c26b87b840, 0x8eef32a23e0b2df3, 0x497cabf3431154fc,
-      0x4e24370570029a8b, 0xd88b5749f090e5ea, 0xc651a582a970692f,
-      0x78fcec2cbb6342f5, 0x463cb745612f55db, 0x352ee4ad1816afe3,
-      0x026ff374c101da7e, 0x811ef0821c3de851,
+  constexpr uint8_t kGolden[] = {
```
derekmauro: If you rebase on HEAD you should be able to revert this file.

derekmauro (changes requested):

absl/random/internal/randen_slow.cc
miladfarca: I've added a `(void)state;` on LE. Please let me know if you prefer removing the variable name instead.

absl/random/internal/randen_test.cc
miladfarca: HEAD doesn't seem to have this change: https://github.com/abseil/abseil-cpp/blob/HEAD/absl/random/internal/randen_test.cc#L43
Would you want me to remove the change?

derekmauro: I just realized it is in our internal repo but hasn't made it to GitHub yet. If you remove it, I'll merge this after it makes it to GitHub.

miladfarca: Not a problem, I have removed the randen_test changes.

derekmauro:
Thanks for working with us on this one.

miladfarca:
Not a problem, thank you for your help.

Issue #1024, opened 2021-09-27
[FreeBSD+Bazel] absl/debugging/internal/elf_mem_image.h:66:11: error: unknown type name 'ElfW'


yesudeep:
**Describe the bug**

```
ERROR:
...
  /usr/bin/clang -U_FORTIFY_SOURCE '-D_FORTIFY_SOURCE=1' -fstack-protector -Wall -fno-omit-frame-pointer '-std=c++0x' -MD -MF bazel-out/freebsd-fastbuild/bin/external/com_google_absl/absl/debugging/_objs/debugging_internal/elf_mem_image.pic.d '-frandom-seed=bazel-out/freebsd-fastbuild/bin/external/com_google_absl/absl/debugging/_objs/debugging_internal/elf_mem_image.pic.o' -fPIC -iquote external/com_google_absl -iquote bazel-out/freebsd-fastbuild/bin/external/com_google_absl '-DHAVE_BAZEL_BUILD=1' '-std=c++2a' -Wall -Wreturn-type -Wuninitialized -Wunused-result '-Werror=narrowing' '-Werror=reorder' -Wunused-local-typedefs '-Werror=conversion-null' '-Werror=overlength-strings' '-Werror=pointer-arith' '-Werror=varargs' '-Werror=vla' '-Werror=write-strings' -Wmissing-declarations '-fdiagnostics-color=always' -Wall -Wextra -Wcast-qual -Wconversion-null -Wformat-security -Wmissing-declarations -Woverlength-strings -Wpointer-arith -Wundef -Wunused-local-typedefs -Wunused-result -Wvarargs -Wvla -Wwrite-strings -DNOMINMAX -no-canonical-prefixes -Wno-builtin-macro-redefined '-D__DATE__="redacted"' '-D__TIMESTAMP__="redacted"' '-D__TIME__="redacted"' -c external/com_google_absl/absl/debugging/internal/elf_mem_image.cc -o bazel-out/freebsd-fastbuild/bin/external/com_google_absl/absl/debugging/_objs/debugging_internal/elf_mem_image.pic.o)
Execution platform: @local_config_platform//:host
In file included from external/com_google_absl/absl/debugging/internal/elf_mem_image.cc:18:
external/com_google_absl/absl/debugging/internal/elf_mem_image.h:66:11: error: unknown type name 'ElfW'
    const ElfW(Sym) *symbol;    // Symbol in the dynamic symbol table.
          ^
[...]

Issue #819, opened 2020-10-15
Missing -std=c++17 causes dependent project failures


yurivict:
For example, Google's ```or-tools``` fails to compile when ```abseil``` is built without ```-std=c++17```.

See https://github.com/google/or-tools/issues/2196

derekmauro:
It looks like you've solved the problem. See https://github.com/abseil/abseil-cpp/blob/master/FAQ.md#how-to-i-set-the-c-dialect-used-to-build-abseil.

PR #1037, opened 2021-10-07
Use freebsd macro definition for ElfW macro for compatibility.

yesudeep:
Addresses: https://github.com/abseil/abseil-cpp/issues/1024

The `ElfW()` macro was added to `elf_generic.h` recently and as such may be absent most current releases of FreeBSD.

See: https://github.com/freebsd/freebsd-src/blob/34e7e4b6a059eee5e4e3e34de5b9d5f0d6e589f9/sys/sys/elf_generic.h#L60

There are other build issues on FreeBSD. I'll address them soonish in other PRs.

@derekmauro @EmployedRussian

### DIFF
--- a/absl/debugging/internal/vdso_support.cc
+++ b/absl/debugging/internal/vdso_support.cc
@@ -50,6 +50,11 @@
 #define AT_SYSINFO_EHDR 33  // for crosstoolv10
 #endif
 
+#if defined(__FreeBSD__)
+using Elf64_aux_t = Elf64_Auxinfo;
+using Elf32_aux_t = Elf32_Auxinfo;
+#endif
+
 namespace absl {
 ABSL_NAMESPACE_BEGIN
 namespace debugging_internal {
--- a/absl/debugging/internal/vdso_support.cc
+++ b/absl/debugging/internal/vdso_support.cc
@@ -51,8 +51,8 @@
 #endif
 
 #if defined(__FreeBSD__)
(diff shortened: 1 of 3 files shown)


yesudeep:
Sent a new PR with clang-autoformatting disabled in the editor.

absl/debugging/internal/vdso_support.cc
```diff
@@ -50,6 +50,11 @@
 #define AT_SYSINFO_EHDR 33  // for crosstoolv10
 #endif
 
+#if defined(__FreeBSD__)
+using Elf64_aux_t = Elf64_Auxinfo;
```
EmployedRussian: s/aux/auxv/

EmployedRussian (changes requested):
This looks good to me, except the `aux` vs. `auxv` typo is back.

absl/debugging/internal/vdso_support.cc
yesudeep: I should have drank the coffee lol.

derekmauro (approved):

danfe:
> There are other build issues on FreeBSD. I'll address them soonish in other PRs.

What are those issues?  We do need to patch the `absl/base/internal/unscaledcycleclock.cc` file in FreeBSD port, but otherwise it all seems file so far.  We also bump the C++ standard because of #819.

PR #1038, opened 2021-10-10
Remove bazelbuild/rules_cc dependency

Vertexwahn:
bazelbuild/rules_cc was put on hold
It seems the recommended way is now to use native cc rules

More info [here](https://github.com/bazelbuild/rules_go/pull/2950)


derekmauro:
Thanks!

Issue #1044, opened 2021-10-19
Over-aligned type causes test failures


miladfarca:
This test case can causes compilation warnings depending on the compiler/platform:
https://github.com/abseil/abseil-cpp/blob/master/absl/container/internal/layout_test.cc#L1350
```
 warning: requested alignment 16 is larger than 8
```
`alignas(2 * M) ` is creating an over-aligned type which makes the behaviour implementation defined and potentially unsupported on some platforms.

Currently it is causing a test failure on PPC and S390 due to this line:
https://github.com/abseil/abseil-cpp/blob/master/absl/container/internal/layout.h#L483
The value is byte aligned whereas `alignment` is equal to `16`.

I was wondering if this test case can be refactored, guarded to be platform specific or removed to fix the above two issues?

miladfarca:
@derekmauro Hi Derek, would you have a suggestion on the best way we can fix this issue?

derekmauro:
That's a good question. Off the top of my head I'm not sure if Abseil uses any over-aligned types in code that isn't platform specific. I do think we want to ensure that `Layout` works with over-aligned types when supported by the platform. It might be sufficient to create a `#define` for platforms that support over-alignment and use it in some kind of `static_assert` about the alignment in `Layout`. But I'll ask around and see if anyone has a better idea.

jyknight:
Looks like this was a GCC bug, fixed in GCC 10 and the 9.3 patch release, see https://gcc.gnu.org/PR89357. A workaround would be to use `__attribute__((aligned(N)))` for those older versions, rather than `alignas(N)`.

miladfarca:
Thank you for the suggestion, this does fix the warning and passes the test, however I'm wonder if this falls under undefined behaviour as it's greater than `std::max_align_t`?

jyknight:
It is implementation-defined whether "extended alignment" (alignments greater than max_align_t) are supported -- and a program which requests an extended alignment which the implementation does not support is deemed to be ill-formed (which means a diagnostic is required). This is not "undefined behavior", but is potentially non-portable.

In practice, compiler support for over-aligned types is pretty much universal.

The bug in GCC was that it was trying to implement said standards requirement, but did so incorrectly. GCC did in fact support over-aligned types all along, yet was emitting a diagnostic (and ignoring the alignas) as if it did not.

miladfarca:
Thank you for the information, I have created this PR: https://github.com/abseil/abseil-cpp/pull/1049

[...]

PR #1049, opened 2021-10-25
Fix over-aligned layout test with older gcc compilers

miladfarca:
Over aligning with `alignas` may not be supported on older gcc compilers due to this bug:
https://gcc.gnu.org/bugzilla/show_bug.cgi?id=89357

Using `__attribute__((aligned()))` instead to fix warnings as well as pass the test.
More information can be found on this issue: https://github.com/abseil/abseil-cpp/issues/1044

### DIFF
--- a/absl/container/internal/layout_test.cc
+++ b/absl/container/internal/layout_test.cc
@@ -1350,7 +1350,9 @@ TEST(Layout, CustomAlignment) {
 TEST(Layout, OverAligned) {
   constexpr size_t M = alignof(max_align_t);
   constexpr Layout<unsigned char, Aligned<unsigned char, 2 * M>> x(1, 3);
-  alignas(2 * M) unsigned char p[x.AllocSize()];
+  // Using __attribute__ ((aligned ())) instead of alignas to bypass a gcc bug:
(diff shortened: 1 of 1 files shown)


absl/container/internal/layout_test.cc
```diff
@@ -1350,7 +1350,9 @@ TEST(Layout, CustomAlignment) {
 TEST(Layout, OverAligned) {
   constexpr size_t M = alignof(max_align_t);
   constexpr Layout<unsigned char, Aligned<unsigned char, 2 * M>> x(1, 3);
-  alignas(2 * M) unsigned char p[x.AllocSize()];
+  // Using __attribute__ ((aligned ())) instead of alignas to bypass a gcc bug:
+  // https://gcc.gnu.org/bugzilla/show_bug.cgi?id=89357
+  __attribute__((aligned(2 * M))) unsigned char p[x.AllocSize()];
```
derekmauro, line 1356: This only works with GCC. I think you will need `#ifdef __GNUC__`, otherwise use `alignas`.

miladfarca: Thank you, I have now added the gcc guard.

PR #1078, opened 2021-12-09
add coverage compile option

littleDrew:
add coverage compile option for better code testing rate detect

### DIFF
--- a/CMake/AbseilHelpers.cmake
+++ b/CMake/AbseilHelpers.cmake
@@ -407,10 +407,18 @@ function(absl_cc_test)
     PRIVATE ${ABSL_CC_TEST_COPTS}
   )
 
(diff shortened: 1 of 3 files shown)


derekmauro:
In general our policy is to limit the number of build flags. Each build flag creates a combinatoric explosion of possible build modes to test and maintain. In this specific case, you shouldn't need a flag at all. You should just be able to do something like like:

`cmake -DABSL_BUILD_TESTING=ON -DABSL_USE_GOOGLETEST_HEAD=ON -DCMAKE_CXX_FLAGS="-O0 -fprofile-arcs -ftest-coverage" ..`

In regards to Abseil's test coverage, thanks for reminding me that we still have some tests that we run internally that we haven't open-sourced.

PR #1083, opened 2021-12-29
cmake: use target aliases from local Google Test checkout.

anpol:
`ABSL_FIND_GOOGLETEST` is a dependent option, it is always `OFF` when
`ABSL_USE_EXTERNAL_GOOGLETEST` is `OFF` (i.e. when Google Test is
checked out locally).

`GTest::` target aliases are to be added only if `find_package(GTest)`
has been actually used.


CMakeLists.txt
```diff
@@ -157,7 +157,7 @@ if(BUILD_TESTING AND ABSL_BUILD_TESTING)
     include(CMake/Googletest/DownloadGTest.cmake)
   endif()
 
-  if (NOT ABSL_FIND_GOOGLETEST)
+  if (ABSL_USE_EXTERNAL_GOOGLETEST AND NOT ABSL_FIND_GOOGLETEST)
```
derekmauro: Maybe I'm misunderstanding this since we rarely use CMake ourselves, but could/should this logic be merged into the logic on lines 135-141?

anpol: Now when you asked, it's easier to justify a larger change.

I'm pretty sure the aliases could be moved into `NOT ABSL_FIND_GOOGLETEST` section, but you might need to ask someone else to confirm, as the change becomes non-trivial.

To clarify:

* those aliases are not needed for `ABSL_USE_EXTERNAL_GOOGLETEST` case anymore, since google/googletest#3155 has been merged.
* the actual problem is `NOT ABSL_USE_EXTERNAL_GOOGLETEST` case; in that case `GTest::gtest` target or alias already exists.

derekmauro: Thanks. Looks good, should be merged soon.

derekmauro (approved):

Issue #1093, opened 2022-01-19
The Mix function fallback for arm64 is too weak in diffusion


renzibei:
**Describe the bug**

The [Mix function](https://github.com/abseil/abseil-cpp/blob/c59e7e59f5d29619ddc07fcb59be3dcba9585814/absl/hash/internal/hash.h#L915-L932) uses an alternative solution on arm64 architecture, with a note in the comments that uint128_t multiplication is too slow for arm64 cpus because two instructions are required.

However, the alternative provided as a function of mixing bits does a poor job of diffusion.

Below is what is does on arm64 platform(some comments were omitted):

```c++
ABSL_ATTRIBUTE_ALWAYS_INLINE static uint64_t Mix(uint64_t state, uint64_t v) {
    using MultType = uint64_t;
    MultType m = state + v;
    m *= kMul;
    return static_cast<uint64_t>(m ^ (m >> (sizeof(m) * 8 / 2)));
  }
```

Basically, there are three steps,
1. Add `state` and `v` with `m = state + v`. 
2. Multiply m by a constant `m *= kMul`. 
3. XOR the upper 32 bits and the lower 32 bits of m as the new lower 32 bits `m ^= (m >> 32)`. And return m as the Mix value. 

If we specify that the highest bit of a uint64_t type is the 64th bit and the lowest bit is the 1st bit. The problem with this function is that the i-th bit of state and the i-th bit of v (i is greater than 33) do not affect the final bits from i - 1 to 33 and from i - 33 to 1. For example, the value of 64-th bit of state will only affect bits 64 and 32 of the return value. This makes Mix's diffusion as an important part of the hash function (and the only part of integer hashing) very weak.

In particular, if only the high-order bits of the input keys are different, and the low-order bits are all the same, then using it as a hash function will have a lot of collisions. And this input feature is actually very common. Examples are pointers and 64-bit integers made with shorter keys.

The results in [this benchmark](https://github.com/renzibei/hashtable-bench#kv-uint64_t-with-bits-in-low-position-masked-out-uint64_t) further confirmed this. The following figure is the lookup test results from that benchmark. Please focus on the performance of `absl::flat_hash_map` with `absl::Hash`.

![mask_low_bits_uint64_t,uint64_t ,avg_hit_find_with_rehash](https://user-images.githubusercontent.com/31380016/150161698-62b5c456-7806-458d-bc8a-6221dde879e1.png)

**Steps to reproduce the bug**

[...]

PR #1094, opened 2022-01-19
Fix the weak diffusion Mix implementation on arm64

renzibei:
## Related problem
The problem this PR tries to fix is described in https://github.com/abseil/abseil-cpp/issues/1093.  

In short, the diffusion feature of the original Mix implementation on the arm64 architecture is too weak to be effective. Causes the `absl::Hash` for integer types based on it to have a lot of collisions when dealing with keys that differ only in the high-order bits.

## PR Purpose

We hope that the repaired Mix function can achieve the goals on the arm64 platform: 
1. It has a good diffusion effect and reduces conflicts of `absl::Hash<uint64_t>` for most common data.
2. It has a good performance.

## Methodology

The implementation of the Mix function on other architectures is to use 64x64->128 multiplication and then XOR the high-order 64bit and the low-order 64bit. This method has a good diffusion effect and is a good Mix function. Plus it only requires a mul instruction on x86-64 platforms. The quality and efficiency of this method are both good.

On the arm64 platform, 64x64->128 multiplication requires two multiplication instructions, but does this mean that the required latency is too high and the achievable throughput is too low? More importantly, are there any better alternatives to achieve acceptable obfuscation quality?

## Test

We tested the Mix function using 128-bit multiply on both Apple M1 Max and raspberry pi 4(ARM Cortex-A72). Probably we should all roughly agree that the former is one of the best performing arm64 processors out there, while the latter is one of the weakest.

The content of the test is to test the speed of the successful look of `absl::flat_hash_map` with different hash functions on datasets of uint64_t keys with different characteristics. If the lookup time is relatively long, either the hash function itself is time-consuming, or the hash function causes too many collisions.

The test codes and data are from one [hash table benchmark](https://github.com/renzibei/hashtable-bench).
In addition to the original Mix-based `absl::hash` and our repaired version, a pure 128-bit multiplication hash function and `robin_hood::hash` are added for comparison. [robin_hood::hash](https://github.com/martinus/robin-hood-hashing/blob/9145f963d80d6a02f0f96a47758050a89184a3ed/src/include/robin_hood.h#L748-L759) uses a combination of right-shift XOR, multiplication, and right-shift XOR.

Here come the results:

mask_low_bits_uint64_t,uint64_t ,avg_hit_find_without_rehash
Figure 1: Look up uint64_t keys with low-bits masked out, tested on raspberry pi 4

mask_low_bits_uint64_t,uint64_t ,avg_hit_find_without_rehash
Figure 2: Look up uint64_t keys with low-bits masked out, tested on Apple M1 Max

mask_split_bits_uint64_t,uint64_t ,avg_hit_find_without_rehash
Figure 3: Look up uint64_t keys with several split bits masked out, tested on raspberry pi 4

mask_split_bits_uint64_t,uint64_t ,avg_hit_find_without_rehash
Figure 4: Look up uint64_t keys with several split bits masked out, tested on Apple M1 Max

uniform_uint64_t,uint64_t ,avg_hit_find_without_rehash
Figure 5: Look up uint64_t keys uniformly distributed, tested on raspberry pi 4

uniform_uint64_t,uint64_t ,avg_hit_find_without_rehash
Figure 6: Look up uint64_t keys uniformly distributed, tested on Apple M1 Max

What can we learn from the test results? Let's first look at the results on the worse performing Raspberry Pi platform, i.e. Figure 1, Figure 3, Figure 5. This way we can better know what our worst performance is.

From figure 1, for keys with low-order bits masked out, the fixed version of Mix can perform much better than before. Although it is slower than `robin_hood::hash` on some data scales, it is more stable for different data scales because its hash quality is higher than that of `robin_hood::hash`.

From figure 3, for keys with some bits in the middle covered, we can better see the stability of the confusion quality of the repaired Mix function for keys with different characteristics.

From figure 5, for keys that are evenly distributed in the uint64_t range, even the simplest equivalent hash can have very few collisions. The speed on this data roughly describes the speed of computing the hash function itself. It can be seen that although the calculation speed of the repaired absl hash function is slower than that of the `robin_hood::hash`, it is only slightly slower than the pre-fixed version, and almost no difference when the number of elements is large. Therefore, it can be said that there is at least no loss in performance compared to before, and the hash quality is also enhanced.

Figure 2, Figure 4, and Figure 6 show the test results on the M1 chip. First of all, it can be seen that on the most modern arm64 chips, the performance is strong enough that two multiply instructions are not slower than the `robin_hood::hash` of one multiply and 2 pairs of shift+XOR instructions. In addition Figure 4 shows that the repaired absl hash has a much higher hash quality than the hash `robin_hood::hash`.

All in all, I think this PR achieves its goal: to make the Mix function have good diffusion properties on the arm64 platform while also having decent performance.


atdt:
@renzibei, thank you for the detailed analysis. We observed poor performance when calculating a 128-bit product on the Arm Neoverse microarchitecture, which is the most popular one for cloud workloads. On the Neoverse N1, `MUL` and `MULH` have a latency of 4 and 5 cycles respectively. Additionally, `MULH` stalls the multiplier pipeline for 3 cycles (see [Arm Neoverse N1 Software Optimization Guide](https://documentation-service.arm.com/static/5f05e93dcafe527e86f61acd?token=), page 17). As a result, bursts of multiples can be very slow. It does not look like Apple's M1 chips have this issue.

I'd like to find a way to fix the diffusion issue you pointed out without re-introducing the multiplies. I think it can be done by replacing one of the xor operations with a subtraction, but I need to test this to make sure.

renzibei:
@atdt you mentioned that the `MULH` on Neoverse N1 has an exec latency of 5 cycles and stalls the multiplier pipeline for 3 cycles. According to [Cortex-A72 Software Optimization Guide](https://developer.arm.com/documentation/uan0016/a/), the `UMULH` in Cortex-A72 has an exec latency of 6 cycles and stalls the multiplier pipeline for 3 cycles. In other words, the `MULH` instruction should be more expensive on Cortex-A72 than on Neoverse N1.

The raspberry pi 4 I tested used a Cortex-A72 SoC. From the test results of "Figure 5: Look up uint64_t keys uniformly distributed, tested on raspberry pi 4", when the data can fit in the L1 cache, the 64x64->128 version `fix_absl::Hash` is about 2.4 ns slower than `absl::Hash` for a successful lookup. This is close to the theory cost of one `UMUL` instruction. When the scale of data is large, the difference in calculation hash value can be omitted.

I don't know is this extra cost in hash noticeable for google arm-based cloud tasks. However, for non-uniformly distributed data, current implementations tend to lead to worse performance in hash tables.

Hopefully we can find a solution that balances efficiency and hash quality. If not, try to find an acceptable solution.

atdt:
I think using a CRC instruction for mixing might be an acceptable alternative; I'm working on a PR.
Happy Lunar New Year, by the way :)

renzibei:
> I think using a CRC instruction for mixing might be an acceptable alternative; I'm working on a PR.
> Happy Lunar New Year, by the way :)

Hi, it has been some time since last discussion. Have you got any progress in the direction of utilizing CRC? 

Since CRC32 instruction is a 32-bit instruction, I think it might be some problems that need to be solved when the target is a 64bit hash value?

atdt:
Sorry for the delay, I was held up by other commitments. I want to continue looking at this but I will have an answer one way or another by this weekend.

atdt:
I think it's fine to land this now while we continue looking for a suitable alternative.

renzibei:
> I think it's fine to land this now while we continue looking for a suitable alternative.

Ok. What do I need to get this merged?

By the way, for the direction using CRC32 instruction, I think there will be problems when mixing the outputs of two CRC32 instructions. When the target is a 64-bit hash value and the two existing hash values are 32-bit, and we need to maintain good obfuscation and diffusion properties, then the mixing operation will probably require a fair amount of arithmetic instructions.

derekmauro:
> Ok. What do I need to get this merged?

I will get this merged. Please be patient.

PR #1095, opened 2022-01-19
[Fuchsia] Reduce number of threads for LoadZonesConcurrently test.

Colemalban:
This reduces the thread count from 128 -> 64 when running on Fuchsia.
With the higher thread count, the test was flaky and would often time
out.

To test, I ran this 100 times to verify that it passes each time.


derekmauro:
Please send your pull-request to https://github.com/google/cctz. We regularly import the source from there.

PR #1097, opened 2022-01-21
Avoid defining ABSL_INTERNAL_ENABLE_FORMAT_CHECKER when compiling with vscode/intellisense

romange:
This macro is defined in such way that it can be overriden with the compiler -D option.
However, `#ifdef` guards only check for the existence of this macro, thus it's defacto impossible to override.
This PR fixes this issue. Needed to workaround vscode/clang spurious error https://github.com/microsoft/vscode-cpptools/issues/3683

### DIFF
--- a/absl/strings/internal/str_format/checker.h
+++ b/absl/strings/internal/str_format/checker.h
@@ -22,9 +22,15 @@
 // Compile time check support for entry points.
 
 #ifndef ABSL_INTERNAL_ENABLE_FORMAT_CHECKER
-#if ABSL_HAVE_ATTRIBUTE(enable_if) && !defined(__native_client__)
+
+// We disable format checker under vscode intellisense compilation.
+// See https://github.com/microsoft/vscode-cpptools/issues/3683 for
+// more details.
+#if ABSL_HAVE_ATTRIBUTE(enable_if) && !defined(__native_client__) && \
+    !defined(__INTELLISENSE__)
 #define ABSL_INTERNAL_ENABLE_FORMAT_CHECKER 1
-#endif  // ABSL_HAVE_ATTRIBUTE(enable_if) && !defined(__native_client__)
+#endif  // ABSL_HAVE_ATTRIBUTE(enable_if) && !defined(__native_client__) &&
+        // !defined(__INTELLISENSE__)
 #endif  // ABSL_INTERNAL_ENABLE_FORMAT_CHECKER
 
 namespace absl {


derekmauro (changes requested):
No symbol containing the string INTERNAL is meant to be spelled by Abseil users, ever. If there is a bug, we should change the conditions under which it is defined.

romange:
@derekmauro  I found out how to recognize intellisense compilation. Indeed, now the change is more straightforward. PTAL.

absl/strings/internal/str_format/checker.h
```diff
@@ -22,7 +22,7 @@
 // Compile time check support for entry points.
 
 #ifndef ABSL_INTERNAL_ENABLE_FORMAT_CHECKER
-#if ABSL_HAVE_ATTRIBUTE(enable_if) && !defined(__native_client__)
+#if ABSL_HAVE_ATTRIBUTE(enable_if) && !defined(__native_client__) && !defined(__INTELLISENSE__)
```
derekmauro: https://google.github.io/styleguide/cppguide.html#Line_Length - Please limit lines to 80 chars.

derekmauro: Please leave a comment or a link to a bug so feature readers can figure out why we disabled this under intellisense.

absl/strings/internal/str_format/checker.h
```diff
@@ -22,7 +22,7 @@
...
 #define ABSL_INTERNAL_ENABLE_FORMAT_CHECKER 1
 #endif  // ABSL_HAVE_ATTRIBUTE(enable_if) && !defined(__native_client__)
```
derekmauro: Fix this comment since it now doesn't match.

derekmauro (changes requested):

absl/strings/internal/str_format/checker.h
romange: the new comment will cross 80-char limit. How do you suggest to format it @derekmauro ?

absl/strings/internal/str_format/checker.h
romange: fixed

romange: fixed

absl/strings/internal/str_format/checker.h
derekmauro: The way you wrote it is fine.

derekmauro (approved):

derekmauro:
I'll merge this soon.

PR #1110, opened 2022-02-21
Fit for build on loong64 architecture

imaiguo:
build abseil success on loong64, CPU Loongson-3A5000LL,and run ctest ok.

code with reference  to url:
https://github.com/loongson/glibc/blob/la64/main/devel/sysdeps/unix/sysv/linux/loongarch/sys/ucontext.h


imaiguo:
abseil used in chromium 93 and later version,we want to run chromium on loongarch64 platform.

derekmauro:
This will be merged soon.

PR #1113, opened 2022-03-01
Some trivial OpenBSD-related fixes

noiseless:
This is not an exhaustive set of fixes required for abseil to work correctly on OpenBSD, but I suppose it would be a good idea to start with these diffs.

Comments, suggestions and objections are welcome. If you don't have any objections, I'll be happy to see this merged into the main branch. Thanks in advance.


noiseless:
Thanks!

PR #1134, opened 2022-03-16
Exclude #pragma intrinsic(_umul128) from ARM64EC

mcfi:
ARM64EC is a Microsoft-designed ARM64 ABI compatible with AMD64 applications on ARM64 Windows 11.  The ARM64EC does not support _umul128 as an intrinsic, but an inline function. Therefore, exclude #pragma intrinsic(_umul128) from ARM64EC compilation.

### DIFF
--- a/absl/base/internal/unscaledcycleclock.cc
+++ b/absl/base/internal/unscaledcycleclock.cc
@@ -131,7 +131,9 @@ double UnscaledCycleClock::Frequency() {
 
 #elif defined(_M_IX86) || defined(_M_X64)
 
+#if !defined(_M_ARM64EC)
 #pragma intrinsic(__rdtsc)
+#endif
 
 int64_t UnscaledCycleClock::Now() { return __rdtsc(); }
 
--- a/absl/base/internal/unscaledcycleclock.cc
+++ b/absl/base/internal/unscaledcycleclock.cc
@@ -132,6 +132,8 @@ double UnscaledCycleClock::Frequency() {
 #elif defined(_M_IX86) || defined(_M_X64)
 
 #if !defined(_M_ARM64EC)
+// ARM64EC is a Microsoft-designed ARM64 ABI compatible with AMD64 applications on ARM64 Windows 11.
+// The ARM64EC does not support __rdtsc as an intrinsic, but an inline function.
 #pragma intrinsic(__rdtsc)
 #endif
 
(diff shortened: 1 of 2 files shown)


absl/base/internal/unscaledcycleclock.cc
```diff
@@ -131,7 +131,11 @@ double UnscaledCycleClock::Frequency() {
 
 #elif defined(_M_IX86) || defined(_M_X64)
 
+#if !defined(_M_ARM64EC)
```
derekmauro, line 134: Using emulated `__rdtsc` on ARM seems sub-optimal. `UnscaledCycleClock` is an optional interface used by `CycleClock` when it is available. Have a look at `unscaledcycleclock.h`. I think it would be better to just disable this interface when `_M_ARM64EC` is defined.

absl/numeric/int128.h
```diff
@@ -46,7 +46,11 @@
 #define ABSL_INTERNAL_WCHAR_T __wchar_t
 #if defined(_M_X64)
 #include <intrin.h>
+// ARM64EC is a Microsoft-designed ARM64 ABI compatible with AMD64 applications on ARM64 Windows 11.
+// The ARM64EC does not support _umul128 as an intrinsic, but an inline function.
+#if !defined(_M_ARM64EC)
 #pragma intrinsic(_umul128)
```
derekmauro, line 52: I have a similar concern here. I wonder if the portable code is faster than the emulated intrinsic.

derekmauro (changes requested):

absl/base/internal/unscaledcycleclock.cc
mcfi: __rdstc on ARM64EC is implemented as _ReadStatusReg(ARM64_PMCCNTR_EL0), which is not emulated.

derekmauro: I don't think this changes anything. In `unscaledcycleclock.h` there is code that says "UnscaledCycleClock frequency is the same as CPU frequency", which isn't true of all ARM processors. It is also the case that `UnscaledCycleClock` is really only necessary on our production platform, which doesn't use `_M_ARM64EC`. I think it better to just disable this interface.

mcfi: Ok, sounds good, I'll abandon this PR and send a new one. Thanks.

mcfi: New PR here https://github.com/abseil/abseil-cpp/pull/1135

PR #1135, opened 2022-03-17
Exclude unsupported x64 intrinsics from ARM64EC

mcfi:
ARM64EC is a Microsoft-designed ARM64 ABI compatible with AMD64 applications on ARM64 Windows 11. The ARM64EC does not support _umul128 and __rdtsc as x64 intrinsics, though it provides inline function implementations for them, by emulation. Since the code already has portable code paths without using the intrinsics, instead of using the emulated intrinsic implementations, we use the said portable code paths for ARM64EC.


mcfi:
@derekmauro , could you please take a look at this new PR? Sorry for the delay.

PR #1137, opened 2022-03-18
Fix absl/string/numbers doesn't compile on Windows

MBkkt:
https://stackoverflow.com/a/22291538

### DIFF
--- a/absl/strings/numbers.h
+++ b/absl/strings/numbers.h
@@ -24,7 +24,26 @@
 #define ABSL_STRINGS_NUMBERS_H_
 
 #ifdef __SSE4_2__
+#if defined(_MSC_VER)
+/* Microsoft C/C++-compatible compiler */
+#include <intrin.h>
+#elif defined(__GNUC__) && (defined(__x86_64__) || defined(__i386__))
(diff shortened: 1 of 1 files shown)


absl/strings/numbers.h
```diff
@@ -24,7 +24,26 @@
 #define ABSL_STRINGS_NUMBERS_H_
 
 #ifdef __SSE4_2__
```
derekmauro, line 26: This does not look correct to me. `__SSE4_2__` is not defined by MSVC.

How was this change tested?

derekmauro:
I've submitted a change to our internal repo to fix this issue. It should make it to GitHub either today or Monday.

absl/strings/numbers.h
MBkkt: `__SSE4_2__` is not defined by MSVC. Its defined by some cmake script. 
And I agree with you, its not correct.
But why msvc can't define it in future?

MBkkt:
> I've submitted a change to our internal repo to fix this issue. It should make it to GitHub either today or Monday.

Don't see yet

derekmauro:
It is live now.

MBkkt:
Thanks, I will try

PR #1145, opened 2022-03-23
fix build with uclibc-ng

ffontaine:
uclibc-ng doesn't provide `getauxval` which results in the following build failure on arm or ppc with any user of abseil-cpp such as grpc:

```
/home/buildroot/autobuild/instance-0/output-1/host/opt/ext-toolchain/bin/../lib/gcc/arm-buildroot-linux-uclibcgnueabi/10.3.0/../../../../arm-buildroot-linux-uclibcgnueabi/bin/ld: /home/buildroot/autobuild/instance-0/output-1/host/arm-buildroot-linux-uclibcgnueabi/sysroot/usr/lib/libabsl_random_internal_randen_hwaes.so.2111.0.0: undefined reference to `getauxval'
```

To fix this build failure, check that `__UCLIBC__` is not defined before using `getauxval` (as Babel is not able to check function availability)

Fixes:
 - http://autobuild.buildroot.org/results/775f3ca3dedebff29e212b29dfa896b7613b7a02

Signed-off-by: Fabrice Fontaine <fontaine.fabrice@gmail.com>

### DIFF
--- a/absl/debugging/internal/vdso_support.cc
+++ b/absl/debugging/internal/vdso_support.cc
@@ -33,7 +33,7 @@
 #endif
 #include <unistd.h>
 
-#if defined(__GLIBC__) && \
+#if !defined(__UCLIBC__) && defined(__GLIBC__) && \
     (__GLIBC__ > 2 || (__GLIBC__ == 2 && __GLIBC_MINOR__ >= 16))
 #define ABSL_HAVE_GETAUXVAL
 #endif
--- a/absl/random/internal/randen_detect.cc
+++ b/absl/random/internal/randen_detect.cc
@@ -24,6 +24,11 @@
 
 #include "absl/random/internal/platform.h"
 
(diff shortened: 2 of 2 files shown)


derekmauro (changes requested):
The change proposed in this PR would break our Bazel build (which is our primary build system).

ffontaine:
Indeed, do you know the counterpart of cmake's `check_symbol_exists` (or autotools's `AC_CHECK_FUNCS`) in Bazel buildsystem?

derekmauro:
There is no equivalent in Bazel. Does using `__UCLIBC__` work?

ffontaine:
Yes, `__UCLIBC__ ` should also work. I'll update the PR.

ffontaine:
Do you have additional comments on this PR?

derekmauro:
I didn't realized you updated the PR. This should be fine, I will import it soon.

Issue #1143, opened 2022-03-23
To Support abseil-cpp in QNX 


c6supper:
## To Support abseil-cpp in QNX 
Bring some fresh air to automotive industry.

## Background
I was porting gPRC-cpp to QNX arm based platform, 
When I cross compile the gRPC, I found the abseil-cpp is not compatible with QNX too, so I made changes and have been successfully port them to QNX 704 aarch. 

## Proposed API Change (s)
I just forked and made some changes on your branch lts_2020_02_25, which is a submodule for gRPC, I have made changes ,and please help to review ,if it's suitable for a pull request, thx.
[Changes](https://github.com/abseil/abseil-cpp/commit/b3cea002cc0900a86637e8a367810955d4b2d0ca)

derekmauro:
Similar changes would be fine if applied to the main branch. Feel free to submit a pull request, or let me know and I can make the changes for you.

c6supper:
> Similar changes would be fine if applied to the main branch. Feel free to submit a pull request, or let me know and I can make the changes for you.

Thanks for you quick response, pull request have been created.
Please help to review , if it's compatilbe to main branch, you could help to  merge  to it either, thx.

PR #1146, opened 2022-03-24
to support QNX, confirmed with QNX 700 aarch

c6supper:
As we talked in #1143 ,  create this pull request to support QNX, Thanks for your quick respone @derekmauro .


derekmauro:
@c6supper I mentioned in #1143 that the patch needs to be applied to the master branch. This patch targets lts_2020_02_25.

c6supper:
Try to merge to master ,and pull later.

PR #1166, opened 2022-05-03
Allow forced enabling of debug assertions

krockot:
ABSL_ASSERT() is always a no-op when NDEBUG is defined. This introduces
an indepdendent secondary mechanism for force-enabling these assertions
by defining ABSL_ENABLE_DEBUG_ASSERTS.

See https://crbug.com/1245687 for motivation.


krockot:
Abandoning this in favor of an internal change.

PR #1186, opened 2022-05-24
Use `thread_local` when C++11 is supported

tamird:
This enables native TLS on platforms where it is available.

Before this change, native TLS is used via the `__thread` keyword only
on (Linux && (clang || libcstd++)). After this change, native TLS is
used on all systems where C++11 is available, regardless of compiler of
C++ standard library flavor (e.g. libc++).

/cc @drfloob


derekmauro:
As per https://github.com/abseil/abseil-cpp/blob/master/CONTRIBUTING.md#guidelines-for-pull-requests
> If you are a Googler, it is preferable to first create an internal CL and have it reviewed and submitted. The code propagation process will deliver the change to GitHub.

Skielersmommy (approved):
@derekmauro

tamird:
Closing in favor of internal change 450744401.

PR #1217, opened 2022-07-12
Disable sigaltstack() on macOS to make backtrace() work.

anpol:
There are many reports that sigaltstack() on Apple platforms stops backtrace() from tracing an original stack, e.g.

* https://github.com/llvm/llvm-project/commit/d39ead41056b00d10b6223543562cd304307ac9b
* https://reviews.llvm.org/D28265

If we disable an alternate stack for signal handler, we won't be able to catch stack overflow errors, but backrace() will work again.


Terry316:
If a Copyright licence can claim to hold a right of use for free then how can it charge conditions requiring the fee of time and or work and or contributions? The word contributions by itself, is just another way of saying cost.

PR #1223, opened 2022-07-20
fix(mutex): safely call snprintf

ElijahPepe:
In the PostSynchEvent() function, the pos integer uses an implementation of snprintf that is fundamentally unsafe: since the return value of snprintf is the number of characters that would have been written to the buffer, if an operation reaches the end of the buffer with more than one character discarded, the return value will be greater than the buffer size, requiring a check of the buffer's current size.


ElijahPepe:
@suertreus I've fixed an issue where the b integer wasn't being initialized properly (= over +=). Could you rerun build tests?

PR #1231, opened 2022-07-25
Buildfix OpenBSD.

3405691582:
OpenBSD does not expose `vasprintf` when `_XOPEN_SOURCE` is defined, but
`__bsd_locale_fallbacks.h` in the platform libc++ requires this to be
visible, thereby causing a compilation failure. In fact, `strptime` and
`vasprintf` are normally visible without defining any standard-visibility
`#defines`, so don't do anything special here.


derekmauro:
The files under `cctz/` come from https://github.com/google/cctz.

Please send your fix to https://github.com/google/cctz. It will be automatically imported here.

PR #1232, opened 2022-07-25
Make StrCat effective for single rvalue string

MBkkt:
Main motivation for me is patter like:

```C++
void SomeCall(Args&&...);

auto msg = StrCat(...);
LOG(unique_id, msg);
SomeCall(std::move(msg)); // internally have absl::StrCat for case when argument more than one or some of them not is string
```

### DIFF
--- a/absl/strings/str_cat.h
+++ b/absl/strings/str_cat.h
@@ -333,8 +333,14 @@ void AppendPieces(std::string* dest,
 
 ABSL_MUST_USE_RESULT inline std::string StrCat() { return std::string(); }
 
-ABSL_MUST_USE_RESULT inline std::string StrCat(const AlphaNum& a) {
-  return std::string(a.data(), a.size());
+template<typename Arg>
+ABSL_MUST_USE_RESULT inline std::string StrCat(Arg&& arg) {
+  if constexpr (std::is_same_v<Arg&&, std::string&&>) {
+    return std::move(arg);
+  } else {
+    const AlphaNum& a{std::forward<Arg>(arg)};
+    return std::string(a.data(), a.size());
+  }
 }
 
 ABSL_MUST_USE_RESULT std::string StrCat(const AlphaNum& a, const AlphaNum& b);


derekmauro:
Can you give a more concrete example?

absl/strings/str_cat.h
```diff
@@ -333,8 +333,14 @@ void AppendPieces(std::string* dest,
 
 ABSL_MUST_USE_RESULT inline std::string StrCat() { return std::string(); }
 
-ABSL_MUST_USE_RESULT inline std::string StrCat(const AlphaNum& a) {
-  return std::string(a.data(), a.size());
+template<typename Arg>
+ABSL_MUST_USE_RESULT inline std::string StrCat(Arg&& arg) {
+  if constexpr (std::is_same_v<Arg&&, std::string&&>) {
+    return std::move(arg);
```
glebm, line 339: This is a pessimizing move

MBkkt: Why? As I remember nrvo doesn't work for function parameters

glebm: Ah, you are probably right

MBkkt:
@derekmauro 
> Can you give a more concrete example?

I write function that looks like `log(some sink, parameters);`, this function concat parameters to single string.
Sometime parameters are single parameter.
And sometime it is rvalue string
I don't understand why do you want a make copy in this case.
This patch fix it.

derekmauro:
This PR breaks users that are relying on the function signature/overload resolution. I ran it through our code base verify this and found lots of code that no longer builds.

I asked for a complete example so that I might be able suggest an alternative. If you want to provide a complete example maybe I can help.

MBkkt:
> This PR breaks users that are relying on the function signature/overload resolution

Ok good point

anpol:
Excuse me, may I come up with an example?

```cpp
#include "absl/strings/str_cat.h"
#include "absl/strings/str_join.h"
#include "absl/log/log.h"

template<typename... Args>
void log_info(Args&&... args) // A library function.
{
    LOG(INFO) << absl::StrCat(std::forward<Args>(args)...);
}

void usage_example(std::vector<std::string> vec)
{
    log_info(absl::StrJoin(vec, ", "));
}
```

----

I would suggest adding an overload to handle this, e.g.:

```cpp
ABSL_MUST_USE_RESULT inline std::string StrCat(std::string&& arg) {
  return std::move(arg);
}
```

That doesn't seem to break anything.

BTW, original patch uses `if constexpr`, which is a C++17 feature, while Abseil is C++14.

MBkkt:
Do you test it?
@anpol 
I already don't remember.
But I probably have some problems with overload

PR #1246, opened 2022-08-10
- Fix deprecated header in visual studio version 17.3.0 C++20

wxilas21:
There was a header that was no longer used in c++20 of Visual studio version 17.3.0, so I fixed it.

### DIFF
--- a/absl/base/options.h
+++ b/absl/base/options.h
@@ -70,7 +70,7 @@
 // Include a standard library header to allow configuration based on the
 // standard library in use.
 #ifdef __cplusplus
-#include <ciso646>
+#include <iso646.h>
 #endif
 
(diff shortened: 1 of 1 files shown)


absl/base/options.h
```diff
@@ -70,7 +70,7 @@
 // Include a standard library header to allow configuration based on the
 // standard library in use.
 #ifdef __cplusplus
-#include <ciso646>
+#include <iso646.h>
```
derekmauro: This is vestigial #include. It can simply be removed.

derekmauro (changes requested):

derekmauro:
Thank you. This will be merged soon.

PR #1249, opened 2022-08-11
add new CMake option `ABSL_USE_SYSTEM_INCLUDES`

### LINKED ISSUE #964: cmake: include directories don't have SYSTEM property which can cause warnings
**Summary**

When integrating the abseil-cpp library into a cmake project, the include directories are not marked as SYSTEM. This enables the project-specific diagnostics to be applied to absl headers.

Is there an good way to mark include directories of libraries as `SYSTEM`? I couldn't find one from researching a bit.

Otherwise, I think it would be useful if abseil had an option to apply `SYSTEM` property to the `target_include_directories`.

**Details**

Here is a minimal cmake project to demonstrate the issue (this should reproduce with most common systems/compilers, I was using Ubuntu 20.04, cmake 3.20, gcc 9.3.0/clang 10.0):

```
matled@fg:~/src/demo% tail -n 10 CMakeLists.txt demo.cc
==> CMakeLists.txt <==
cmake_minimum_required(VERSION 3.5)
project(demo)
add_subdirectory(abseil-cpp)
add_executable(demo demo.cc)
target_link_libraries(demo absl::strings)
target_compile_options(demo PRIVATE -Wconversion -Werror)

==> demo.cc <==
#include <absl/strings/str_cat.h>

int main() {
  printf("%s\n", absl::StrCat("foo", "bar").c_str());
  return 0;
}
matled@fg:~/src/demo% rm -rf build && cmake -S . -B build && cmake --build build --target demo --verbose
-- The C compiler identification is GNU 9.3.0
-- The CXX compiler identification is GNU 9.3.0
[...]
/usr/bin/c++  -I/home/matled/src/demo/abseil-cpp -Wconversion -Werror [...]
In file included from /home/matled/src/demo/demo.cc:1:
/home/matled/src/demo/abseil-cpp/absl/strings/str_cat.h: In constructor ‘absl::Hex::Hex(absl::PadSpec, uint64_t)’:
/home/matled/src/demo/abseil-cpp/absl/strings/str_cat.h:171:19: error: conversion from ‘int’ to ‘uint8_t’ {aka ‘unsigned char’} may change value [-Werror=conversion]
[...]
cc1plus: all warnings being treated as errors
make[3]: *** [CMakeFiles/demo.dir/build.make:76: CMakeFiles/demo.dir/demo.cc.o] Error 1
make[3]: Leaving directory '/mnt/c/Users/matled/source/demo/build'
make[2]: *** [CMakeFiles/Makefile2:551: CMakeFiles/demo.dir/all] Error 2
make[2]: Leaving directory '/mnt/c/Users/matled/source/demo/build'
make[1]: *** [CMakeFiles/Makefile2:558: CMakeFiles/demo.dir/rule] Error 2
make[1]: Leaving directory '/mnt/c/Users/matled/source/demo/build'
make: *** [Makefile:124: demo] Error 2
```

When include directories use the `-isystem` flag instead of `-I`, warnings are suppressed. cmake sets this option based on the `SYSTEM` property being specified on `target_include_directories`. I could not find an easy way to set this option outside of the abseil cmake files.

I've verified manually that adding the `SYSTEM` property to the two uses of `target_include_directories` in `function(absl_cc_library)` solves the issue.

An option in the abseil-cpp cmake file to mark the include directories as `SYSTEM` would be one solution for this.

evanacox:
When using Abseil as a dependency with CMake, files that include Abseil headers and have stricter warning settings can emit compiler warnings. 

CMake does provide a way to express that warnings from headers in a specific include path should be ignored, but it is expected that the library providing a target decides this rather than the user of a target (unless Abseil wants downstream users manually tinkering with the properties of Abseil targets in order to mark includes as `SYSTEM` themselves).

This adds the new option `ABSL_USE_SYSTEM_INCLUDES`, which causes Abseil's include directories be marked with `SYSTEM`. It is disabled by default, but users who wish to have this behavior can enable it. 

Fixes #964.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -71,7 +71,7 @@ if((${CMAKE_VERSION} VERSION_GREATER_EQUAL 3.8) AND (NOT ABSL_PROPAGATE_CXX_STD)
 endif()
 
 option(ABSL_USE_SYSTEM_INCLUDES
-  "Mark C++ include directories with `SYSTEM` to prevent warnings from headers when using Abseil with incompatible warning flags"
+  "Silence warnings in Abseil headers by marking them as `SYSTEM` includes"
   OFF)
 
 list(APPEND CMAKE_MODULE_PATH
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -70,6 +70,10 @@ if((${CMAKE_VERSION} VERSION_GREATER_EQUAL 3.8) AND (NOT ABSL_PROPAGATE_CXX_STD)
   message(WARNING "A future Abseil release will default ABSL_PROPAGATE_CXX_STD to ON for CMake 3.8 and up. We recommend enabling this option to ensure your project still builds correctly.")
 endif()
 
+option(ABSL_USE_SYSTEM_INCLUDES
+  "Mark C++ include directories with `SYSTEM` to prevent warnings from headers when using Abseil with incompatible warning flags"
+  OFF)
+
 list(APPEND CMAKE_MODULE_PATH
   ${CMAKE_CURRENT_LIST_DIR}/CMake
   ${CMAKE_CURRENT_LIST_DIR}/absl/copts
--- a/CMake/AbseilHelpers.cmake
+++ b/CMake/AbseilHelpers.cmake
@@ -26,6 +26,12 @@ if(NOT DEFINED ABSL_IDE_FOLDER)
   set(ABSL_IDE_FOLDER Abseil)
 endif()
 
+if(ABSL_USE_SYSTEM_INCLUDES)
+  set(ABSL_INTERNAL_INCLUDE_WARNING_GUARD SYSTEM)
+else()
+  set(ABSL_INTERNAL_INCLUDE_WARNING_GUARD "")
+endif()
+
 # absl_cc_library()
 #
 # CMake function to imitate Bazel's cc_library rule.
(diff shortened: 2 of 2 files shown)


CMakeLists.txt
```diff
@@ -70,6 +70,10 @@ if((${CMAKE_VERSION} VERSION_GREATER_EQUAL 3.8) AND (NOT ABSL_PROPAGATE_CXX_STD)
   message(WARNING "A future Abseil release will default ABSL_PROPAGATE_CXX_STD to ON for CMake 3.8 and up. We recommend enabling this option to ensure your project still builds correctly.")
 endif()
 
+option(ABSL_USE_SYSTEM_INCLUDES
+  "Mark C++ include directories with `SYSTEM` to prevent warnings from headers when using Abseil with incompatible warning flags"
```
derekmauro: In my opinion this sentence hard to parse. What do you think about something like "Silence warnings in Abseil headers by marking them as `SYSTEM` includes"?

derekmauro (changes requested):
Thank you for this PR. It addresses a common issue for us.

CMakeLists.txt
evanacox: Fair enough, I didn't think very much about the message. Updated in new commit

evanacox:
Meant to request re-review earlier

ojura:
Sorry to be rude, but why was #160, which was doing the basically same thing, left rotting for 4 years?

Anyway glad the thing is resolved now

PR #1250, opened 2022-08-12
Workaround bug in GCC 7.2

jonathan-conder-sm:
See https://gcc.gnu.org/bugzilla/show_bug.cgi?id=83830.

Before GCC 7.4, `__cpp_lib_has_unique_object_representations` was defined but `has_unique_object_representations_v` was not.

I don't personally use old compiler versions, but [conan-center-index](https://github.com/conan-io/conan-center-index) does for CI builds, and they want me to attempt to merge this upstream before accepting a patch for their abseil package (https://github.com/conan-io/conan-center-index/pull/11946).


jonathan-conder-sm:
Thanks!

PR #1267, opened 2022-09-01
Synchronize the --stderrthreshold flag value.

### LINKED ISSUE #1266: absl::SetStderrThreshold does not cause the FLAGS_stderrthreshold flag value to change
**Describe the bug**

In a command line utility I've written I'm using Abseil logging and flags. It makes no sense for this utility to ever log to anyplace but stderr, so I'd like to change the (effective) default value of FLAGS_stderrthreshold to zero. However, I can't figure out a way to do this in a satisfying way.

**Steps to reproduce the bug**

As far as I can tell, my two options are:

1. Call `absl::SetStderrThreshold(absl::LogSeverityAtLeast::kInfo);` early, before `absl::ParseCommandLine` and `absl::InitializeLog`. The problem with this approach is that this new value is not reflected in the flag, and this is exposed when the user passes `--helpfull`, which tells the user the default value of the flag is 2 (despite the effective value being zero, as this is what `absl::StderrThreshod()` is returning).
2. Call `absl::SetFlag(&FLAGS_stderrthreshold, 0)`. This keeps the flag and the global var that holds the value the logging library actually uses in sync, and the help text displays that the current value is `default: 2; currently: 0`, which I can live with. I'd rather see `default: 0` but that is a minor thing. However, to do this I must also `ABSL_DECLARE_FLAG(int, stderrthreshold);`, which is a bad idea -- for one, [absl/log/flags.h](https://github.com/abseil/abseil-cpp/blob/43d3c7a4e290ee96684735daf7b1d528c30a7943/absl/log/flags.h#L26-L27) does not advertise the type of the flag, just its name, so there is no way for me to declare it myself without making assumptions.

Possible fix:

Maybe a call to `absl::SetFlag(&FLAGS_stderrthreshold, absl::StderrThreshold())` belongs in `SyncLoggingFlags()` in [absl/log/flags.cc](https://github.com/abseil/abseil-cpp/blob/43d3c7a4e290ee96684735daf7b1d528c30a7943/absl/log/flags.cc#L40-L41)?

**What version of Abseil are you using?**

92fdbfb301f8b301b28ab5c99e7361e775c2fb8a (the "Release the Abseil Logging library" commit)

**What operating system and version are you using**

n/a

**What compiler and version are you using?**

n/a

**What build system are you using?**

n/a

**Additional context**

I am very happy Abseil logging was released at long last!

matta:
The --minloglevel and --log_prefix were already synchronized, so this
corrects an omission.  I also rewrote the tests for these three flags to
cover more of the API surface as well as more edge cases.

Fixes #1266

### DIFF
--- a/absl/log/flags.cc
+++ b/absl/log/flags.cc
@@ -39,6 +39,8 @@ namespace {
 void SyncLoggingFlags() {
   absl::SetFlag(&FLAGS_minloglevel, static_cast<int>(absl::MinLogLevel()));
   absl::SetFlag(&FLAGS_log_prefix, absl::ShouldPrependLogPrefix());
+  absl::SetFlag(&FLAGS_stderrthreshold,
+                static_cast<int>(absl::StderrThreshold()));
 }
 
 bool RegisterSyncLoggingFlags() {
--- a/absl/log/flags_test.cc
+++ b/absl/log/flags_test.cc
@@ -53,38 +53,104 @@ TEST_F(LogFlagsTest, StderrKnobsDefault) {
 }
 
 TEST_F(LogFlagsTest, SetStderrThreshold) {
-  absl::SetFlag(&FLAGS_stderrthreshold,
-                static_cast<int>(absl::LogSeverityAtLeast::kInfo));
+  // Verify that the API and the flag agree.
+  EXPECT_EQ(absl::GetFlag(FLAGS_stderrthreshold),
+            static_cast<int>(absl::StderrThreshold()));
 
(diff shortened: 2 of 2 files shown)


rogeeff:
Internally we also need to maintain the state to old logtostderr and alsologtostderr flags (until they are eliminated). This makes the whole synchronization business very tricky. We do not want to have different behavior in open source and inside google3. There are a couple workarounds you can do in your code if the --help output really bothers you, including setting the flag value via reflection: absl::FindCommandLineFlag("stderrthreshold")->ParseFrom("0", &errors) (plus some error checking).

So for now I'd like to keep the status quo.

Issue #1201, opened 2022-06-22
abseil's absl::is_trivially_copy_assignable uses deprecated, broken compiler builtin


royjacobson:
As a side effect of implementing DR2171 for clang, we have discovered that `absl::is_trivially_copy_assignable` uses the `__has_trivial_assign` compiler builtin.

This compiler builtin has bad semantics when deleted functions are involved because clang considers them trivial for this purpose. Specifically it appears that MSVC stl delete the copy assignment operator for ` volatile& std::pair` which causes abseil to assert when instantiating `absl::Optional<std::pair<int, int>>` with MSVC stl.

This has been discussed over at https://reviews.llvm.org/D127593. We might revert this change momentarily, but it would be helpful if abseil will stop using the `__has_trivial...` builtins.

Note that GCC uses the same 'deleted is trivial' semantics as clang, and that those builtins have been deprecated for quite some time: this is a discussion of the same issue from 2017: https://github.com/llvm/llvm-project/issues/33063

sylvestre:
On the Mozilla side: https://bugzilla.mozilla.org/show_bug.cgi?id=1779528 (firefox uses libwebrtc which uses abseil-cpp)

sseckler:
In addition to `__has_trivial_assign`, abseil uses `__has_trivial_copy` and `__has_trivial_constructor` in type_traits.h which are also deprecated.

keith:
https://github.com/abseil/abseil-cpp/pull/1289

PR #1277, opened 2022-09-15
Update deprecated builtins __has_trivial to __is_trivially

aaronmondal:
The old builtins have been depracated in Clang 16. Since they are also present in GCC it should be fine to just use the new variants.

Fixes Issue #1201.

### DIFF
--- a/absl/meta/type_traits.h
+++ b/absl/meta/type_traits.h
@@ -301,7 +301,7 @@ struct is_trivially_destructible
 #ifdef ABSL_HAVE_STD_IS_TRIVIALLY_DESTRUCTIBLE
     : std::is_trivially_destructible<T> {
 #else
-    : std::integral_constant<bool, __has_trivial_destructor(T) &&
+    : std::integral_constant<bool, __is_trivially_destructible(T) &&
                                    std::is_destructible<T>::value> {
 #endif
 #ifdef ABSL_HAVE_STD_IS_TRIVIALLY_DESTRUCTIBLE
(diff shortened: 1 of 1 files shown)


aaronmondal:
Well that seems to fail :rofl:
@derekmauro I tested this with Clang and GCC via `bazel test ...` Is there some other command i need to run to test this?

aaronmondal:
Maybe tests are not actually failing but the CI servers are just unreachable. At least the `Details` links just give me `fusion2’s server IP address could not be found.`

derekmauro:
CI is working correctly, but the results are IP blocked. (There is a way to make the results public, but I haven't had time to figure out how to do that.)

I actually thought https://github.com/abseil/abseil-cpp/commit/cfe27e79cfcbefb2b4479e04f80cbb299bc46965 fixed this issue. The remaining uses of `__has_trivial` should be preprocessor guarded. What platform is demonstrating the issue that this change addresses?

I think the issue with your change is that some of the `__is_trivially` functions take two arguments. If I recall correctly, the person who committed cfe27e79cfcbefb2b4479e04f80cbb299bc46965 said that the simpler fix (the one this change is proposing) had some tricky problems, so we decided to just preprocessor guard the old code.

aaronmondal:
@derekmauro The issue arose with Upstream Clang/LLVM, but we were using an abseil version before cfe27e7.

I'll see if cfe27e7 already fixes the issue. If so, I'll close this PR.

Regardless, I think it is strange that `bazel test ...` passes. I would expect obvious errors such as wrong argcounts to immediately be detected by it.

aaronmondal:
cfe27e7 already fixed the issue. Cool! Closing.

keith:
There seem to be a few more cases here https://github.com/abseil/abseil-cpp/pull/1289

PR #1285, opened 2022-10-02
`const_cast`-away volatile during copy

jrandolf:
In Clang 16, `HandleEdgeCase` results in an error due to a `volatile` qualifier:

```
 cannot initialize a parameter of type 'void *' with an lvalue of type 'volatile char *'
```
 
This PR removes the `volatile` qualifier during copy in `HandleEdgeCase`.


derekmauro:
Tests are failing, probably because of the issue described on line 337.

@vslashg

jrandolf:
The issue is not repaired in GCC (not unexpected). I've limited the fix to clang for now.

derekmauro:
> The issue is not repaired in GCC (not unexpected). I've limited the fix to clang for now.

I played with the godbolt link in the bug report and this issue does not appear to be fixed in Clang 6, which is still supported according to https://github.com/google/oss-policies-info/blob/main/foundational-cxx-support-matrix.md. Can this be conditioned more precisely?

jrandolf:
> > The issue is not repaired in GCC (not unexpected). I've limited the fix to clang for now.
> 
> I played with the godbolt link in the bug report and this issue does not appear to be fixed in Clang 6, which is still supported according to https://github.com/google/oss-policies-info/blob/main/foundational-cxx-support-matrix.md. Can this be conditioned more precisely?

Done.

PR #1290, opened 2022-10-07
Add `!define(__clang__)` for GNU compiler matching

jrandolf:
The original PR (https://github.com/abseil/abseil-cpp/pull/1285) left out a condition that did not enforce matching the GNU compiler. The problem comes from clang also defining the `__GNUC__` macro, so clang would escape through the first condition.

### DIFF
--- a/absl/strings/charconv.cc
+++ b/absl/strings/charconv.cc
@@ -347,7 +347,8 @@ bool HandleEdgeCase(const strings_internal::ParsedFloat& input, bool negative,
     // https://bugs.llvm.org/show_bug.cgi?id=37778
     // https://gcc.gnu.org/bugzilla/show_bug.cgi?id=86113
     constexpr ptrdiff_t kNanBufferSize = 128;
-#if defined(__GNUC__) || (defined(__clang__) && __clang_major__ < 7)
+#if (defined(__GNUC__) && !defined(__clang__)) || \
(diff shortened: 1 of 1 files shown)


derekmauro:
Somehow this is breaking a test when using our production Clang compiler. It may take me a while to look into why this is happening. Sorry about this.

jrandolf:
Have you verified the production compiler has the patch against the clang bug linked in the comments?

derekmauro:
> Have you verified the production compiler has the patch against the clang bug linked in the comments?

Yes I have.

LumioseSil:
Any updates on this?

derekmauro:
> Any updates on this?

I actually tried this recently and it is still and issue with our production compiler. I still don't know why and I haven't dug any deeper.

derekmauro:
My theory here is that while LLVM fixed the codegen for `__builtin_nan`, that isn't actually what we are calling. [We are calling `std::nan`](https://github.com/abseil/abseil-cpp/blob/34dd639a768761e555b080224a1381a459ab2885/absl/strings/charconv.cc#L120-L125), and as far as I can tell, libc++ just calls into glibc.
https://github.com/llvm/llvm-project/blob/6c7acc6409cb13c7df264a3d07e2ff2b6119fc40/libcxx/include/cmath#L341

I think a change to use `__builtin_nan` if available will allow this change to work.

Issue #1301, opened 2022-10-23
Value of CMAKE_SYSTEM_PROCESSOR (x86) is unknown and cannot be used to set ABSL_RANDOM_RANDEN_COPTS


KivApple:
When building the library using Emscripten compiler I got a warning:

```
CMake Warning at lib/abseil-cpp/absl/copts/AbseilConfigureCopts.cmake:70 (message):
  Value of CMAKE_SYSTEM_PROCESSOR (x86) is unknown and cannot be used to set
  ABSL_RANDOM_RANDEN_COPTS
```

abseil version: master

compiler version:
```
emcc (Emscripten gcc/clang-like replacement + linker emulating GNU ld) 3.1.13 (531257621816c200bc7c3be53129494afd029aec)
clang version 15.0.0 (https://github.com/llvm/llvm-project 5c6ed60c517c47b25b6b25d8ac3666d0e746b0c3)
Target: wasm32-unknown-emscripten
Thread model: posix
InstalledDir: C:\Users\kiv\Apps\emsdk\upstream\bin
```

KivApple:
Fix: https://github.com/abseil/abseil-cpp/pull/1302

PR #1302, opened 2022-10-23
Fix warning on Emscripten target

KivApple:
Fix for https://github.com/abseil/abseil-cpp/issues/1301


derekmauro:
I removed the warning in bc097314fe5ba91e7f358159f84d7b9477502f75 instead since it doesn't provide much value.

PR #1305, opened 2022-10-29
Fix skipping unused portions in TimeZoneInfo::Load() for version 1 zonefiles.

chenshuo:
In theory, leap-second record of version 1 zonefiles is 4 + 4 bytes per record.
https://www.rfc-editor.org/rfc/rfc8536.html#section-3.2

In practice, this commit won't make any difference because we return early at line 434 if hdr.leapcnt != 0.


derekmauro:
Changes for this code need to be sent to https://github.com/google/cctz.

PR #1306, opened 2022-11-01
build: attempting to fix https://github.com/abseil/abseil-cpp/issues/326

### LINKED ISSUE #326: Foundation framework link option leaking into Android builds
We have a project that builds a native library for Android that depends on Abseil. With this change:
https://github.com/abseil/abseil-cpp/commit/ca3f87560a0eef716195cadf66dc6b938a579ec6#diff-020a0864e5179d94daa487a30519b6acR79
our builds broke when performed on MacOS with the following error:
```
Execution platform: @bazel_tools//platforms:host_platform
external/androidndk/ndk/toolchains/aarch64-linux-android-4.9/prebuilt/darwin-x86_64/lib/gcc/aarch64-linux-android/4.9.x/../../../../aarch64-linux-android/bin/ld: cannot find Foundation: No such file or dir
ectory
```
We suspect we may somehow be specifying crosstool options or the target platform incorrectly, but what's surprising is up until this point everything worked. (We're actually not specifying these options at all explicitly, but rather relying on transitions built into the `android_binary` rule.)

Although Abseil may very well be doing the right thing here, we wanted to reach out to see if anyone was willing and able to provide us some guidance. This may be as much a Bazel question as it is an Abseil question, and we're following up with Bazel folks as well.

If this falls outside the scope of the support issues in this project, feel free to close.

Thanks!

alyssawilk:
Fixes https://github.com/abseil/abseil-cpp/issues/326

### DIFF
--- a/absl/time/internal/cctz/BUILD.bazel
+++ b/absl/time/internal/cctz/BUILD.bazel
@@ -73,9 +73,6 @@ cc_library(
         "include/cctz/zone_info_source.h",
     ],
     linkopts = select({
-        ":osx": [
-            "-framework Foundation",
-        ],
         ":ios": [
             "-framework Foundation",
         ],
--- a/absl/time/internal/cctz/BUILD.bazel
+++ b/absl/time/internal/cctz/BUILD.bazel
@@ -21,20 +21,6 @@ filegroup(
     srcs = glob(["testdata/zoneinfo/**"]),
 )
 
-config_setting(
-    name = "osx",
-    constraint_values = [
-        "@platforms//os:osx",
-    ],
-)
-
-config_setting(
-    name = "ios",
-    constraint_values = [
-        "@platforms//os:ios",
-    ],
-)
-
 ### libraries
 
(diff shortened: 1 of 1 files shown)


alyssawilk:
Alternately cl/485404590 but the presubmits tell me to do OSS first and https://github.com/abseil/abseil-cpp/blob/0064d9db90d32d35e9f9d70e2df4ddf8d0ab1257/CONTRIBUTING.md tell me to do google3 first.  I'm happy to do either and help reconcile the documentation if y'all tell me what's actually correct.

derekmauro:
Sorry, the presubmit message is confusing. Changes to this code need to go to https://github.com/google/cctz. Abseil imports it from there.

derekmauro:
Re-opening this so the CI will run.

alyssawilk:
oh sorry the change is slightly different in cctz now - pushed the latest

alyssawilk:
closing now we verified CI :-)

PR #1310, opened 2022-11-03
Use UB impl of launder on when using clang < 6 and c++17

### LINKED ISSUE #1309: std::launder was not implemented in clang until version 6
Edit: @glandium is correct in that this was not implemented in clang until version 8. This is easily verifiable using Compiler Explorer.

std::launder was implemented in clang 6.0, in revision https://github.com/llvm/llvm-project/commit/9180eb1f4ab96481413e93d35f35c3685de3084c . This causes a build failure when building with clang 5 in c++17 mode, see https://github.com/abseil/abseil-cpp/blob/master/absl/functional/internal/any_invocable.h#L198.  It should be possible to check for clang < 6 and fall back to the UB based implementation here https://github.com/abseil/abseil-cpp/blob/master/absl/functional/internal/any_invocable.h#L204.

fuzzy-coats:
Fixes #1309

### DIFF
--- a/absl/functional/internal/any_invocable.h
+++ b/absl/functional/internal/any_invocable.h
@@ -195,7 +195,8 @@ union TypeErasedState {
 template <class T>
 T& ObjectInLocalStorage(TypeErasedState* const state) {
   // We launder here because the storage may be reused with the same type.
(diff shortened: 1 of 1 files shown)


derekmauro:
As mentioned in #1309, closing as clang < 6 is unsupported.

PR #1323, opened 2022-11-24
Fix CMake whitespace warnings.

kisielk:
eg:
CMake Warning (dev) at CMake/AbseilDll.cmake:408:
  Syntax Warning in cmake code at column 17

  Argument not separated from preceding token by whitespace.


derekmauro:
Thank you for sending a pull request. We did get an identical change from an internal contributor a bit before you sent us this pull request, so this issue is fixed by 9f4bde36965c4de0781ea292fa7737a4fd60f265.

PR #1325, opened 2022-11-26
Few crc32c small improvements

MBkkt:
* Allow ToCrc32c in constexpr context (constexpr function also is inline)
* Remove unnecessary big include
* Fix my local compilation with invalid crc32c_t construction


absl/crc/internal/crc_memcpy_x86_64.cc
```diff
@@ -164,7 +164,7 @@ crc32c_t AcceleratedCrcMemcpyEngine<vec_regions, int_regions>::Compute(
     void* __restrict dst, const void* __restrict src, std::size_t length,
     crc32c_t initial_crc) const {
   constexpr std::size_t kRegions = vec_regions + int_regions;
-  constexpr crc32c_t kCrcDataXor = crc32c_t{0xffffffff};
+  constexpr crc32c_t kCrcDataXor = ToCrc32c(0xffffffffU);
```
MBkkt, line 167: I more like, but I don't know your codestyle
```suggestion
  constexpr auto kCrcDataXor = ToCrc32c(0xffffffffU);
```

PR #1349, opened 2022-12-29
direct_mmap: Use off_t on linux

kraj:
off64_t is not provided without defining _LARGEFILE64_SOURCE on musl this define is not defined automatically like glibc where it gets defined when _GNU_SOURCE is defined. Using off_t makes it portable across musl/glibc and for using 64bit off_t on glibc 32bit systems -D_FILE_OFFSET_BITS=64 can be defined during build via CXXFLAGS

Signed-off-by: Khem Raj <raj.khem@gmail.com>


orbea:
Why was this reverted?

derekmauro:
It broke an internal platform. b/265015843 is our tracking bug for this. I haven't had time to investigate.

PR #1364, opened 2023-01-18
Add ABSL_NO_SYMBOLIZE definition to avoid doing symbolization when ne…

ZezhongWang:
XBox platform currently doesn't support dbghelp.lib/.ddl. Add this definition to turn off symbolization when needed,


derekmauro:
Pull requests that add new features to LTS branches are not accepted. You may propose adding a feature to the main development branch.

Please also note that we don't like features flags that are defined on the command line. A preprocessor flag that is defined on the Xbox platform is preferred.

PR #1367, opened 2023-01-24
Make absl::optional::operator-> constexpr

hferreiro:


absl/types/optional.h
```diff
@@ -414,7 +414,7 @@ class optional : private optional_internal::optional_data<T>,
   // `optional` is empty, behavior is undefined.
   //
   // If you need myOpt->foo in constexpr, use (*myOpt).foo instead.
```
derekmauro, line 416: This comment is no longer accurate.

derekmauro (changes requested):
Please add a test to optional_test.cc that demonstrates that this works.

PR #1387, opened 2023-02-10
Remove Workarounds for Old Clang Bug

LumioseSil:
https://bugs.llvm.org/show_bug.cgi?id=38289 has been addressed since 2019.

We just have to check to see if the clang being used to compile with is any version before major version 9.

### DIFF
--- a/absl/numeric/int128.cc
+++ b/absl/numeric/int128.cc
@@ -111,7 +111,7 @@ uint128 MakeUint128FromFloat(T v) {
   return MakeUint128(0, static_cast<uint64_t>(v));
 }
 
-#if defined(__clang__) && !defined(__SSE3__)
+#if defined(__clang__) && (__clang_major__ < 9) && !defined(__SSE3__)
 // Workaround for clang bug: https://bugs.llvm.org/show_bug.cgi?id=38289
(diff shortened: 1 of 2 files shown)


LumioseSil:
@derekmauro Could you please look at this?

derekmauro (commented):
According to https://github.com/google/oss-policies-info/blob/main/foundational-cxx-support-matrix.md, we still support Clang 6.0.0.

According to https://en.wikipedia.org/wiki/Clang, Clang 6.0.0 was released in March of 2018.

So I think we still need these checks. Perhaps it is possible to detect if the code is being compiled with a version of clang that has the fix?

LumioseSil:
> According to https://github.com/google/oss-policies-info/blob/main/foundational-cxx-support-matrix.md, we still support Clang 6.0.0.
> 
> According to https://en.wikipedia.org/wiki/Clang, Clang 6.0.0 was released in March of 2018.
> 
> So I think we still need these checks. Perhaps it is possible to detect if the code is being compiled with a version of clang that has the fix?

Done!

AtariDreams:
@derekmauro

LumioseSil:
@derekmauro It passed!

PR #1388, opened 2023-02-15
Fix Read1To3 on Big Endian for 2 byte inputs

miladfarca:
Current implementation does not work with 2 byte inputs as mem1 and mem2 will both point to the least significant byte.

### DIFF
--- a/absl/hash/internal/hash.h
+++ b/absl/hash/internal/hash.h
@@ -1083,7 +1083,7 @@ class ABSL_DLL MixingHashState : public HashStateBase<MixingHashState> {
     unsigned char significant0 = mem0;
 #else
     unsigned char significant2 = mem0;
-    unsigned char significant1 = mem1;
+    unsigned char significant1 = len == 2 ? mem0 : mem1;
     unsigned char significant0 = mem2;
 #endif
     return static_cast<uint32_t>(significant0 |                     //


absl/hash/internal/hash.h
```diff
@@ -1083,7 +1083,7 @@ class ABSL_DLL MixingHashState : public HashStateBase<MixingHashState> {
     unsigned char significant0 = mem0;
 #else
     unsigned char significant2 = mem0;
-    unsigned char significant1 = mem1;
+    unsigned char significant1 = len == 2 ? mem0 : mem1;
```
derekmauro, line 1086: I actually think we don't need any of endian complications here. We should be fine returning `(mem0 << 16) | (mem1 << 8) | mem2`.

This only needs to be deterministic and use all the bytes at least once. It can duplicate bytes as long as it does so consistently.

See also https://github.com/abseil/abseil-cpp/blob/b54044578175b6949fb151b4d8be9153d7093c31/absl/hash/internal/low_level_hash.cc#L97-L98.

miladfarca: But then the returned value won't match this on BE which will cause a number of test failures:
https://github.com/abseil/abseil-cpp/blob/b54044578175b6949fb151b4d8be9153d7093c31/absl/hash/internal/hash.h#L1105

derekmauro: Which tests fail and why does the return value have to match? Abseil hash is randomized and has no stability guarantees at all across platforms or even across runs on the same platform.

miladfarca: This recently added test case fails:
https://github.com/abseil/abseil-cpp/blob/2d4c6872da56fa0d8156b01ab5658348d9c5de4d/absl/hash/hash_test.cc#L1259

derekmauro: It took me a while to realize that there is an `IntegralFastPath` above that this has to match, so this does have to be a precise read, it can't just use all the bytes. LowLevelHash has no such need.

I've sent your change to another colleague for review in case there is some other trick that can be used, but otherwise looks good.

miladfarca: Not a problem, thanks for reviewing it.

PR #1389, opened 2023-02-16
Prefer static variable std::string::npos over access via instance

LumioseSil:
npos is the same for all strings, so comparing to std::string::npos over the npos field of the string instances being check is clearer.

### DIFF
--- a/absl/strings/match.h
+++ b/absl/strings/match.h
@@ -45,11 +45,11 @@ ABSL_NAMESPACE_BEGIN
 // Returns whether a given string `haystack` contains the substring `needle`.
 inline bool StrContains(absl::string_view haystack,
                         absl::string_view needle) noexcept {
-  return haystack.find(needle, 0) != haystack.npos;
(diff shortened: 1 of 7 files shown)


absl/strings/match.h
```diff
@@ -45,11 +45,11 @@ ABSL_NAMESPACE_BEGIN
 // Returns whether a given string `haystack` contains the substring `needle`.
 inline bool StrContains(absl::string_view haystack,
                         absl::string_view needle) noexcept {
-  return haystack.find(needle, 0) != haystack.npos;
+  return haystack.find(needle, 0) != std::string::npos;
```
derekmauro, line 48: Many of these are wrong. For example, here this should be `absl::string_view::npos`. There is no guarantee that `absl::string_view::npos == std::string::npos`.

I disagree that this is a clear improvement. Here if you were to change the type of the input from `absl::string_view` to `std::string` or even `std::string_view`, you'd also have to change the type you compare it to, whereas with the current version, you do not.

PR #1391, opened 2023-02-16
Use noexcept for move constructors

LumioseSil:
We should not throw exceptions during move constructors.

### DIFF
--- a/absl/status/internal/statusor_internal.h
+++ b/absl/status/internal/statusor_internal.h
@@ -225,7 +225,7 @@ class StatusOrData {
     return *this;
   }
 
-  StatusOrData& operator=(StatusOrData&& other) {
+  StatusOrData& operator=(StatusOrData&& other) noexcept {
     if (this == &other) return *this;
     if (other.ok())
       Assign(std::move(other.data_));
--- a/absl/status/statusor.cc
+++ b/absl/status/statusor.cc
@@ -40,7 +40,8 @@ BadStatusOrAccess& BadStatusOrAccess::operator=(
   return *this;
 }
 
-BadStatusOrAccess& BadStatusOrAccess::operator=(BadStatusOrAccess&& other) {
+BadStatusOrAccess& BadStatusOrAccess::operator=(
+    BadStatusOrAccess&& other) noexcept {
   // Ensure assignment is correct regardless of whether this->InitWhat() has
   // already been called.
   other.InitWhat();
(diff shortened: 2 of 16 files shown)


absl/status/statusor.cc
```diff
@@ -50,7 +51,7 @@ BadStatusOrAccess& BadStatusOrAccess::operator=(BadStatusOrAccess&& other) {
 }
 
 BadStatusOrAccess::BadStatusOrAccess(BadStatusOrAccess&& other)
-    : status_(std::move(other.status_)) {}
+    : status_(std::move(other.status_)) noexcept {}
```
derekmauro: This doesn't build because the syntax here is wrong. Please be sure to run the tests first.

LumioseSil:
@derekmauro Done!

derekmauro (changes requested):
The build is still broken. Please try to build the code.

absl/container/inlined_vector.h
```diff
@@ -493,7 +493,7 @@ class InlinedVector {
   //
   // NOTE: as a result of calling this overload, `other` is left in a valid but
   // unspecified state.
-  InlinedVector& operator=(InlinedVector&& other) {
+  InlinedVector& operator=(InlinedVector&& other) noexcept {
```
brenoguim: This is not a move constructor. Is this intended?

brenoguim: Reading through the patch it seems it was added to other move assignment operators as well. Never mind.

But I wonder if just adding noexcept is the right thing to do. This means if any operation inside the move constructor throws, it will terminate the program.
Generally, containers use `std::move_if_noexcept` to use the copy constructor in case the move constructor throws.
And then they use `noexcept( expression that determines if internal operations may throw )` to make the constructor conditionally `noexcept`

absl/status/internal/statusor_internal.h
```diff
@@ -346,43 +342,43 @@ struct MoveCtorBase<T, false> {
   MoveCtorBase(const MoveCtorBase&) = default;
   MoveCtorBase(MoveCtorBase&&) = delete;
   MoveCtorBase& operator=(const MoveCtorBase&) = default;
-  MoveCtorBase& operator=(MoveCtorBase&&) = default;
+  MoveCtorBase& operator=(MoveCtorBase&&) noexcept = default;
```
MBkkt: noexcept for default not required.
If it's really noexcept it will be noexcept implicit

absl/container/inlined_vector.h
```diff
@@ -493,7 +493,7 @@ class InlinedVector {
...
     if (ABSL_PREDICT_TRUE(this != std::addressof(other))) {
       MoveAssignment(MoveAssignmentPolicy{}, std::move(other));
```
MBkkt: Is MoveAssignment noexcept?

I suspect it can be not true for case when inlined elements doesn't have noexcept move assignment

LumioseSil:
@derekmauro @MBkkt Done!

PR #1392, opened 2023-02-16
Simplify trivial boolean expressions

LumioseSil:
There is no reason to manually compare with true or false when the boolean operators do that for us.

### DIFF
--- a/absl/debugging/internal/demangle.cc
+++ b/absl/debugging/internal/demangle.cc
@@ -914,7 +914,7 @@ static bool ParseSeqId(State *state) {
   if (guard.IsTooComplex()) return false;
   const char *p = RemainingInput(state);
   for (; *p != '\0'; ++p) {
-    if (!IsDigit(*p) && !(*p >= 'A' && *p <= 'Z')) {
+    if (!IsDigit(*p) && IsLower(*p)) {
       break;
     }
   }
--- a/absl/synchronization/blocking_counter.cc
+++ b/absl/synchronization/blocking_counter.cc
@@ -31,7 +31,7 @@ bool IsDone(void *arg) { return *reinterpret_cast<bool *>(arg); }
 BlockingCounter::BlockingCounter(int initial_count)
     : count_(initial_count),
       num_waiting_(0),
-      done_{initial_count == 0 ? true : false} {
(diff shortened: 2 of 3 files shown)


absl/debugging/internal/demangle.cc
```diff
@@ -896,7 +896,7 @@ static bool ParseFloatNumber(State *state) {
   if (guard.IsTooComplex()) return false;
   const char *p = RemainingInput(state);
   for (; *p != '\0'; ++p) {
-    if (!IsDigit(*p) && !(*p >= 'a' && *p <= 'f')) {
+    if (!IsDigit(*p) && (*p < 'a' || *p > 'f')) {
```
derekmauro: This seems like a style choice that doesn't fit with the description of this change ("There is no reason to manually compare with true or false when the boolean operators do that for us").

absl/debugging/internal/demangle.cc
```diff
@@ -914,7 +914,7 @@ static bool ParseSeqId(State *state) {
   if (guard.IsTooComplex()) return false;
   const char *p = RemainingInput(state);
   for (; *p != '\0'; ++p) {
-    if (!IsDigit(*p) && !(*p >= 'A' && *p <= 'Z')) {
+    if (!IsDigit(*p) && (*p < 'A' || *p > 'Z')) {
```
derekmauro: Same thing here. Maybe a better simplification would be to use a named helper like `IsUpper()`?

absl/debugging/internal/demangle.cc
```diff
@@ -914,7 +914,7 @@ static bool ParseSeqId(State *state) {
   if (guard.IsTooComplex()) return false;
   const char *p = RemainingInput(state);
   for (; *p != '\0'; ++p) {
-    if (!IsDigit(*p) && !(*p >= 'A' && *p <= 'Z')) {
+    if (!IsDigit(*p) && IsLower(*p)) {
```
derekmauro, line 917: This change is not equivalent to that the previous code was doing. Think about what would happen, for example, if the char was `'?'`.

derekmauro (changes requested):

PR #1395, opened 2023-02-16
Use const and static for member functions

LumioseSil:
This shows that these are member functions that do not modify a class's data.

### DIFF
--- a/absl/crc/internal/crc.cc
+++ b/absl/crc/internal/crc.cc
@@ -261,7 +261,7 @@ void CRC32::Extend(uint32_t* crc, const void* bytes, size_t length) const {
   const uint8_t* e = p + length;
   uint32_t l = *crc;
 
-  auto step_one_byte = [this, &p, &l] () {
+  auto step_one_byte = [this, &p, &l]() {
     int c = (l & 0xff) ^ *p++;
     l = this->table0_[c] ^ (l >> 8);
   };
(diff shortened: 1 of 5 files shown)


absl/strings/string_view.cc
```diff
@@ -64,7 +64,7 @@ std::ostream& operator<<(std::ostream& o, string_view piece) {
     size_t rpad = 0;
     if (static_cast<size_t>(o.width()) > piece.size()) {
       size_t pad = static_cast<size_t>(o.width()) - piece.size();
-      if ((o.flags() & o.adjustfield) == o.left) {
+      if ((o.flags() & std::ostream::adjustfield) == std::ostream::left) {
```
derekmauro: This change doesn't seem consistent with the rest of the changes in this change. I would have to ask that we avoid the churn for this one.

derekmauro (changes requested):

AtariDreams:
Addressed! @derekmauro

absl/crc/internal/crc32c_inline.h
```diff
@@ -28,7 +28,7 @@ namespace crc_internal {
 // CRC32C implementation optimized for small inputs.
 // Either computes crc and return true, or if there is
 // no hardware support does nothing and returns false.
-inline bool ExtendCrc32cInline(uint32_t* crc, const char* p, size_t n) {
+inline bool ExtendCrc32cInline(const uint32_t* crc, const char* p, size_t n) {
```
derekmauro: Please see line 56. This change doesn't build.

derekmauro (changes requested):

LumioseSil:
Fixed! @derekmauro

PR #1401, opened 2023-02-21
Allow later versions of Xcode to not have the apple_build_version workaround

LumioseSil:
Apple's clang fork has the missing commit now, and we can safely use the above codepath.

### DIFF
--- a/absl/hash/internal/hash.h
+++ b/absl/hash/internal/hash.h
@@ -1136,7 +1136,7 @@ class ABSL_DLL MixingHashState : public HashStateBase<MixingHashState> {
   // probably per-build and not per-process.
   ABSL_ATTRIBUTE_ALWAYS_INLINE static uint64_t Seed() {
 #if (!defined(__clang__) || __clang_major__ > 11) && \
-    !defined(__apple_build_version__)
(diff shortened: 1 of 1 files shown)


absl/hash/internal/hash.h
```diff
@@ -1136,7 +1136,7 @@ class ABSL_DLL MixingHashState : public HashStateBase<MixingHashState> {
   // probably per-build and not per-process.
   ABSL_ATTRIBUTE_ALWAYS_INLINE static uint64_t Seed() {
 #if (!defined(__clang__) || __clang_major__ > 11) && \
-    !defined(__apple_build_version__)
+    (!defined(__apple_build_version__) || __apple_build_version__ >= 19558921) // Xcode 12
```
derekmauro, line 1139: It looks like this changes the code from "only do this on Xcode >= 12" to "don't do this on any version of Xcode". Is that intended?

derekmauro (changes requested):

absl/hash/internal/hash.h
AZero13: Other way around

derekmauro: I must have been looking at the diff backwards. Thanks.

PR #1403, opened 2023-02-21
Resolve TODO: remove C++11 workarounds

LumioseSil:
We no longer support C++11 as per the Google C++ support documentation: we support C++14 and up, so we can remove these workarounds.

### DIFF
--- a/absl/time/duration.cc
+++ b/absl/time/duration.cc
@@ -96,13 +96,6 @@ inline bool IsValidDivisor(double d) {
   return d != 0.0;
 }
 
-// Can't use std::round() because it is only available in C++11.
-// Note that we ignore the possibility of floating-point over/underflow.
-template <typename Double>
-inline double Round(Double d) {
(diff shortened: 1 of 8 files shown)


absl/time/duration.cc
```diff
@@ -96,11 +96,9 @@ inline bool IsValidDivisor(double d) {
   return d != 0.0;
 }
 
-// Can't use std::round() because it is only available in C++11.
-// Note that we ignore the possibility of floating-point over/underflow.
 template <typename Double>
 inline double Round(Double d) {
-  return d < 0 ? std::ceil(d - 0.5) : std::floor(d + 0.5);
+  return std::round(d);
```
derekmauro: I don't think the comment is about using `std::round()` in the implementation of `Round()`. I think `Round()` should be replaced with `std::round()` now that `std::round()` is available.

derekmauro (changes requested):

LumioseSil:
@derekmauro It worked!

PR #1416, opened 2023-03-16
Prefer copy_n and fill_n over copy and fill where appropriate.

LumioseSil:
This lets us avoid having to do the addition manually.

### DIFF
--- a/absl/random/internal/nanobenchmark.cc
+++ b/absl/random/internal/nanobenchmark.cc
@@ -361,7 +361,7 @@ void CountingSort(T* values, size_t num_values) {
   // Write that many copies of each unique value to the array.
   T* ABSL_RANDOM_INTERNAL_RESTRICT p = values;
   for (const auto& value_count : unique) {
-    std::fill(p, p + value_count.second, value_count.first);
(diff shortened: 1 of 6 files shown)


absl/random/internal/randen_engine.h
```diff
@@ -105,7 +105,7 @@ class alignas(8) randen_engine {
     // Zeroes the inner state and fills the outer state with seed_value to
     // mimic the behaviour of reseed
     auto* begin = state();
-    std::fill(begin, begin + kCapacityT, 0);
+    std::fill_n(begin, kCapacityT, 0);
```
derekmauro: This one is less readable in my opinion because the next line has a `std::fill` on it that starts in the middle of the array. It's easier to read when they consistently do the same thing in the same way. I think these should either both be `std::fill` or `std::fill_n`, and I'd lean towards `std::fill`.

derekmauro (changes requested):

absl/random/internal/randen_engine.h
AZero13: Fixed!

LumioseSil:
@derekmauro Done!

PR #1421, opened 2023-03-29
Update string_view.h in case of std::

MBkkt:
Remove unnecessary includes.

I understand it can be difficult to update something like this but maybe at least as TODO


derekmauro:
I tried to merge this today, but there was way too much code in our internal codebase depending on the transitive includes. There is a proposal for a tool to fix includes at scale, but work on this has not started. I don't have an estimate for when it will be feasible to make this change.

MBkkt:
I understand, should I close PR?

PR #1426, opened 2023-04-03
Make flat_hash_map and flat_hash_set default constructors constexpr

btzy:
This allows one to declare constinit global flat_hash_map and flat_hash_set variables without fear of the static initialization order fiasco.

### DIFF
--- a/absl/container/flat_hash_map.h
+++ b/absl/container/flat_hash_map.h
@@ -156,7 +156,7 @@ class flat_hash_map : public absl::container_internal::raw_hash_map<
   //
   //   std::vector<std::pair<int, std::string>> v = {{1, "a"}, {2, "b"}};
   //   absl::flat_hash_map<int, std::string> map7(v.begin(), v.end());
-  flat_hash_map() {}
+  constexpr flat_hash_map() {}
   using Base::Base;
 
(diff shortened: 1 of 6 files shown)


derekmauro:
> This allows one to declare constinit global flat_hash_map and flat_hash_set variables without fear of the static initialization order fiasco.

It looks like at least some builds don't consider this to be constinit. That would have to be resolved first.

But I'm not so sure about the value making this change. Global variables of this type would still be [forbidden by the Google C++ Style Guide because they have non-trivial destructors](https://google.github.io/styleguide/cppguide.html#Static_and_Global_Variables). Initialization order is important, but so is destructor order. Further, `constexpr` is forever, in the sense that once it is added, removing it is a breaking change. This would constrain our implementation, arguably without any benefit if we still are not free to use the type as a global.

btzy:
That's a bit unfortunate.  My use case is some kind of global cache of objects, and destruction is safe because the cache _owns_ the objects.

But yeah, if we only consider code that adheres to the Google C++ Style Guide then this change wouldn't be useful.

PS: Is there any way to see the build log?  The link that is shown here goes to `http://fusion2/blah`, which I guess is only accessible within an internal network.

PR #1433, opened 2023-04-19
Fix incorrect timespec definition on 32-bit platforms with 64-bit time_t

olegartys:
Some 32-bit configurations may use 64-bit time_t, which leads to different layout of userspace timespec and the one expected by SYS_futex implementation in kernel. In particular the issue occurs when using musl libc which has switched to unconditional 64-bit time_t definition.

This patch introduces custom struct timespec with two longs when old SYS_futex is used to match the kernel timespec definition.

### DIFF
--- a/absl/synchronization/internal/futex.h
+++ b/absl/synchronization/internal/futex.h
@@ -32,6 +32,7 @@
 
 #include <atomic>
 #include <cstdint>
+#include <limits>
 
 #include "absl/base/optimization.h"
 #include "absl/synchronization/internal/kernel_timeout.h"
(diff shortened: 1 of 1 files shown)


absl/synchronization/internal/futex.h
```diff
@@ -79,6 +79,12 @@ namespace synchronization_internal {
 
 #if defined(SYS_futex_time64) && !defined(SYS_futex)
 #define SYS_futex SYS_futex_time64
+using futex_timespec = struct timespec;
+#else
+struct futex_timespec {
```
derekmauro, line 91: Is there a reason not to make this unconditional?

absl/synchronization/internal/futex.h
```diff
@@ -128,6 +138,14 @@ class FutexImpl {
     }
     return 0;
   }
+
+ private:
+  static futex_timespec* FutexTimespec(futex_timespec* fts, const struct timespec* ts) {
+    if (!ts) return nullptr;
+    *fts = {static_cast<decltype(fts->tv_sec)>(ts->tv_sec),
```
derekmauro: Do we need to cap the value if it would overflow long?

absl/synchronization/internal/futex.h
derekmauro: I think I understand why this is done now, but code comments would be helpful for future readers.

absl/synchronization/internal/futex.h
olegartys: Thanks for the comment. Actually, I don't know if those checks are necessary, since overflow may happen only due to the year 2038 problem and SYS_futex which uses old 32-bit userspace should not be used by the time

Nevertheless, I have added overflow value capping, please check

absl/synchronization/internal/futex.h
olegartys: Please check the comment in the new patchset

PR #1448, opened 2023-05-17
Use UB impl of launder on when using clang < 8 and c++17

### LINKED ISSUE #1309: std::launder was not implemented in clang until version 6

glandium:
libstdc++'s definition of std::launder places it behind a check for __builtin_launder, which is not available before clang 8.

Fixes: #1309

### DIFF
--- a/absl/functional/internal/any_invocable.h
+++ b/absl/functional/internal/any_invocable.h
@@ -197,7 +197,7 @@ union TypeErasedState {
 template <class T>
 T& ObjectInLocalStorage(TypeErasedState* const state) {
   // We launder here because the storage may be reused with the same type.
-#if ABSL_INTERNAL_CPLUSPLUS_LANG >= 201703L
+#if ABSL_INTERNAL_CPLUSPLUS_LANG >= 201703L && __cpp_lib_launder >= 201606L
   return *std::launder(reinterpret_cast<T*>(&state->storage));
 #elif ABSL_HAVE_BUILTIN(__builtin_launder)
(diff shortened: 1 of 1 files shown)


absl/functional/internal/any_invocable.h
```diff
@@ -197,7 +197,8 @@ union TypeErasedState {
 template <class T>
 T& ObjectInLocalStorage(TypeErasedState* const state) {
   // We launder here because the storage may be reused with the same type.
-#if ABSL_INTERNAL_CPLUSPLUS_LANG >= 201703L
+#if ABSL_INTERNAL_CPLUSPLUS_LANG >= 201703L && \
```
derekmauro: The proposed patch works in my testing, but I think I more principled thing to do would be to use `#if __cpp_lib_launder >= 201606L`. This seems to produce the same result when I try it on godbolt.

https://godbolt.org/z/sGezba1b9

derekmauro (changes requested):

PR #1449, opened 2023-05-18
Update time_zone_lookup.cc

wony-zheng:
use window timezone api to read localtimezone
no need to set enviorment variable


derekmauro:
Changes to this file need to be sent to https://github.com/google/cctz. We automatically import changes from the CCTZ project.

PR #1459, opened 2023-05-24
Use -fp-model=precise for IntelLLVM builds

niranjan-nilakantan:
IntelLLVM uses -fp-model=fast instead of -fp-model=precise by default.

See https://www.intel.com/content/www/us/en/developer/articles/guide/porting-guide-for-icc-users-to-dpcpp-or-icx.html#inpage-nav-2

To handle IsInf and IsNan tests correctly, we override the default copts in both the Windows and Linux builds.


absl/copts/copts.py
```diff
@@ -83,6 +83,18 @@
     "-DNOMINMAX",
 ]
 
+ABSL_INTEL_LLVM_ADDITIONAL_FLAGS = [
+    # IntelLLVM uses fp-model=fast by default for Release builds,
+    # which breaks NaN and Inf handling.
+    "-ffp-model=precise",
```
derekmauro, line 89: This flag worries me. In general the flags in this file are about warning options, not options that affect generated code. Options that affect generated code need to be applied on a whole program basis, otherwise we risk running afoul of the [One Definition Rule](https://en.cppreference.com/w/cpp/language/definition). These options are only used on individual translation units in the Abseil library.

I think using this option here only silences a problem in tests.

IMO IntelLLVM has not picked a safe default.

niranjan-nilakantan: Good point - I treated this as a "per-file" flag, which might not be correct.  I will remove this PR.

To suppress the warnings for callers, I have sent this PR - https://github.com/abseil/abseil-cpp/pull/1464

Would this be a good middle ground for callers that want to use fast-math e.g. CUDA or oneAPI ?

Some tests will still fail with fp-model=fast , and we leave that as is.

PR #1464, opened 2023-06-02
Do not check for inf or nan if  __FAST_MATH__==1

niranjan-nilakantan:
IntelLLVM uses `-ffp-model=fast` by default,

which indirectly defines __FAST_MATH__=1

See https://clang.llvm.org/docs/UsersManual.html#cmdoption-ffast-math

We make the changes only in the non-test code, so that no warnings are generated
for calling projects.

Strange things will happen if the callers use fast-math incorrectly,
and actually create inf/nan values
e.g. some of the absl tests fail with fast-math.

### DIFF
--- a/absl/flags/marshalling.cc
+++ b/absl/flags/marshalling.cc
@@ -177,7 +177,9 @@ std::string UnparseFloatingPointVal(T v) {
   // conversions, but may not be enough to represent all the values correctly.
   std::string digit10_str =
       absl::StrFormat("%.*g", std::numeric_limits<T>::digits10, v);
+#if !defined(__FAST_MATH__) || !__FAST_MATH__
   if (std::isnan(v) || std::isinf(v)) return digit10_str;
+#endif
 
   T roundtrip_val = 0;
   std::string err;
--- a/absl/random/beta_distribution.h
+++ b/absl/random/beta_distribution.h
@@ -89,10 +89,12 @@ class beta_distribution {
         method_ = JOEHNK;
         a_ = result_type(1) / alpha_;
         b_ = result_type(1) / beta_;
+#if !defined(__FAST_MATH__) || !__FAST_MATH__
         if (std::isinf(a_) || std::isinf(b_)) {
           method_ = DEGENERATE_SMALL;
           x_ = inverted_ ? result_type(1) : result_type(0);
         }
+#endif
         return;
       }
       if (a_ >= ThresholdForLargeA()) {
--- a/absl/strings/internal/str_format/float_conversion.cc
+++ b/absl/strings/internal/str_format/float_conversion.cc
@@ -1040,7 +1040,7 @@ bool ConvertNonNumericFloats(char sign_char, Float v,
   char text[4], *ptr = text;
   if (sign_char != '\0') *ptr++ = sign_char;
 #if defined(__FAST_MATH__) && __FAST_MATH__
-  (void) v;
+  (void)v;
   (void)conv;
   (void)sink;
   return false;
--- a/absl/strings/internal/str_format/float_conversion.cc
+++ b/absl/strings/internal/str_format/float_conversion.cc
@@ -1039,6 +1039,12 @@ bool ConvertNonNumericFloats(char sign_char, Float v,
                              FormatSinkImpl *sink) {
   char text[4], *ptr = text;
   if (sign_char != '\0') *ptr++ = sign_char;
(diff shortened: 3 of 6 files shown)


derekmauro (commented):
I'm not sure I understand what problem this change is trying to solve. It seems to be trying to suppress a correct warning. Why do you want to do that? Is this change trying to solve some other problem as well?

niranjan-nilakantan:
> I'm not sure I understand what problem this change is trying to solve. It seems to be trying to suppress a correct warning. Why do you want to do that? Is this change trying to solve some other problem as well?


I include absl "from head" in my containing CMAKE project like this.

```
include(FetchContent)

function(include_third_party)
  FetchContent_Declare(
    googletest
    GIT_REPOSITORY https://github.com/google/googletest.git
    GIT_TAG        main
  )

  FetchContent_Declare(
    abseil-cpp
    GIT_REPOSITORY https://github.com/abseil/abseil-cpp.git
    GIT_TAG        master
  )

  FetchContent_MakeAvailable(googletest abseil-cpp)
endfunction()

include_third_party()

# my targets follow here.
```

When I build my project, absl is built indirectly, and I get the following warnings.  I am looking for aprroaches that will get me to 0 warnings when I build my project.

```
[build] [217/254  83% :: 3.909] Building CXX object _deps/abseil-cpp-build/absl/time/CMakeFiles/absl_time.dir/Release/duration.cc.o
[build] .../build/_deps/abseil-cpp-src/absl/time/duration.cc:89:7: warning: explicit comparison with NaN in fast floating point mode [-Wtautological-constant-compare]
[build]   if (std::isnan(d)) return false;
[build]       ^~~~~~~~~~~~~
...
[build] 4 warnings generated.
```

As part of my earlier change, I was looking into either

* Allow absl to control its behavior independent of the caller (i.e. always use -ffp-model=precise, almost all absl tests pass)
* or be aware of the limitations of the calling code under certain global flags (i.e. -ffp-model=fast), and handle it appropriately (-this PR is a partial solution to that)

Let me know if you have seen any other alternatives.

niranjan-nilakantan:
I also tried adding a modified version of the following (to try a local override) in the **CMAKE function** above.

```
set (CMAKE_CXX_FLAGS -Wno-tautological-constant-compare)
```

But it did not work because the CMAKE_CXX_FLAGS are used **before** the ABSL_..._FLAGS,
and so **-Wno-tautological-constant-compare** is overridden by the later **-Wall**

derekmauro:
The IntelLLVM compiler is not currently on our supported compiler list. As as explained before, if you want this to work, the real fix is to compile the entire program with `-ffp-model=precise`. Compiling Abseil only with that flag will introduce ODR violations.

I find that most people trying to suppress warnings want to compile their program with `-Werror` or equivalent (treat all warnings as errors). I think it is a great idea to compile your own code with `-Werror`, but compiling third party code with `-Werror` going to be a never-ending battle. Most compilers offer a way to exclude warnings-as-errors from headers included from third party code. Our CMake config has `ABSL_USE_SYSTEM_INCLUDES` for this purpose.

This pull request takes code that is correct when built with safe code generation flags, and makes it incorrect by removing code, in an attempt to silence a warning that would actually help someone figure out why the code is not correct when using unsafe code generation flags. This seems like a bad idea to me.

niranjan-nilakantan:
Thanks for the pointer to the SYSTEM keyword - this appears to work for suppressing warnings from absl headers that I include, but not while **building absl** when I use FetchContent in CMake (unless I missed something in my experiment).

I understand the preference for a single choice of fp-model=precise.

To handle the constraints that I am working with (some necessary optimizations are triggered only with fast floating point enabled etc.), I will add the suppressions at my end, with the attendant risks, to leverage the advantages that absl provides.
If this turns out to be invalid, I will look at other options.

e.g. something like this in my CMakeLists.txt

```
  # TODO: Check the current fp-model flags to see if a suppression is needed.

  # Disable warnings for just one file in the target.
  # set_source_files_properties(${absl_SOURCE_DIR}/absl/flags/marshalling.cc
  #     TARGET_DIRECTORY absl_flags_marshalling
  #     PROPERTIES COMPILE_OPTIONS "-Wno-tautological-constant-compare")

  # Or, just disable the warnings for the entire target.
  # target_compile_options(
  #     absl_flags_marshalling PRIVATE
  #         -Wno-tautological-constant-compare
  # )
```

Thanks for looking into this.

Issue #1251, opened 2022-08-12
Deprecation warnings with Visual Studio 17.3


Sovak:
**Describe the bug**

Upon upgrading the Visual Studio to 17.3 I get the following warnings:

warning C4996: '_Header_ciso646': warning STL4036: <ciso646> is removed in C++20. You can define _SILENCE_CXX20_CISO646_REMOVED_WARNING or _SILENCE_ALL_CXX20_DEPRECATION_WARNINGS to acknowledge that you have received this warning.

abseil-cpp/absl/meta/type_traits.h(125,50): warning C4996: 'std::aligned_storage<_Len,_Align>': warning STL4034: std::aligned_storage and std::aligned_storage_t are deprecated in C++23. Prefer alignas(T) std::byte t_buff[sizeof(T)]. You can define _SILENCE_CXX23_ALIGNED_STORAGE_DEPRECATION_WARNING or _SILENCE_ALL_CXX23_DEPRECATION_WARNINGS to acknowledge that you have received this warning.
60>C:\Program Files\Microsoft Visual Studio\2022\Community\VC\Tools\MSVC\14.33.31629\include\type_traits(1070,1): message : see declaration of 'std::aligned_storage'

**Steps to reproduce the bug**

Build abseil with latest MSVC

**What version of Abseil are you using?**
0c92330442d6b1be934e0407115c8084250ef347

**What operating system and version are you using**

Latest Windows 10

**What compiler and version are you using?**

MSVC 14.33.31629 (Visual Studio 17.3)

**What build system are you using?**

MSBuild with CMake 3.21.4

derekmauro:
Fixed by 3e54e87f020c71440217ce2ad382b800c4658970 yesterday.

derekmauro:
Actually the first warning was fixed. The second needs a fix.

Sil3ntStorm:
The ciso646 include seems to be still present in release 20220623.1, when can we expect a release that includes this fix?

derekmauro:
We do releases approximately every 6 months.

Romain-Geissler-1A:
Hi,

FYI, the second warning (about deprecated `std::aligned_storage`) is still present in the current release + the current development branches. Is it planned to at least silence it (with a pragma) if you don't immediately plan on re-writing this part of the code to not use this deprecated class ?

Edit: I have submitted a pull request in #1468 which works this around for gcc/clang (but not MSVC as I clearly now nothing about the Windows platform).

Cheers,
Romain

PR #1468, opened 2023-06-05
Silence std::aligned_storage warnings in C++23 mode.

Romain-Geissler-1A:
This is only a workaround, working only with gcc/clang, while waiting for a better long term fix not using std::aligned_storage, which is tracked by b/260219225.

Note: I don't know how to silence warnings with MSCV, so I let Windows users submitting the same fix for their platform.

This warning was already reported in #1251.

### DIFF
--- a/absl/meta/type_traits.h
+++ b/absl/meta/type_traits.h
@@ -40,6 +40,7 @@
 #include <type_traits>
 
 #include "absl/base/config.h"
+#include "absl/base/attributes.h"
 
 // Defines the default alignment. `__STDCPP_DEFAULT_NEW_ALIGNMENT__` is a C++17
 // feature.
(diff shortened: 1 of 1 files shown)


absl/meta/type_traits.h
```diff
@@ -278,6 +278,10 @@ using remove_extent_t = typename std::remove_extent<T>::type;
 template <typename T>
 using remove_all_extents_t = typename std::remove_all_extents<T>::type;
 
+#ifdef __GNUC__
```
derekmauro: We have a macro to do this correctly that should be used here:
https://github.com/abseil/abseil-cpp/blob/66406fdf1553de6ad672576167e7cc7ca2d764cb/absl/base/attributes.h#L688-L708

I *think* the `__clang__` case is for `clang-cl`.

derekmauro (changes requested):

absl/meta/type_traits.h
Romain-Geissler-1A: Ok I have submitted an updated version where I tried to use these.

PR #1472, opened 2023-06-09
Use InlinedVector

pateldeev:
It accomplishes the same thing and is cleaner in terms of code. It will also take up less inlined space.


derekmauro:
That's a nice simplification. Merging.

PR #1481, opened 2023-06-17
Add hash function for array smart pointers

MrJia1997:
Add hash function for array smart pointers. This is essential for putting them into associative containers like flat_hash_[set,map] or node_hash_[set,map].

### DIFF
--- a/absl/container/internal/hash_function_defaults_test.cc
+++ b/absl/container/internal/hash_function_defaults_test.cc
@@ -280,11 +280,13 @@ struct NoDeleter {
   void operator()(const T* ptr) const {}
 };
 
-using PointerTypes =
-    ::testing::Types<const int*, int*, std::unique_ptr<const int>,
-                     std::unique_ptr<const int, NoDeleter>,
-                     std::unique_ptr<int>, std::unique_ptr<int, NoDeleter>,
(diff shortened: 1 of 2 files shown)


MrJia1997:
I tried clicking on the "Details" link but it seemed like it couldn't be resolved. How can I know the detailed reason why the builds failed?

fsorry:
Yuanhao Jia ***@***.***> schrieb am Mo., 26. Juni 2023, 16:32:

absl/container/internal/hash_function_defaults_test.cc
```diff
@@ -309,17 +311,33 @@ TYPED_TEST(EqPointer, Works) {
   const int* cptr = ptr;
   std::unique_ptr<int, NoDeleter> uptr(ptr);
   std::unique_ptr<const int, NoDeleter> cuptr(ptr);
+  auto saptr = std::shared_ptr<int[]>(new int[1]);
```
derekmauro, line 314: As far as I can tell, this isn't legal in C++14, even though most compiler/library combinations accept it. libc++ seems to reject it.

https://godbolt.org/z/qK61vPj65

I think it has something to do with `new int[1]` decaying to `int*`, which makes the library think the deleter is mismatched.

PR #1496, opened 2023-07-28
Update ABSL_ATTRIBUTE_NO_SANITIZE_ADDRESS condition

johnnkp:
This PR use [`__SANITIZE_ADDRESS__`](https://learn.microsoft.com/en-us/cpp/sanitizers/asan-building?view=msvc-170#__sanitize_address__) to check if address sanitizer is enabled. It fixes an error when address sanitizer is installed but not enabled during compilation.

### DIFF
--- a/absl/base/attributes.h
+++ b/absl/base/attributes.h
@@ -215,7 +215,7 @@
     ABSL_HAVE_ATTRIBUTE(no_sanitize_address)
 #define ABSL_ATTRIBUTE_NO_SANITIZE_ADDRESS __attribute__((no_sanitize_address))
 #elif defined(ABSL_HAVE_ADDRESS_SANITIZER) && defined(_MSC_VER) && \
-    _MSC_VER >= 1928
+    defined(__SANITIZE_ADDRESS__)
 // https://docs.microsoft.com/en-us/cpp/cpp/no-sanitize-address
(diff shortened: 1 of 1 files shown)


absl/base/attributes.h
```diff
@@ -215,7 +215,7 @@
     ABSL_HAVE_ATTRIBUTE(no_sanitize_address)
 #define ABSL_ATTRIBUTE_NO_SANITIZE_ADDRESS __attribute__((no_sanitize_address))
 #elif defined(ABSL_HAVE_ADDRESS_SANITIZER) && defined(_MSC_VER) && \
-    _MSC_VER >= 1928
+    defined(__SANITIZE_ADDRESS__)
```
derekmauro, line 218: Please help me understand why `__SANITIZE_ADDRESS__` is needed here. It looks like `ABSL_HAVE_ADDRESS_SANITIZER` is already guarded by `__SANITIZE_ADDRESS__`.

https://github.com/abseil/abseil-cpp/blob/c9c0fd5185067208e1cc4227d16122571762af39/absl/base/config.h#L745-L746

johnnkp: Oops, thanks for pointing this out. What I saw is an outdated copy from tensorflow. I will report this issue to them then.

PR #1520, opened 2023-08-19
Make use of C++20 feature test macros to bestow better compatibility

### LINKED ISSUE #1519: [Bug]: std type detection broken on MacOS when using vendored STLs
### Describe the issue

Using MacOS version based switch to determine ABSL_INTERNAL_APPLE_CXX17_TYPES_UNAVAILABLE breaks building abseil-cpp against vendored STL's.

Read this text

20230802.0

I am targeting iOS 11.0, but I have vendored STL which has necessary symbols

clang 14.

Custom internal build system

### Additional context

С++20 introduces feature testing macros, which provide more steady solution for checking feature availability
https://en.cppreference.com/w/cpp/feature_test

georgthegreat:
Fixes #1519 and removes the burden of manually feature testing all the platforms supported by absl.

### DIFF
--- a/absl/base/config.h
+++ b/absl/base/config.h
@@ -557,6 +557,8 @@ static_assert(ABSL_INTERNAL_INLINE_NAMESPACE_STR[0] != 'h' ||
 // Checks whether C++17 std::any is available.
 #ifdef ABSL_HAVE_STD_ANY
 #error "ABSL_HAVE_STD_ANY cannot be directly set."
+#elif defined(__cpp_lib_any)
+#define ABSL_HAVE_STD_ANY 1
 #elif defined(ABSL_INTERNAL_CPLUSPLUS_LANG) && \
     ABSL_INTERNAL_CPLUSPLUS_LANG >= 201703L && \
     !ABSL_INTERNAL_APPLE_CXX17_TYPES_UNAVAILABLE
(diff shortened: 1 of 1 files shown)


georgthegreat:
@derekmauro, how can I cherry pick this to 20230802.0 branch?

derekmauro:
I'm collecting issues for a patch release in about 2 weeks. I can include this one.

georgthegreat:
It would be nice to have this one include, as it is the only patch we apply to abseil-cpp locally.
Cutting the number of patches to zero would be nice.

derekmauro:
It looks like this cannot actually be included in a patch release as it is potentially a significant change to which type is used, and iOS 11 is technically unsupported.

georgthegreat:
@derekmauro, prior to 20230802 #ifdef were quite correct, see [20230125.3 for example](https://github.com/abseil/abseil-cpp/blob/20230125.3/absl/base/config.h#L629).

So it looks more like a degradation that should be (hot)fixed.

Prior to this update we had a different patch that was just checking for `std::is_same_v<absl::string_view, std::string_view>`.

georgthegreat:
@derekmauro, could you, please, comment on the above?

derekmauro:
Sorry, but this will not be included in a patch release, for the reason stated above.

PR #1529, opened 2023-09-08
Adds ABSL_PRETTY_FUNCTION support for clang.

jiawen:
[Clang has had this feature](https://clang.llvm.org/docs/LanguageExtensions.html#source-location-builtins) for quite some time now.


derekmauro (commented):
I'm pretty sure `clang` defines `__GNUC__`  and I think it even defines `_MSC_VER` when using `clang-cl`. Do we actually need this? Does it fix some problem? Was it tested?

derekmauro:
Closing due to no response from the author of this pull request.

PR #1534, opened 2023-09-18
Abseil LTS branch, Aug 2023, Patch 1

derekmauro:
This cherry-picks changes to fix certain builds on FreeBSD, NetBSD, OpenBSD, and MinGW
(and possibly other platforms).

---
Add StdcppWaiter to the end of the list of waiter implementations

Since ABSL_INTERNAL_HAVE_STDCPP_WAITER is defined on all systems
it is effectively a fallback. I left the condition there in case
we have to disable it on some platform in the future.

---
Use native methods to implement absl::base_internal::GetPID() on 
FreeBSD, NetBSD, and OpenBSD

https://man.freebsd.org/cgi/man.cgi?query=pthread_getthreadid_np
https://man.netbsd.org/_lwp_self.2
https://man.openbsd.org/getthrid.2


b7r3333:
جواهر بنت عيادة قحبة بنت حرام

PR #1536, opened 2023-09-23
Avoid export internal target in non-test builds

MiguelBarro:
The target is test_allocator associated with the containers


Bobjbayli (approved):

derekmauro:
We need the CLA to accept this.

PR #1550, opened 2023-10-20
Replace rdcycle instruction with rdtime

marv:
The `rdcycle` instruction causes SIGILL with RISCV_PMU_SBI=y in the kernel[1]

According to a PR for the highway project[2] which fixed the same problem there `rdtime` is the proper equivalent for the x86 `rdtsc` instruction on RISC-V

Using `rdtime` instead of `rdcycle` makes all tests pass on my VisionFive 2 board

[1] https://lists.infradead.org/pipermail/linux-riscv/2022-September/018862.html
[2] https://github.com/google/highway/pull/961

### DIFF
--- a/absl/base/internal/unscaledcycleclock.cc
+++ b/absl/base/internal/unscaledcycleclock.cc
@@ -125,7 +125,7 @@ double UnscaledCycleClock::Frequency() {
 
 int64_t UnscaledCycleClock::Now() {
   int64_t virtual_timer_value;
-  asm volatile("rdcycle %0" : "=r"(virtual_timer_value));
+  asm volatile("rdtime %0" : "=r"(virtual_timer_value));
   return virtual_timer_value;
 }
 


derekmauro:
@bbarenblat had proposed a similar (identical?) change internally a few months ago. I pointed out that it is a little bit more complicated because `ABSL_INTERNAL_UNSCALED_CYCLECLOCK_FREQUENCY_IS_CPU_FREQUENCY` would also need to change.

ghost (approved):

aurel32:
Removing the `ABSL_INTERNAL_UNSCALED_CYCLECLOCK_FREQUENCY_IS_CPU_FREQUENCY` define for RISC-V seems easy to do. But then it seems we have to also update `UnscaledCycleClock::Frequency`. The frequency of the `rdtime` timer varies from board to board, and TTBOMK there is no easy way to get its frequency. In such condition, is it fine changing `UnscaledCycleClock::Frequency` to return `1.0`, like it is done for the CPU when it is not possible to determine the frequency?

aurel32:
> Removing the `ABSL_INTERNAL_UNSCALED_CYCLECLOCK_FREQUENCY_IS_CPU_FREQUENCY` define for RISC-V seems easy to do. But then it seems we have to also update `UnscaledCycleClock::Frequency`. The frequency of the `rdtime` timer varies from board to board, and TTBOMK there is no easy way to get its frequency. In such condition, is it fine changing `UnscaledCycleClock::Frequency` to return `1.0`, like it is done for the CPU when it is not possible to determine the frequency?

Given I got no answer I gave it a tried, and I have found it is not necessary to provide `UnscaledCycleClock::Frequency`. I have therefore opened a new PR: https://github.com/abseil/abseil-cpp/pull/1631

derekmauro:
This was addressed by #1644.

PR #1560, opened 2023-11-02
Provide absl::Hash for std::filesystem::path

### LINKED ISSUE #655: AbslHashValue for std::filesystem::path
There are AbslHashValue function templates for many std types, but not for std::filesystem::path.

georgthegreat:
It turned out to be quite hard to find a way to provide AbslHashValue for a class declared inside `namespace std`.

I am using C++20 feature testing macro to detect if std::filesystem is available.

Fixes #655.

### DIFF
--- a/absl/hash/hash_test.cc
+++ b/absl/hash/hash_test.cc
@@ -21,6 +21,7 @@
 #include <cstdint>
 #include <cstdlib>
 #include <cstring>
+#include <filesystem>
 #include <functional>
 #include <initializer_list>
 #include <ios>
@@ -34,6 +35,7 @@
 #include <unordered_map>
 #include <utility>
 #include <vector>
+#include <version>
 
 #include "gtest/gtest.h"
 #include "absl/base/config.h"
@@ -477,6 +479,19 @@ TEST(HashValueTest, U32StringView) {
                       std::u32string_view(U"ABC"),
                       std::u32string_view(U"Some other different string_view"),
                       std::u32string_view(U"Iñtërnâtiônàlizætiøn"))));
+#endif
+}
+
+TEST(HashValueTest, StdFilesystemPath) {
+#ifndef __cpp_lib_filesystem
+  GTEST_SKIP()
+#else
+  namespace fs = std::filesystem;
+  EXPECT_TRUE((is_hashable<std::filesystem::path>::value));
+
+  EXPECT_TRUE(absl::VerifyTypeImplementsAbslHashCorrectly(
+      std::make_tuple(fs::path(), fs::path("a/b/c"),
+                      fs::path(u8"Iñ/tër/nât/iôn/àli/zæt/iøn"))));
 #endif
 }
 
--- a/absl/hash/internal/hash.h
+++ b/absl/hash/internal/hash.h
@@ -42,6 +42,10 @@
 #include <utility>
 #include <vector>
 
+#if defined(__cpp_lib_filesystem)
+#include <filesystem>
+#endif
+
 #include "absl/base/config.h"
 #include "absl/base/internal/unaligned_access.h"
 #include "absl/base/port.h"
@@ -563,6 +567,15 @@ H AbslHashValue(H hash_state, std::basic_string_view<Char> str) {
 }
 
 #endif  // ABSL_HAVE_STD_STRING_VIEW
+
+#if defined(__cpp_lib_filesystem)
+
+template <typename H>
+H AbslHashValue(H hash_state, const std::filesystem::path& path) {
+    return AbslHashValue(std::move(hash_state), path.native());
+}
+
+#endif
 
 // -----------------------------------------------------------------------------
 // AbslHashValue for Sequence Containers
--- a/absl/hash/internal/hash.h
+++ b/absl/hash/internal/hash.h
@@ -41,6 +41,7 @@
 #include <unordered_set>
 #include <utility>
 #include <vector>
+#include <version>
 
 #if defined(__cpp_lib_filesystem)
 #include <filesystem>
@@ -572,7 +573,10 @@ H AbslHashValue(H hash_state, std::basic_string_view<Char> str) {
 
 template <typename H>
 H AbslHashValue(H hash_state, const std::filesystem::path& path) {
-    return AbslHashValue(std::move(hash_state), path.native());
+  using PathView = std::basic_string_view<std::filesystem::path::value_type>;
+  PathView view = path.native();
+  return H::combine(
+      H::combine_contiguous(std::move(hash_state), view.data(), view.size()));
 }
 
 #endif


derekmauro (changes requested):
Please run the tests (`bazel test --cxxopt=-std=c++17 absl/hash:hash_test`). This appears to be breaking HashValueTest.Strings. Perhaps it is allowing an unexpected conversion?

This also needs a test of its own.

georgthegreat:
Hmm... I am unable to reproduce the problems neither with `-std=c++17` nor with `-std=c++2a`.
I have installed bazel (via bazelisk) and running it as follows:

```
(dflt) thegeorg@jakku:~/contrib/abseil-cpp@patch-1$ ~/bin/bazel test --cxxopt=-std=c++2a absl/hash:hash_test
INFO: Analyzed target //absl/hash:hash_test (0 packages loaded, 0 targets configured).
INFO: Found 1 test target...
Target //absl/hash:hash_test up-to-date:
  bazel-bin/absl/hash/hash_test
INFO: Elapsed time: 0.259s, Critical Path: 0.00s
INFO: 1 process: 1 internal.
INFO: Build completed successfully, 1 total action
//absl/hash:hash_test                                           (cached) PASSED in 0.1s

Executed 0 out of 1 test: 1 test passes.
There were tests whose specified size is too big. Use the --test_verbose_timeout_warnings command line option to see which ones these are.
```

Are there any requirements like compiler version / libc++ / libstdc++ version?
I am compiling on Linux using gcc-9.3.0 against libstdc++.

georgthegreat:
UPD: I was able to make the code compile, yet the test fails with 

```
absl/hash/hash_test.cc:492: Failure
Value of: absl::VerifyTypeImplementsAbslHashCorrectly( std::make_tuple(fs::path(), fs::path("a/b/c"), fs::path(u8"Iñ/tër/nât/iôn/àli/zæ
t/iøn")))
  Actual: false (Hash expansion of #0("") is a suffix of the hash expansion of #1("a/b/c").)
Expected: true
```

As I barely delegate fs::path hashing to std::string_view one, this hardly should be a problem.
@derekmauro, could you help me with interpreting this output?

absl/hash/internal/hash.h
```diff
@@ -564,6 +569,18 @@ H AbslHashValue(H hash_state, std::basic_string_view<Char> str) {
 
 #endif  // ABSL_HAVE_STD_STRING_VIEW
 
+#if defined(__cpp_lib_filesystem)
+
+template <typename H>
+H AbslHashValue(H hash_state, const std::filesystem::path& path) {
+  using PathView = std::basic_string_view<std::filesystem::path::value_type>;
+  PathView view = path.native();
+  return H::combine(
```
georgthegreat, line 578: It would be nice to simply delegate this call to `AbslHashValue(hash_state, view)` but delegation is explicitly forbidden in runtime.

Hence the copy-paste.

ghost (approved):
> #1560

absl/hash/hash_test.cc
ghost, line 1: - absl/hash/hash_test.cc -

ghost (approved):

derekmauro:
This is proving quite hard to get right for a number of reasons.

In some cases `H AbslHashValue(H hash_state, const std::filesystem::path& path)` is allowing implicit conversions to `path` that are taking precedence over the expected overload. This can be resolved be defining it as

```
template <typename Path, typename H,
          typename = absl::enable_if_t<
              std::is_same<Path, std::filesystem::path>::value>>
H AbslHashValue(H hash_state, const Path& path)
```

Now how do we implement this function? The approach in this PR does not work because it is required that when the items compare equal, the hashes must be equal as well. The standard library does not implement `==` as a simple string comparison, it seems to do some normalization.

I tried a few things including calling `lexically_normal()` and iterating though the path components, but everything I've tried is broken in some way.

The closest thing I've found to getting this right is relying on the standard library to get the hash right using `std::filesystem::hash_value()`
```
template <typename Path, typename H,
          typename = absl::enable_if_t<
              std::is_same<Path, std::filesystem::path>::value>>
H AbslHashValue(H hash_state, const Path& path) {
  return H::combine(std::move(hash_state), std::filesystem::hash_value(path));
}
```
but even this is buggy.

For example, libstdc++ fails on `//`.
https://gcc.gnu.org/bugzilla/show_bug.cgi?id=106452
https://godbolt.org/z/YK1fbdvhc

I've also found that Clang for Windows fails on `c:\` vs `c:/`.

Here is the unit test I'm working with that demonstrates the problems:
```
TEST(HashValueTest, StdFilesystemPath) {
#if ABSL_INTERNAL_CPLUSPLUS_LANG < 201703L
  GTEST_SKIP() << "std::filesystem::path requires C++17";
#else
  EXPECT_TRUE((is_hashable<std::filesystem::path>::value));

  // clang-format off
  const auto kTestCases = std::make_tuple(
      std::filesystem::path(),
      std::filesystem::path("/"),
      std::filesystem::path("//"),
      std::filesystem::path("/a/b"),
      std::filesystem::path("/a//b"),
      std::filesystem::path("a/b"),
      std::filesystem::path("a/b/"),
      std::filesystem::path("a//b"),
      std::filesystem::path("a//b/"),
      std::filesystem::path("c:/"),
      std::filesystem::path("c:\\"),
      std::filesystem::path("c://"),
      std::filesystem::path("c:\\/"),
      std::filesystem::path("c://\\"),
      std::filesystem::path("c:\\//"),
      std::filesystem::path("/e/p"),
      std::filesystem::path("/s/../e/p"),
      std::filesystem::path("e/p"),
      std::filesystem::path("s/../e/p"));
  // clang-format on

  EXPECT_TRUE(absl::VerifyTypeImplementsAbslHashCorrectly(kTestCases));
#endif
}
```

wseife01:
Unsubscribe

georgthegreat:
It looks like the best option we have will be to rely onto `std::filesystem::hash_value()`.

It's pretty weird that `"C:\\Windows"` split is `["C:", "\\", "Windows"]` (i. e. not `["C:\\", "Windows"]`) (see cppreference [example](https://en.cppreference.com/w/cpp/filesystem/path/begin)). I believe that libc++ bug is caused by this.

@derekmauro, would you accept such hash implementation even if it does not work in certain STL / Compiler / OS combinations?
I will comment / ifdef out certain test cases to keep the CI green.

derekmauro:
I'll handle this. I would like to setup a `clang-cl` with C++17 build first (should be done this week) so we have coverage for this. Currently we only have C++14 with `clang-cl`.

derekmauro:
This was rolled back due to `std::filesystem::path` being marked unavailable on iOS 13. It will be rolled forward with a fix soon.

georgthegreat:
Maybe use can check `ifdef  __cpp_lib_filesystem` from C++20 feature testing macros instead of duplicating OS / STL version checks instead.

I did the same in #1520 with std::string_view

PR #1591, opened 2024-01-03
memutil.cc: always use tolower() in memcasecmp() for better branch prediction

IAKOBVS:


derekmauro:
We need the CLA before we can consider contributions.

This is essentially proposing reverting a recent change that claimed to show an improvement:
https://github.com/abseil/abseil-cpp/commit/6c8338c252116e429307361ff4fdc1fd0532902d

We would need real data to convince us to revert https://github.com/abseil/abseil-cpp/commit/6c8338c252116e429307361ff4fdc1fd0532902d.

I also highly recommend https://abseil.io/fast/39, particularly the part about how small microbenchmarks can mislead about actual performance in large programs which cache pressure is high. That is the most important thing to consider here. Macrobenchmarks in large programs are what really matter.

PR #1600, opened 2024-01-17
Prevent redefinition errors on CHECK et al macros

elfringham:
Prevent errors and warnings on redefinition of these macros when abseil is included in a project that also includes another Google project eg tsl that also defines the same macros.


derekmauro:
We already provide `absl/log/absl_check.h` and `absl/log/absl_log.h`. Those that have this problem should be using those.

PR #928, opened 2021-03-31
Silence GCC warning about non-virtual-dtor of base class

KSSysLab:
Co-authored-by: Zhiwei Chen <condy0919@gmail.com>

PR #1603, opened 2024-01-20
Add missing virtual destructor

agajdac:
Add missing virtual destructor


derekmauro:
This is not the first time we've received this pull request. See https://github.com/abseil/abseil-cpp/pull/928. I believe it still does not work.

agajdac:
Interesting case: `class CommandLineFlag` is leaking its destructor via friend class and that's why we are getting a warning about missing virtual destructor. Also, we can't make it virtual, because it would make the type not literal.
I added `-Wnon-virtual-dtor` flag and pragmas to suppress the warning. Could you please take a look? @derekmauro

PR #1615, opened 2024-02-11
inlined_vector: Use trivial relocation for `erase`

Quuxplusone:
Prior art for the `vector::erase` optimization:
https://github.com/AmadeusITGroup/amc/blob/efcb7be/include/amc/vectorcommon.hpp#L176-L180
https://github.com/bloomberg/bde/blob/e15f05be6/groups/bsl/bslalg/bslalg_arrayprimitives.h#L3787-L3799
https://github.com/facebook/folly/blob/d24bf04/folly/FBVector.h#L1254-L1262
https://github.com/qt/qtbase/blob/fbfee2d/src/corelib/tools/qarraydataops.h#L856-L861

attn @derekmauro @jacobsa

PR #1618, opened 2024-02-12
inlined_vector: Use trivial relocation for `SwapInlinedElements`

Quuxplusone:
I noticed while working on #1615 that `inlined_vector` could use the trivial relocatability trait here, too.
Here the memcpy codepath already exists; we just have to opt in to using it.

### DIFF
--- a/absl/container/inlined_vector_test.cc
+++ b/absl/container/inlined_vector_test.cc
@@ -304,6 +304,33 @@ TEST(UniquePtr, MoveAssign) {
   }
 }
 
+// Swapping containers of unique pointers should work fine, with no
+// leaks, despite the fact that unique pointers are trivially relocatable but
+// not trivially destructible.
+TEST(UniquePtr, Swap) {
+  for (size_t size1 = 0; size1 < 5; ++size1) {
+    for (size_t size2 = 0; size2 < 5; ++size2) {
+      absl::InlinedVector<std::unique_ptr<size_t>, 2> a;
+      absl::InlinedVector<std::unique_ptr<size_t>, 2> b;
+      for (size_t i = 0; i < size1; ++i) {
+        a.push_back(std::make_unique<size_t>(i + 10));
+      }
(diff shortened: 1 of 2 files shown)


derekmauro:
Going to try to import this now. I don't expect any issues.

Quuxplusone:
Awesome, 92c8575d24cb5c27718e1cd1ed21a9847351a11b LGTM!
FWIW, re this comment:
```
// TODO(absl-team): Using unique_ptr here is technically correct, but
// a trivially relocatable struct would be less semantically confusing.
TEST(UniquePtr, Swap)
```
notice that we definitely _don't_ want to use a trivially _copyable_ struct, because that wouldn't sufficiently test the behavior we're trying to test. I assume the tests in the `UniquePtr` group use `unique_ptr` heap-allocation because we can easily valgrind the tests to double-check that each element's value is destroyed once (not twice or zero times). At least I hope it's true that you can easily (and nightly) valgrind the tests. If not, I'd be the first to agree that all this stuff is woefully under-tested! :)

derekmauro:
> FWIW, re this comment:
> 
> ```
> // TODO(absl-team): Using unique_ptr here is technically correct, but
> // a trivially relocatable struct would be less semantically confusing.
> TEST(UniquePtr, Swap)
> ```
> 
> notice that we definitely _don't_ want to use a trivially _copyable_ struct, because that wouldn't sufficiently test the behavior we're trying to test. I assume the tests in the `UniquePtr` group use `unique_ptr` heap-allocation because we can easily valgrind the tests to double-check that each element's value is destroyed once (not twice or zero times). At least I hope it's true that you can easily (and nightly) valgrind the tests. If not, I'd be the first to agree that all this stuff is woefully under-tested! :)

A code reviewer suggested that comment, but you are correct. Maybe it would be better to `static_assert` that the `unique_ptr` has the properties that we want for testing as a form of self documentation, along with a comment about why we choose that type. We don't have valgrind testing, but we do have MSAN testing, which (I think?) is just as good.

PR #1626, opened 2024-02-15
time: Add a way for the client application to provide fallback zoneinfo data

electroly:
This is a feature addition to the Abseil `time` library. It adds the functions `absl:::LoadStaticZoneInfoFile()` and `absl:::UnloadStaticZoneInfoFiles()` which allows a client application to provide its own zoneinfo data, loaded through whatever means it finds appropriate. These are used as a fallback when no other zoneinfo data is present on the system. The client application calls this function once per zoneinfo file at startup. I am using this to support two use cases:

- Non-standard Linux environments such as an empty chroot or "distroless" Docker container where only the application binary itself is present on the filesystem.
- Standard Windows environments.

By shipping my own statically linked zoneinfo data, I can ensure there is a fallback in these cases.

I did my best but it's my first Abseil contribution, so I'm sure this code needs some rework. The code change is trivial and I'm happy to rewrite it to address any review guidance. There are two main points I was unsure about:

- The implementation touches the internals of CCTZ. Should I try to upstream the change to the CCTZ project first, or has Abseil's fork of CCTZ fully diverged? I've assumed the latter.
- The new functionality is not thread safe. A client application would need to ensure they aren't loading or unloading zoneinfo files at the same time they are querying for time zones. Should we protect the static storage with a mutex or is that not appropriate to do here? I've assumed the latter.


derekmauro:
Abseil's primary purpose is to be used as a dependency by Google's open source C++ projects. While we do hope that Abseil is also useful to the C++ community at large, this added constraint also means that we are unlikely to accept a contribution of utility code that isn't already widely used by Google. As implemented, this feature isn't something we need.

We do however use almost exactly this feature ourselves, we just implement it differently.

You will note that CCTZ contains a weak symbol [`zone_info_source_factory`](https://github.com/google/cctz/blob/981dc6b192121ebbae8431c26069cadbb86b5930/include/cctz/zone_info_source.h#L84) that we override to provide CCTZ with exactly this type of static data that can be linked into a binary.

While this method isn't 100% portable, it does have the advantage that we can use it without calling an initialization function first (`LoadStaticZoneInfoFile` in your implementation), which means that CCTZ and the Abseil time library can be used at any point in the program's lifetime. This is important for us.

In Abseil, the weak symbol is in an "internal" namespace, which means we could take it away or change it at any time, but you are free to use it at your own risk.

As for your question about the relationship between CCTZ and Abseil, they have not diverged. We pull in changes to CCTZ automatically all the time. Changes to CCTZ must go to the CCTZ project.

@devbww for thoughts as the primary owner of CCTZ.

electroly:
Hi Derek, thanks so much for the response. `zone_info_source_factory` appears to be exactly what I need. I'll close this PR--it's now clear this functionality already exists. I appreciate you taking the time.

PR #1627, opened 2024-02-21
[Android] Remove unnecessary __system_property_get definition

jstaahl:
See: https://github.com/chromium/chromium/commit/03d8ffbb86e87ee36e2aa6dc8e650dd0a2ba1fe2

The example cited in the code comment from chromium has been removed. It turns out this was never necessary and it is not clear why it was added to chromium in the first place (see the commit message from the commit linked above).

Although this definition of __system_property_get does not cause any issues because it is defined in an anonymous namespace, it is unnecessary and misleading, and we should remove it in favor of the one directly available from libc.so.

### DIFF
--- a/absl/time/internal/cctz/src/time_zone_lookup.cc
+++ b/absl/time/internal/cctz/src/time_zone_lookup.cc
@@ -17,9 +17,6 @@
 
 #if defined(__ANDROID__)
 #include <sys/system_properties.h>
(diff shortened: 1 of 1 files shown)


derekmauro:
Please send this to https://github.com/google/cctz. It will be automatically imported from there.

jstaahl:
> Please send this to https://github.com/google/cctz. It will be automatically imported from there.

Opened a PR here:
https://github.com/google/cctz/pull/286

PR #1631, opened 2024-02-29
unscaledcycleclock: use RDTIME instead of RDCYCLE on RISC-V

aurel32:
Starting with Linux 6.6 [1], RDCYCLE is a privileged instruction on RISC-V and can't be used directly from userland. There is a sysctl option to change that as a transition period, but it will eventually disappear.

Use RDTIME instead, which while less accurate has the advantage of being synchronized between CPU (and thus monotonic) and of constant frequency. Therefore this change also fixes another issue caused by wrong time measurement when a process is rescheduled from one CPU to another [2].

Given RDTIME uses a timer independent from the CPU, this patche removes the ABSL_INTERNAL_UNSCALED_CYCLECLOCK_FREQUENCY_IS_CPU_FREQUENCY define accordingly.

This fixes the testsuite on riscv64, tested on a VisionFive 2 board.

[1] https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=cc4c07c89aada16229084eeb93895c95b7eabaa3
[2] https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1025221

### DIFF
--- a/absl/base/internal/unscaledcycleclock_config.h
+++ b/absl/base/internal/unscaledcycleclock_config.h
@@ -53,7 +53,7 @@
 #if ABSL_USE_UNSCALED_CYCLECLOCK
 // This macro can be used to test if UnscaledCycleClock::Frequency()
 // is NominalCPUFrequency() on a particular platform.
-#if (defined(__i386__) || defined(__x86_64__) || defined(__riscv) || \
+#if (defined(__i386__) || defined(__x86_64__) || \
      defined(_M_IX86) || defined(_M_X64))
 #define ABSL_INTERNAL_UNSCALED_CYCLECLOCK_FREQUENCY_IS_CPU_FREQUENCY
 #endif
--- a/absl/base/internal/unscaledcycleclock.cc
+++ b/absl/base/internal/unscaledcycleclock.cc
@@ -125,7 +125,7 @@ double UnscaledCycleClock::Frequency() {
 
 int64_t UnscaledCycleClock::Now() {
   int64_t virtual_timer_value;
-  asm volatile("rdcycle %0" : "=r"(virtual_timer_value));
+  asm volatile("rdtime %0" : "=r"(virtual_timer_value));
   return virtual_timer_value;
 }
 


absl/base/internal/unscaledcycleclock_config.h
```diff
@@ -53,7 +53,7 @@
 #if ABSL_USE_UNSCALED_CYCLECLOCK
 // This macro can be used to test if UnscaledCycleClock::Frequency()
 // is NominalCPUFrequency() on a particular platform.
-#if (defined(__i386__) || defined(__x86_64__) || defined(__riscv) || \
+#if (defined(__i386__) || defined(__x86_64__) || \
      defined(_M_IX86) || defined(_M_X64))
 #define ABSL_INTERNAL_UNSCALED_CYCLECLOCK_FREQUENCY_IS_CPU_FREQUENCY
```
derekmauro, line 58: We need an implementation of `UnscaledCycleClock::Frequency()` now that these are not the same thing. The code in this repo may not call that method, but code exists that does.

https://riscv.org/wp-content/uploads/2016/06/riscv-spec-v2.1.pdf says
> The RDTIME pseudo-instruction reads the low XLEN bits of the time CSR, which counts wall-clock
real time that has passed from an arbitrary start time in the past. RDTIMEH is an RV32I-only instruction that reads bits 63–32 of the same real-time counter. The underlying 64-bit counter should
never overflow in practice. The execution environment should provide a means of determining the
period of the real-time counter (seconds/tick). The period must be constant. The real-time clocks
of all hardware threads in a single user application should be synchronized to within one tick of the
real-time clock. The environment should provide a means to determine the accuracy of the clock.

I have no idea how to get that.

If you are just trying to get something to build, `-DABSL_USE_UNSCALED_CYCLECLOCK=0` should work.

derekmauro (changes requested):

absl/base/internal/unscaledcycleclock_config.h
aurel32: The frequency is implementation dependent, and there is no easy way to get it. Could it be measured instead? Basically how it is done for the `ABSL_INTERNAL_UNSCALED_CYCLECLOCK_FREQUENCY_IS_CPU_FREQUENCY`?

aurel32:
Closing in favor of https://github.com/abseil/abseil-cpp/pull/1644

PR #1637, opened 2024-03-06
Allow including with angle brackets from Bazel

### LINKED ISSUE #740: Allow Bazel users to include Abseil headers with angle brackets.
Because we use Abseil with both CMake and Bazel, and we sometimes use Abseil from `/usr/include` (or `/usr/local/include` or some other $PREFIX) we need to always include Abseil headers with either quotes (`#include "absl/..."`) or angle brackets (`#include <absl/...>`).

We would prefer to include Abseil with angle brackets. In some platforms that allows us to treat warnings as errors (`-Werror -Wall -W...` or `/WX /W3`) even if the Abseil code is not "clean" for some warning. In general Abseil is very very clean, thanks! But there are a few exceptions, e.g. #582 .

adzenith:
Fixes #740 by adding `includes` to the config target. We have been using this locally and it works great for us.

### DIFF
--- a/absl/base/BUILD.bazel
+++ b/absl/base/BUILD.bazel
@@ -140,6 +140,10 @@ cc_library(
         "policy_checks.h",
     ],
     copts = ABSL_DEFAULT_COPTS,
+    # This library is the root of the absl dependency tree.
+    # If we add this `includes` path here, then all absl headers
+    # can be included via angle brackets in external projects.
+    includes = ["../.."],
     linkopts = ABSL_DEFAULT_LINKOPTS,
 )
 


derekmauro:
I'd really like to support angle brackets from Bazel, but this is not the solution. This solution is cute, but ultimately it's a hack.

This also simply fails if Abseil is built as the root project:
```
bazel test ...
Starting local Bazel server and connecting to it...
ERROR: /usr/local/google/home/dmauro/tmp/ang/abseil-cpp/absl/base/BUILD.bazel:135:11: in cc_library rule //absl/base:config: 
Traceback (most recent call last):
	File "/virtual_builtins_bzl/common/cc/cc_library.bzl", line 61, column 56, in _cc_library_impl
	File "/virtual_builtins_bzl/common/cc/cc_helper.bzl", line 1077, column 17, in _system_include_dirs
Error in fail: attribute includes: '../..' resolves to the workspace root, which would allow this rule and all of its transitive dependents to include any file in your workspace. Please include only what you need
ERROR: /usr/local/google/home/dmauro/tmp/ang/abseil-cpp/absl/base/BUILD.bazel:135:11: Analysis of target '//absl/base:config' failed
ERROR: Analysis of target '//absl/time/internal/cctz:civil_time' failed; build aborted: Analysis failed
INFO: Elapsed time: 5.860s, Critical Path: 0.04s
INFO: 1 process: 1 internal.
ERROR: Build did NOT complete successfully
ERROR: No test targets were found, yet testing was requested
```

adzenith:
Fair enough! Thanks for taking a look. And others who come across #740 at least can add this patch locally if they want to use angle brackets for now.

fruffy:
Thanks for this patch, we also ran into this problem.

fruffy:
>I'd really like to support angle brackets from Bazel, but this is not the solution. This solution is cute, but ultimately it's a hack.

Do you have a suggestion for a better solution? Abseil not supporting system includes is causing significant problems for us. We are currently using this patch, but this can cause problems with other repositories depending on our project which may include their own version of Abseil.

derekmauro:
> Do you have a suggestion for a better solution? Abseil not supporting system includes is causing significant problems for us. We are currently using this patch, but this can cause problems with other repositories depending on our project which may include their own version of Abseil.

I can think of two options.
1) Restructure the Abseil project so that the root directory is not the directory that needs to be specified as the includes directory, for example, by moving the source code from `absl/` to `src/absl/`.
2) Consider this a deficiency in Bazel and get Bazel to allow using the root directory for angle includes. If projects need to do (1) to play nice in the ecosystem, then this seems like a Bazel issue.

Please do not send us a PR for (1). It may seem easy to you, but it would require a lot of work on our end to import that. It would be better left to us.

smolkaj:
> > I'd really like to support angle brackets from Bazel, but this is not the solution. This solution is cute, but ultimately it's a hack.
> 
> Do you have a suggestion for a better solution? Abseil not supporting system includes is causing significant problems for us. We are currently using this patch, but this can cause problems with other repositories depending on our project which may include their own version of Abseil.

+1. Worth pointing out that this is happening in repositories owned by/relied upon by Google.

smolkaj:
Would `include_prefix` help here? https://stackoverflow.com/q/46498441

derekmauro:
> Would include_prefix help here? https://stackoverflow.com/q/46498441

The Bazel docs say that it is only legal under `third_party/`.

smolkaj:
Could be wrong, but I suspect that comment is referring to blaze / google? I do see the attribute being used all over GitHub: https://github.com/search?q=%22include_prefix+%3D+%5C%22%22&type=code

fruffy:
> > Do you have a suggestion for a better solution? Abseil not supporting system includes is causing significant problems for us. We are currently using this patch, but this can cause problems with other repositories depending on our project which may include their own version of Abseil.
> 
> I can think of two options.
> 
>     1. Restructure the Abseil project so that the root directory is not the directory that needs to be specified as the includes directory, for example, by moving the source code from `absl/` to `src/absl/`.
> 
>     2. Consider this a deficiency in Bazel and get Bazel to allow using the root directory for angle includes. If projects need to do (1) to play nice in the ecosystem, then this seems like a Bazel issue.
> 
> 
> Please do not send us a PR for (1). It may seem easy to you, but it would require a lot of work on our end to import that. It would be better left to us.

So what is different about the structure of, say, Protobuf, that allows us to have angled-bracketed includes. Is it only because the source files are in `src/google/protobuf`?

derekmauro:
At least on GCC and Clang, from my testing it looks like you can use `bazel build --features=external_include_paths //path/to:target` to use angle brackets. I did not try MSVC, but https://github.com/bazelbuild/bazel/commit/850b33430edbe188249a9dde784271fbd8919c7f makes it look like it works there too.

fruffy:
Great! Needs some more testing but it looks like this trick may work https://github.com/p4lang/p4c/pull/4640.

fruffy:
Do you by any chance know whether we can export this feature flag to any downstream dependency which may depend on our project? Unclear whether the `.bazelrc` we are using is respected here.

derekmauro:
> Do you by any chance know whether we can export this feature flag to any downstream dependency which may depend on our project? Unclear whether the .bazelrc we are using is respected here.

I'm pretty sure the answer is that you can't do that. The root project needs to decide how everything is built, otherwise you will get conflicting options which lead to incorrect builds.

fruffy:
> > Do you by any chance know whether we can export this feature flag to any downstream dependency which may depend on our project? Unclear whether the .bazelrc we are using is respected here.
> 
> I'm pretty sure the answer is that you can't do that. The root project needs to decide how everything is built, otherwise you will get conflicting options which lead to incorrect builds.

Yes, we just confirmed. How much work is necessary for (1)? 

>Restructure the Abseil project so that the root directory is not the directory that needs to be specified as the includes directory, for example, by moving the source code from absl/ to src/absl/.

It looks like it is the only viable option at this point. Or are we stuck?

PR #1640, opened 2024-03-15
Adding cmake option ABSL_USE_HARDENED_MODE

IcebrakerCZ:
Adding cmake option ABSL_USE_HARDENED_MODE with default value set to OFF.


CMakeLists.txt
```diff
@@ -149,6 +149,9 @@ set(ABSL_LOCAL_GOOGLETEST_DIR "/usr/src/googletest" CACHE PATH
   "If ABSL_USE_GOOGLETEST_HEAD is OFF and ABSL_GOOGLETEST_URL is not set, specifies the directory of a local GoogleTest checkout."
   )
 
+option(ABSL_USE_HARDENED_MODE
+  "If ON, abseill will sue hardened mode." OFF)
```
derekmauro: Please fix the spelling here.

derekmauro (changes requested):
I think we need a test that this actually works. We don't want to accidentally break the option.

CMakeLists.txt
IcebrakerCZ: Fixed the spelling in commit 43795ba.

IcebrakerCZ: Added hardened mode to cmake test builds where I found that shared mode is tested in commit d767e1e.

I have checked the changes against typos but I can't test it so if you can test it in your environment please test it and propose changes if I did something wrong.

Or is there a way how I can test the testing build scripts on my own?

IcebrakerCZ: So maybe I should be able to test it on my Ubuntu 20.04 and my macOS Ventura.

IcebrakerCZ: linux_gcc_alpine_cmake.sh and linux_gcc-latest_libstdcxx_cmake.sh and macos_xcode_cmake.sh finished successfully on my systems (Ubuntu Focal 20.04 and macOS Ventura 13.6.4)

IcebrakerCZ: I can't test windows_msvc_cmake.bat so please run tests in your env if possible.

If you would like to test the new cmake option somehow else please let me know.

Another approach would be set 0 to hardened mode in absl_alternate_options.h and set -DABSL_USE_HARDENED_MODE=ON in the cmake test scripts but I didn't find where the absl_alternate_options.h is used :/

CMakeLists.txt
```diff
@@ -149,6 +149,9 @@ set(ABSL_LOCAL_GOOGLETEST_DIR "/usr/src/googletest" CACHE PATH
   "If ABSL_USE_GOOGLETEST_HEAD is OFF and ABSL_GOOGLETEST_URL is not set, specifies the directory of a local GoogleTest checkout."
   )
 
+option(ABSL_USE_HARDENED_MODE
+  "If ON, abseil will sue hardened mode." OFF)
```
derekmauro: The word "use" is still misspelled here. Also, please capitalize Abseil.

derekmauro (changes requested):
As implemented, the tests are neither scalable nor effective.

The problem with build flags is that they generate a combinatoric explosion of builds that need to be run. The loop added in the tests doubles the build time to test a single feature. Imagine if there were three more boolean features that needed to be tested. That would cause a 16x increase in test time.  The workaround we use in a Bazel build is to assume all features are orthogonal and use a single file turn this into another build: https://github.com/abseil/abseil-cpp/blob/master/ci/absl_alternate_options.h

I say the tests are not effective because nothing is testing that the CMake change is actually changing `options.h`. That is what really ought to be tested. Test coverage for hardened mode already exists.

CMakeLists.txt
IcebrakerCZ: Fixed in commit 3e2bf646

IcebrakerCZ: Reverted testing changes in commit 3082c45b

IcebrakerCZ: I understand that double the number of builds and tests is generally not a good idea.

Thanks for pointing out bazel builds I found that the alternate options file is used there form environment so hardened mode is tested.

I don't know how you want to be the cmake option tested.
Normally I would create another shell script that:
    - check that default values for boolean options from 'cmake -L'
    - configure abseil with non-default boolean values
    - check that generated options.h match configured non-default values
but this I'm not sure I can develop it just in few code review cycles to match Abseil expectations.

I really don't know how to proceed further so please suggest how the test should look like and where should be placed etc.

IcebrakerCZ:
Any advice how to proceed with the requested new cmake option testing?
There are not tests for cmake options now.
I need advice how you would like it and if it must work on all OSes etc.

derekmauro:
There are actually new plans for hardened mode options, so I'd rather not accept this change at this time.

IcebrakerCZ:
I see so in that case please reject this code review :) thanks Derek

PR #1642, opened 2024-03-15
Small cleanup Bazel build system

TsynkPavel:
Hello !

I was exploring Bazel build system and noticed several insignificant `select` constructions.
Let me know if you need them for anything.\

My goal is to make project cleaner!


derekmauro:
These are intentional. The code you see on GitHub is the result of exporting Abseil from our internal codebase. Prior to export some of our internal selects are removed, resulting in the empty select. These need to remain.

PR #1655, opened 2024-04-18
Remove unnecessary limitation for clang-cl or icx

pps83:
clang-cl or icx (Intel's new llvm-based compiler) both define _MSC_VER on windows. Both compile this code without issues.


pps83:
This PR depends on https://github.com/abseil/abseil-cpp/pull/1654. Perhaps, the `__clang__` check was there instead of correct check for x64 build. Once #1654 is merged, then this PR can be applied.

PR #1662, opened 2024-04-25
Replace shift with addition in crc multiply

pps83:


derekmauro:
This looks like a nice simplification, but I didn't see any change to the benchmarks. Did you see an improvement anywhere? How did you find this?

derekmauro:
Getting a build error on an ARM platform:
```
absl/crc/internal/crc32_x86_arm_combined_simd.h:292:61: error: no matching function for call to 'vaddq_s64'
  292 | inline V128 V128_Add64(const V128 l, const V128 r) { return vaddq_s64(l, r); }
      |                                                             ^~~~~~~~~
aarch64/usr/lib64/clang/18/include/arm_neon.h:2266:16: note: candidate function not viable: no known conversion from 'const V128' (aka 'const uint64x2_t') to 'int64x2_t' (vector of 2 'int64_t' values) for 1st argument
 2266 | __ai int64x2_t vaddq_s64(int64x2_t __p0, int64x2_t __p1) {
```

pps83:
> This looks like a nice simplification, but I didn't see any change to the benchmarks. Did you see an improvement anywhere? How did you find this?

simply because modern clang/gcc actually removed your shl 1 and replaced with an add. ms compiler doesn't do that.

pps83:
> Getting a build error on an ARM platform:

I'll address that shortly when I get back to my pc

pps83:
> How did you find this?

https://godbolt.org/z/xGr9rKnEv xmul in the clang version doesn't have shl. That's how I noticed that it's better to add just like the compiler optimized it.

pps83:
> Getting a build error on an ARM platform:
> 
> ```
> absl/crc/internal/crc32_x86_arm_combined_simd.h:292:61: error: no matching function for call to 'vaddq_s64'
>   292 | inline V128 V128_Add64(const V128 l, const V128 r) { return vaddq_s64(l, r); }
>       |                                                             ^~~~~~~~~
> aarch64/usr/lib64/clang/18/include/arm_neon.h:2266:16: note: candidate function not viable: no known conversion from 'const V128' (aka 'const uint64x2_t') to 'int64x2_t' (vector of 2 'int64_t' values) for 1st argument
>  2266 | __ai int64x2_t vaddq_s64(int64x2_t __p0, int64x2_t __p1) {
> ```

@derekmauro PR updated, please retry

derekmauro:
Still getting an error here: https://godbolt.org/z/x9aT4nqcW (Note that `-fno-lax-vector-conversions` is required to get the error.)

It looks like `vaddq_u64(l, r);` fixes this. I'm testing this now.

pps83:
> Still getting an error here: https://godbolt.org/z/x9aT4nqcW (Note that `-fno-lax-vector-conversions` is required to get the error.)

result from vaddq_s64 should be converted back.

> It looks like `vaddq_u64(l, r);` fixes this. I'm testing this now.

that's the proper change actually. shl is unsigned add. PR updated

PR #1663, opened 2024-04-28
Simplify next_weekday and prev_weekday calculus

teksturi:
It seems that calculating next and previous weekday is over complicated. We can make it simpler and faster. On my test on x86-64 I get following results:

before (release build)
Benchmark                                 Time             CPU   Iterations
BM_NextWeekday                         5.25 ns         5.25 ns    133517356
BM_PrevWeekday                         4.53 ns         4.52 ns    154113771

after (release build)
BM_NextWeekday                         3.72 ns         3.72 ns    177678067
BM_PrevWeekday                         3.51 ns         3.51 ns    188256873

This is also safer on situations where this function gets weekday as not enum. Yeah that is unlikely but one more plus for this.

Only point for get_day_num() over casting was that it looked like it is habit that we do not cast weekdays straight. Straight casting was slightly faster (about 0.25ns) but this way was chosen.

### DIFF
--- a/absl/time/internal/cctz/include/cctz/civil_time_detail.h
+++ b/absl/time/internal/cctz/include/cctz/civil_time_detail.h
@@ -561,44 +561,43 @@ CONSTEXPR_F weekday get_weekday(const civil_second& cs) noexcept {
 
 ////////////////////////////////////////////////////////////////////////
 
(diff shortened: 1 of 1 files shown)


derekmauro:
This pull request needs to be sent to https://github.com/google/cctz. We import this code from there.

PR #1672, opened 2024-05-11
Optimize StrJoin with tuple without user defined formatter

MBkkt:
https://github.com/abseil/abseil-cpp/discussions/1671

Unfortunately the benchmarks can't be used with CMake and I don't have time to fix that or figure out how to use Bazel.


derekmauro:
It may take me a few days to run some tests to make sure this doesn't break anything, but initially I'd say this is good work.

For reference, to build the benchmark:
`bazel build -c opt absl/strings:str_join_benchmark`

Then to run the benchmark:
`bazel-bin/absl/strings/str_join_benchmark`

Although it looks like the existing benchmark doesn't cover this case.

PR #1692, opened 2024-06-20
Add missing `<utility>` include

poconn:
`optimization.h` uses `std::unreachable()` if available but does not include `<utility>`, causing errors in `absl/strings/ascii.cc`.

### DIFF
--- a/absl/base/optimization.h
+++ b/absl/base/optimization.h
@@ -24,6 +24,10 @@
 
 #include <assert.h>
 
+#if defined(__cpp_lib_unreachable) && __cpp_lib_unreachable >= 202202L
+#include <utility>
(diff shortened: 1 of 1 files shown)


poconn:
I see this was reverted because it needs to be compilable as C. Looks like somebody edited the commit to make the include unconditional (6f972e2), I think the original should work as is?

derekmauro:
We will handle this.

derekmauro:
Now fixed in https://github.com/abseil/abseil-cpp/commit/a305e859b6d0abe6d88a3fe12c6891152cc6e859. Thanks!

poconn:
Great, thanks!

PR #1695, opened 2024-06-21
Fix time library build for Apple platforms

EduMenges:
#1495 converted the `if (APPLE)` check for linking against `CoreFoundation` to  generator expressions, which is fine and all.
The issue is that they forgot the other Apple platforms, making builds for iOS impossible.
This patch fixes this issue by adding the other Apple platforms that CMake support to the generator expression.

### DIFF
--- a/absl/time/CMakeLists.txt
+++ b/absl/time/CMakeLists.txt
@@ -83,7 +83,7 @@ absl_cc_library(
     Threads::Threads
     # TODO(#1495): Use $<LINK_LIBRARY:FRAMEWORK,CoreFoundation> once our
     # minimum CMake version >= 3.24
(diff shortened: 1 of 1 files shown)


EduMenges:
@googlebot I signed it!

derekmauro (commented):
According to https://cmake.org/cmake/help/latest/release/3.15.html#id15, this syntax is not supported until CMake 3.15. Our global configuration (https://github.com/abseil/abseil-cpp/blob/master/CMakeLists.txt) says our minimum is 3.10, but https://github.com/google/oss-policies-info/blob/main/foundational-cxx-support-matrix.md says we can raise this to 3.16. We should raise the minimum version first.

PR #1697, opened 2024-06-26
Fix poisoned pointer on powerpc

miladfarca:
PPC has 64K pages instead of 4k, fix is needed for mprotect as it needs a page aligned input.
We build and test abseil as it's a dependency of V8 which we maintain on PPC.

### DIFF
--- a/absl/base/internal/poison.cc
+++ b/absl/base/internal/poison.cc
@@ -35,7 +35,11 @@ namespace absl {
 ABSL_NAMESPACE_BEGIN
 namespace base_internal {
 namespace {
+#if defined(__powerpc64__)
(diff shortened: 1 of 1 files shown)


derekmauro:
Thanks for letting us know about the page size issue. The other issue with the implementation is that it uses a global constructor, which is banned by Chromium. I would actually like to change the implementation to avoid both of these issues. I'm working on something that uses `getpagesize()` and friends instead.

miladfarca:
Thanks Derek, is it possible to land this change until your patch is applied?

derekmauro:
I actually rolled this back because we were not satisfied that we had a performant, portable way of implementing this that did not rely on undefined behavior.

PR #1704, opened 2024-07-06
Use reference in range-for with non trivial type

gruenich:
Found by Clazy (range-loop-reference)


derekmauro:
Pull requests for this code must be sent to https://github.com/google/cctz. We will pull this in from there.

PR #1709, opened 2024-07-10
Handle RPATH CMake configuration

BotellaA:
When depending on a compiled abseil code, abseil libraries cannot find each other even if they are all located in the same directory. 

One fix is to edit the LD_LIBRARY_PATH, but it is not always that simple in a development environment. 

Another way is to add the current location into the RPATH so all libraries can find each other. CMake provides simple command to control this at project scale.


BotellaA:
I see there is an issue on the MacOS job but I cannot open the log...

derekmauro:
The MacOS issue was an infrastructure failure (the remote machine disconnected). I'm re-running it.

nanonyme:
`set(CMAKE_INSTALL_RPATH "$ORIGIN")`  sounds incorrect. This is not for development but installs. abseil is now writing useless RPATH for distributions. `set(CMAKE_BUILD_RPATH_USE_ORIGIN ON)` seems fine.

PR #1711, opened 2024-07-12
Disambiguate use of template as per CWG1835

glandium:
Clang trunk has recently landed the resolution for https://cplusplus.github.io/CWG/issues/1835.html, yielding the following errors:

In file included from /tmp/abseil-cpp/absl/status/status.cc:14: In file included from /tmp/abseil-cpp/absl/status/status.h:66: In file included from /tmp/abseil-cpp/absl/status/internal/status_internal.h:26: In file included from /tmp/abseil-cpp/absl/container/inlined_vector.h:53: In file included from /tmp/abseil-cpp/absl/container/internal/inlined_vector.h:32: /tmp/abseil-cpp/absl/container/internal/compressed_tuple.h:251:42: error: no member named 'get' in the global namespace
  251 |     return std::move(*this).StorageT<I>::get();
      |                                        ~~^
/tmp/abseil-cpp/absl/container/internal/compressed_tuple.h:256:42: error: no member named 'get' in the global namespace
  256 |     return std::move(*this).StorageT<I>::get();
      |                                        ~~^
2 errors generated.

### DIFF
--- a/absl/container/internal/compressed_tuple.h
+++ b/absl/container/internal/compressed_tuple.h
@@ -248,12 +248,14 @@ class ABSL_INTERNAL_COMPRESSED_TUPLE_DECLSPEC CompressedTuple
 
   template <int I>
   constexpr ElemT<I>&& get() && {
-    return std::move(*this).StorageT<I>::get();
+    using StorageInt = StorageT<I>;
+    return std::move(*this).StorageInt::get();
   }
 
   template <int I>
   constexpr const ElemT<I>&& get() const&& {
-    return std::move(*this).StorageT<I>::get();
+    using StorageInt = StorageT<I>;
+    return std::move(*this).StorageInt::get();
   }
 };
 


derekmauro (commented):
This change only compiles under Clang and breaks under all other compilers.

glandium:
What is GCC/libstdc++ (floor)? I guess it's an old version of GCC? GCC 12 works for me, and the GCC/libstdc++ (latest) work...
How does one get the logs from the failing jobs?

pranavk:
Can we use similar workaround that was used in clang PR:

That is, use following instead:
```
using StorageInt = StorageT<I>;
return std::move(*this).StorageInt::get();
```

@glandium Would you be able to update your PR with this change and see if it fixes the issue please? This is blocking us internally. So I would like to get it ASAP. But I also don't want to hijack your currently open PR.

Here's what failing logs from GCC are like:
```
./absl/container/internal/compressed_tuple.h: In member function 'constexpr absl::container_internal::CompressedTuple<Ts>::ElemT<I>&& absl::container_internal::CompressedTuple<Ts>::get() &&':
[./absl/container/internal/compressed_tuple.h:251]:49: error: expected ';' before '::' token
     return std::move(*this).template StorageT<I>::get();
                                                 ^~
[./absl/container/internal/compressed_tuple.h:251]:51: error: '::get' has not been declared
     return std::move(*this).template StorageT<I>::get();
 ```
 
 They are exactly same as discussed in https://github.com/llvm/llvm-project/pull/98252/files

**Comment by glandium** [comment]
Funnily enough, this adds a warning:
```
/tmp/abseil-cpp/absl/container/internal/compressed_tuple.h:251:11: warning: unused type alias 'StorageInt' [-Wunused-local-typedef]
  251 |     using StorageInt = StorageT<I>;
      |           ^
/tmp/abseil-cpp/absl/container/internal/compressed_tuple.h:257:11: warning: unused type alias 'StorageInt' [-Wunused-local-typedef]
  257 |     using StorageInt = StorageT<I>;
      |           ^
```

**Comment by pranavk** [comment]
Yeah, I also stumbled upon this. It smells like a bug to me. Not a language lawyer but I fail to see why that is marked as 'unused'. Could be a bug?

**Comment by glandium** [comment]
It sounds like one. GCC doesn't complain, but even older clang versions complain.

**Comment by glandium** [comment]
I updated the PR with the version with a warning, to at least see how it goes with MSVC.

**Comment by hokein** [comment]
> Yeah, I also stumbled upon this. It smells like a bug to me. Not a language lawyer but I fail to see why that is marked as 'unused'. Could be a bug?

Yeah, the unused `unused-local-typedef` warning seems like a false positive (clang doesn't make the reference bit in the alias decl if the use in a dependent nested name specifier?), I will file a bug (edit: filed https://github.com/llvm/llvm-project/issues/98878)

Maybe we can disable this warning via the pragma.

```
#if defined(__clang__)
#pragma clang diagnostic push
#pragma clang diagnostic ignored "-Wunused-local-typedef"
#endif

// the fix ...


#if defined(__clang__)
#pragma clang diagnostic pop
#endif
```

derekmauro:
> Maybe we can disable this warning via the pragma.

I had the same suggestion in the internal code review thread.

I'm closing this because this is being handled in the internal code review. cl/651824347 for anyone interested.

cor3ntin:
FYI, the fix here should probably be along the lines of `return std::move(*this).template StorageT<I>::get();`

glandium:
> FYI, the fix here should probably be along the lines of `return std::move(*this).template StorageT<I>::get();`

That's what this [PR was doing first](https://github.com/abseil/abseil-cpp/commit/867b5efb9ca3588927ffce3a278cd0c43aa5a6b5), but MSVC and older versions of GCC don't like it.

PR #1713, opened 2024-07-12
Use alias declaration for new clang

pranavk:


derekmauro:
We don't need a PR. You can test your internal CL. I put the instructions in the code review thread.

PR #1730, opened 2024-07-26
chore(bzlmod): set dev_dependency true for googletest

4kangjc:


derekmauro:
It's understandable that you think this is the right thing to do, but I believe it is incorrect. Abseil provides libraries for writing tests that depend on GoogleTest. `dev_dependency` means "If true, this dependency will be ignored if the current module is not the root module or `--ignore_dev_dependency` is enabled." If you want to use one of the test libraries, you need a dependency on GoogleTest. I actually don't think `dev_dependency` solves any real problem and should be removed from Bazel.

PR #1738, opened 2024-08-03
Fix shell option group handling in pkgconfig files

ZhongRuoyu:
Since #1707 and #1710, `-Xarch_` option groups are present as single elements in the libraries' COPTS, and a `SHELL:` prefix is added to each of them. While these addressed certain build issues, they broke the handling of `-Xarch_` option groups in pkgconfig files.

Fix that by taking care of the `SHELL:` prefix in COPTS when generating pkgconfig files. The skip-next-flag mechanism is also removed as the option groups are now present as single elements.

Cc: @Royna2544 @anpol @derekmauro

### DIFF
--- a/CMake/AbseilHelpers.cmake
+++ b/CMake/AbseilHelpers.cmake
@@ -186,15 +186,15 @@ function(absl_cc_library)
         endif()
       endif()
     endforeach()
-    set(skip_next_cflag OFF)
     foreach(cflag ${ABSL_CC_LIB_COPTS})
-      if(skip_next_cflag)
-        set(skip_next_cflag OFF)
-      elseif(${cflag} MATCHES "^-Xarch_")
+      # Strip out the CMake-specific `SHELL:` prefix, which is used to construct
+      # a group of space-separated options.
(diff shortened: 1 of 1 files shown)


CMake/AbseilHelpers.cmake
```diff
@@ -186,15 +186,14 @@ function(absl_cc_library)
         endif()
       endif()
     endforeach()
-    set(skip_next_cflag OFF)
     foreach(cflag ${ABSL_CC_LIB_COPTS})
-      if(skip_next_cflag)
-        set(skip_next_cflag OFF)
-      elseif(${cflag} MATCHES "^-Xarch_")
+      if(cflag MATCHES "^SHELL:")
```
anpol: This `if()` is redundant, because `string(REGEX REPLACE)` works in either case.

I suggest adding another comment here, clarifying that `SHELL:` is a CMake-specific prefix used to make up a group of space-separated options.

anpol (commented):
LGTM

ZhongRuoyu:
Thanks @anpol!

PR #1762, opened 2024-09-26
Add 8/16-bit integer support to SimpleAtoi

0xERIC:
This makes `SimpleAtoi` more ergonomic for tasks like parsing port numbers. These changes were mostly trivial copies of existing code paths. Two spots worth looking more closely at are:
- The signed-ness check (`kIsSigned`) which didn't work properly for these new narrower types, and has been updated.
- The change from a boolean for type width, to the value of the width (`kUse64Bit` vs `kIntTypeSize`).

A bit of previous discussion was [here](https://github.com/abseil/abseil-cpp/discussions/1753). Even if these changes aren't merged via this PR, I'd love for this to expedite the process of getting them made internally.

### DIFF
--- a/absl/strings/numbers.cc
+++ b/absl/strings/numbers.cc
@@ -1110,6 +1110,16 @@ ABSL_CONST_INIT ABSL_DLL const char kHexTable[513] =
     "e0e1e2e3e4e5e6e7e8e9eaebecedeeef"
     "f0f1f2f3f4f5f6f7f8f9fafbfcfdfeff";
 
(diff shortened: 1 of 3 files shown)


0xERIC:
Hi @derekmauro, I wanted to check in about this change. Is there any intent to merge it? Or is this feature on ice until somebody works on it internally? Thanks in advance.

derekmauro:
Just by chance someone started working on it this week. There is a fair chance it will happen next week.

PR #1773, opened 2024-10-17
Fetch local_config_cc from rules_cc instead of bazel

keith:
With bazel 8.x cc_configure.bzl has moved to rules_cc

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -20,23 +20,34 @@ module(
     compatibility_level = 1,
 )
 
-cc_configure = use_extension("@bazel_tools//tools/cpp:cc_configure.bzl", "cc_configure_extension")
-use_repo(cc_configure, "local_config_cc")
-
 # Only direct dependencies need to be listed below.
 # Please keep the versions in sync with the versions in the WORKSPACE file.
 
-bazel_dep(name = "bazel_skylib",
-          version = "1.5.0")
+bazel_dep(
+    name = "bazel_skylib",
+    version = "1.5.0",
+)
 
-bazel_dep(name = "google_benchmark",
-          version = "1.8.3",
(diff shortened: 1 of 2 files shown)


derekmauro (changes requested):
Was this tested with any version of Bazel? I haven't tried Bazel 8, but Bazel 7 gives
```
(01:08:18) Analyzing: 510 targets (26 packages loaded, 0 targets configured)
[0 / 1] checking cached actions
(01:08:19) ERROR: Traceback (most recent call last):
	File "/root/.cache/bazel/_bazel_root/7eca63b212bb77491fc771313ed1b4be/external/rules_cc~0.0.13/cc/extensions.bzl", line 21, column 34, in _cc_configure_extension_impl
		return ctx.extension_metadata(reproducible = True)
Error in extension_metadata: extension_metadata() got unexpected keyword argument 'reproducible'
(01:08:19) Analyzing: 510 targets (63 packages loaded, 628 targets configured)
[1 / 1] checking cached actions
(01:08:20) WARNING: errors encountered while analyzing target '//absl/log:log_macro_hygiene_test', it will not be built.
error evaluating module extension cc_configure_extension in @@rules_cc~0.0.13//cc:extensions.bzl
```

Is there a way to make this change in a backwards compatible way?

keith:
this requires bazel 7.1.0+, so you must be testing with 7.0.0? it would probably be easier to rely on that point release vs trying to make this backwards compatible unless there's a strong reason not to

keith:
fixed upstream instead https://github.com/bazelbuild/bazel/pull/24154

keith:
nvm fixed here by https://github.com/abseil/abseil-cpp/commit/9e3b68ac3b328212e4ad7baada1523a05f615dde

PR #1774, opened 2024-10-21
Min viable Clang-libunwind support.

helly25:
Min viable Clang-libunwind support.

This is from an early version of GLog and the generic implementation.

It works and is intended fro using Abseil in Bazel with Clang via https://github.com/bazel-contrib/toolchains_llvm

### DIFF
--- a/absl/debugging/BUILD.bazel
+++ b/absl/debugging/BUILD.bazel
@@ -40,6 +40,7 @@ cc_library(
         "internal/stacktrace_config.h",
         "internal/stacktrace_emscripten-inl.inc",
         "internal/stacktrace_generic-inl.inc",
+        "internal/stacktrace_libunwind-inl.inc",
         "internal/stacktrace_powerpc-inl.inc",
         "internal/stacktrace_riscv-inl.inc",
         "internal/stacktrace_unimplemented-inl.inc",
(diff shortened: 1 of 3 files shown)


derekmauro:
We are not going to take this contribution, at least as-is. We have our own up-to-date copy of this file that we do not export because we do not intend to support it. In addition, the use by default when using Clang isn't going to work for us.

Can I ask what is the motivation for wanting this? That would help me figure out what to do.

helly25:
> We are not going to take this contribution, at least as-is. We have our own up-to-date copy of this file that we do not export because we do not intend to support it. In addition, the use by default when using Clang isn't going to work for us.
(I know what you have - check epitaphs)
> 
> Can I ask what is the motivation for wanting this? That would help me figure out what to do.

Here goes:
* When you compile Abseil with Bazel and use `--experimental_omitfp` then that header will be used. So it fails as the header is not present.
* The default Abseil unwind config does not work for me (due exclusively using Clang).
* The patch in its current form is geared towards that and is based on the fact that when you get Clang through toolchains_llvm, then the you will also have libunwind in external/llvm_toolchain_llvm. The detection also works if you have say Ubuntu with the corresponding Clang unwind header package (e.g. `libunwind-17-dev`).

I was also contemplating a purely unwind based solution but was happy to just drop the remaining parts of GLog in my codebase which was used for stacktrace generation prior to this patch. So it works fine as is - for me at least.

PR #1775, opened 2024-10-22
Support symbolise only on WINAPI_PARTITION_DESKTOP

redbaron:
Win32 `dbhelp.h` defines `SYMBOL_INFO` structure under `WINAPI_FAMILY_PARTITION(WINAPI_PARTITION_DESKTOP)` guard. Any other partition will have these symbols undefined.

Follow same logic in `symbolise.cc` to prevent compilation failure on non-Desktop platforms.


derekmauro (commented):
I noticed that there are 3 places where we have this exact check: https://github.com/search?q=repo%3Aabseil%2Fabseil-cpp%20WINAPI_PARTITION_DESKTOP&type=code

Should all of them be changed? (I don't know, I'm asking. I looked at the history of the code and I'm not sure whoever added the code did the right thing.)

redbaron:
They use `&& !WINAPI_FAMILY_PARTITION(WINAPI_PARTITION_DESKTOP)` condition so it will evaluate to the correct value on !DESKTOP platforms. Rest of expression is unnecessary, but it doesn't affect correctness.

PR #1777, opened 2024-10-23
Avoid std::ldexp in `operator double(int128)`.

degasus:
This patch replaces all instances of

  std::ldexp(msb, 64)

with

  msb * (2**64)

as it turns out that this optimization is not done by MSVC. Worse, it emited a function call with error checking, even if the int128 cannot hit the inf limitation.

Sadly even the constant `std::ldexp(1.0, 64)` is not inlined: https://gcc.godbolt.org/z/oGhGz77sx


derekmauro:
Nice find, thanks! Importing.

PR #1784, opened 2024-11-05
Fix CXX version build issue in Apple clang

infomaniac777:
Fix CXX version build issue in Apple clang

On the latest Mac OS 15.1, the default C++ version is gnu++98 with the default Apple clang version 16.0.0 compiler.
So the following does not complete successfully with Abseil requiring C++14 as minimum.

```
cmake -DABSL_BUILD_TESTING=ON -DABSL_USE_GOOGLETEST_HEAD=ON  ..
make -j 4
```
A check for C++14 being set at minimum is also added to the cmake file so the version failure is notified at an earlier stage when first running cmake.


Dragonfly5596:
cmake -DABSL_BUILD_TESTING=ON -DABSL_USE_GOOGLETEST_HEAD=ON  ..
make -j 44

infomaniac777:
Hello @Dragonfly5596 , can you please clarify your comment ? Is it to build check with the PR included ?

PR #1785, opened 2024-11-07
add github workflow action to build and install abseil with CMake

zacharyburnett:
adds an `action.yml` with the goal that a GitHub workflow could use Abseil to build their project by doing
```yaml
  - uses: abseil/abseil-cpp
    with:
      cmake-version: "X.X.X"
      cmake-cxx-standard: 14
```

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,25 @@
+name: build
+
+on:
+  pull_request:
+  push:
+    branches:
+      - master
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  action:
+    strategy:
+      matrix:
+        runs-on:
+          - ubuntu-latest
+          - macos-13
+          - macos-latest
+          - windows-latest
+    runs-on: ${{ matrix.runs-on }}
+    steps:
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+      - uses: ./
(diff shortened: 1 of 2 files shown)


derekmauro:
Thanks for your interest in Abseil.

Unfortunately we aren't going to take this pull request. Let me try to explain why.

Abseil's primary purpose is to be used as a dependency by Google's open source C++ projects. While we do hope that Abseil is also useful to the C++ community at large, this added constraint also means that we are unlikely to accept a contribution of utility code that isn't already widely used by Google. Most of the pull requests that we take from the community are simple portability fixes.

The other reason is that Abseil's source of truth is our internal code repository. It is exported to GitHub from there. We already have equivalent tests using our own build system. We aren't going to run two different sets of CI systems. Our internal contributors already have enough work maintaining one system, and we need the system that matters be the system that is triggered when changes are proposed to our internal repository. We can't burden internal contributors who do not know anything about GitHub with making sure GitHub Actions works as well.

zacharyburnett:
> Thanks for your interest in Abseil.
> 
> Unfortunately we aren't going to take this pull request. Let me try to explain why.
> 
> Abseil's primary purpose is to be used as a dependency by Google's open source C++ projects. While we do hope that Abseil is also useful to the C++ community at large, this added constraint also means that we are unlikely to accept a contribution of utility code that isn't already widely used by Google. Most of the pull requests that we take from the community are simple portability fixes.
> 
> The other reason is that Abseil's source of truth is our internal code repository. It is exported to GitHub from there. We already have equivalent tests using our own build system. We aren't going to run two different sets of CI systems. Our internal contributors already have enough work maintaining one system, and we need the system that matters be the system that is triggered when changes are proposed to our internal repository. We can't burden internal contributors who do not know anything about GitHub with making sure GitHub Actions works as well.

no worries, that makes sense

PR #1790, opened 2024-11-24
Respect the allocator's .destroy method in ~InlinedVector

Quuxplusone:
InlinedVector goes out of its way to use `A::construct` to construct new elements, so it should not skip the `A::destroy` call for those elements, either. Most codepaths seem fine (I didn't exhaustively explore this), but `DestroyAdapter` specifically failed to check whether the allocator had a non-trivial `destroy` method: it checked only whether the element type was trivially destructible in the absence of any `destroy` method.

Use the same condition in `DestroyAdapter` as we already use in the special short-circuit case on line 357 and in the conditions for the relocation optimizations on lines 304 and 324.

Note to Abseil maintainer: You'll want to make sure this passes all the tests, and probably add a newly-green test for it. Something like this Godbolt: https://godbolt.org/z/aoTar9vdT

```
template<class T>
struct A {
    int i_ = 0;
    using value_type = T;
    explicit A(int i) : i_(i) {}
    template<class U> A(A<U>& a) : i_(a.i_) {}
    T *allocate(size_t n) { return std::allocator<T>().allocate(n); }
    void deallocate(T *p, size_t n) { return std::allocator<T>().deallocate(p, n); }
    template<class... Args> T *construct(T *p, Args&&... args) { printf("constructing(%d) %s at %p\n", i_, __PRETTY_FUNCTION__, p); return ::new(p) T(std::forward<Args>(args)...); }
    void destroy(T *p) { printf("destroying(%d) %s at %p\n", i_, __PRETTY_FUNCTION__, p); }
};

int main() {
    absl::InlinedVector<int, 2, A<int>> v1(A<int>(42));
    v1.emplace_back(1);
    v1.insert(v1.begin(), 2);
    puts("Where are the corresponding destroy calls??");
}
```
(which currently prints)
```
constructing(42) T *A<int>::construct(T *, Args &&...) [T = int, Args = <int>] at 0x7ffc8b8fca80
constructing(42) T *A<int>::construct(T *, Args &&...) [T = int, Args = <int>] at 0x7ffc8b8fca84
Where are the corresponding destroy calls??
```

After this patch, it prints:
```
constructing(42) T *A<int>::construct(T *, Args &&...) [T = int, Args = <int>] at 0x16d9b7448
constructing(42) T *A<int>::construct(T *, Args &&...) [T = int, Args = <int>] at 0x16d9b744c
Where are the corresponding destroy calls??
destroying(42) void A<int>::destroy(T *) [T = int] at 0x16d9b744c
destroying(42) void A<int>::destroy(T *) [T = int] at 0x16d9b7448
```

### DIFF
--- a/absl/container/internal/inlined_vector.h
+++ b/absl/container/internal/inlined_vector.h
@@ -84,7 +84,8 @@ template <typename A>
 using IsSwapOk = absl::type_traits_internal::IsSwappable<ValueType<A>>;
 
 template <typename A, bool IsTriviallyDestructible =
-                          absl::is_trivially_destructible<ValueType<A>>::value>
+                          absl::is_trivially_destructible<ValueType<A>>::value &&
+                          std::is_same<A, std::allocator<ValueType<A>>>::value>
 struct DestroyAdapter;
 
 template <typename A>


Quuxplusone:
Notice that [there is a WG21 proposal for `std::inplace_vector<T, N, A>`](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2024/p3160r2.html), but since there is already an `inplace_vector<T, N>` in the working draft of C++26, the allocator-aware proposal must either make it into the C++26 cycle or be abandoned. You can track the proposal's progress [here](https://github.com/cplusplus/papers/issues/1809).

If Abseil is interested in having a standard `inplace_vector<T, N, A>` with the same semantics as `absl::InlinedVector<T, N, A>`, then this would be the right month to start making noise about that. Three months from now, it will be too late.

derekmauro:
Thanks and sorry for the delay on this one. There are a handful of users that use the allocator functionality so I'm currently verifying their tests still pass.

I was also able to modify an existing test so that it is newly green with this change:
```
TEST(AllocatorSupportTest, CountAllocations) {
  using MyAlloc = CountingAllocator<int>;
  using AllocVec = absl::InlinedVector<int, 4, MyAlloc>;
  const int ia[] = {0, 1, 2, 3, 4, 5, 6, 7};
  int64_t bytes_allocated = 0;
  int64_t instance_count = 0;
  MyAlloc alloc(&bytes_allocated, &instance_count);
  {
    AllocVec ABSL_ATTRIBUTE_UNUSED v(ia, ia + 4, alloc);
    EXPECT_THAT(bytes_allocated, Eq(0));
    EXPECT_THAT(instance_count, Eq(4));
  }
  EXPECT_THAT(bytes_allocated, Eq(0));
  EXPECT_THAT(instance_count, Eq(0));
  {
    AllocVec ABSL_ATTRIBUTE_UNUSED v(ia, ia + ABSL_ARRAYSIZE(ia), alloc);
    EXPECT_THAT(bytes_allocated,
                Eq(static_cast<int64_t>(v.size() * sizeof(int))));
    EXPECT_THAT(instance_count, Eq(static_cast<int64_t>(v.size())));
  }
  EXPECT_THAT(bytes_allocated, Eq(0));
  EXPECT_THAT(instance_count, Eq(0));
  {
    AllocVec v(4, 1, alloc);
    EXPECT_THAT(bytes_allocated, Eq(0));
    EXPECT_THAT(instance_count, Eq(4));

    int64_t bytes_allocated2 = 0;
    MyAlloc alloc2(&bytes_allocated2);
    ABSL_ATTRIBUTE_UNUSED AllocVec v2(v, alloc2);
    EXPECT_THAT(bytes_allocated2, Eq(0));

    int64_t bytes_allocated3 = 0;
    MyAlloc alloc3(&bytes_allocated3);
    ABSL_ATTRIBUTE_UNUSED AllocVec v3(std::move(v), alloc3);
    EXPECT_THAT(bytes_allocated3, Eq(0));
  }
  EXPECT_THAT(bytes_allocated, Eq(0));
  EXPECT_THAT(instance_count, Eq(0));
  {
    AllocVec v(8, 2, alloc);
    EXPECT_THAT(bytes_allocated,
                Eq(static_cast<int64_t>(v.size() * sizeof(int))));
    EXPECT_THAT(instance_count, Eq(static_cast<int64_t>(v.size())));

    int64_t bytes_allocated2 = 0;
    MyAlloc alloc2(&bytes_allocated2);
    AllocVec v2(v, alloc2);
    EXPECT_THAT(bytes_allocated2,
                Eq(static_cast<int64_t>(v2.size() * sizeof(int))));

    int64_t bytes_allocated3 = 0;
    MyAlloc alloc3(&bytes_allocated3);
    AllocVec v3(std::move(v), alloc3);
    EXPECT_THAT(bytes_allocated3,
                Eq(static_cast<int64_t>(v3.size() * sizeof(int))));
  }
  EXPECT_EQ(bytes_allocated, 0);
  EXPECT_EQ(instance_count, 0);
  {
    // Test shrink_to_fit deallocations.
    AllocVec v(8, 2, alloc);
    EXPECT_EQ(bytes_allocated, static_cast<int64_t>(8 * sizeof(int)));
    EXPECT_EQ(instance_count, 8);
    v.resize(5);
    EXPECT_EQ(bytes_allocated, static_cast<int64_t>(8 * sizeof(int)));
    EXPECT_EQ(instance_count, 5);
    v.shrink_to_fit();
    EXPECT_EQ(bytes_allocated, static_cast<int64_t>(5 * sizeof(int)));
    EXPECT_EQ(instance_count, 5);
    v.resize(4);
    EXPECT_EQ(bytes_allocated, static_cast<int64_t>(5 * sizeof(int)));
    EXPECT_EQ(instance_count, 4);
    v.shrink_to_fit();
    EXPECT_EQ(bytes_allocated, 0);
    EXPECT_EQ(instance_count, 4);
  }
  EXPECT_EQ(instance_count, 0);
}
```

Unfortunately I'm not sure anyone has cycles to engage with the standards proposal at this point.

higher-performance:
Just a minor note: I _think_ `std::allocator` can be specialized, in which case this wouldn't necessarily check if `destroy` is equivalent to a destructor call. But this is still an improvement over the status quo and there's no portable way to detect that AFAIK, so I don't see an issue.

Quuxplusone:
> I think std::allocator can be specialized, in which case this wouldn't necessarily check if destroy is equivalent to a destructor call. But this is still an improvement over the status quo and there's no portable way to detect that AFAIK, so I don't see an issue.

I think you're right re the paper standard, but even e.g. libc++ assumes the user never specializes `std::allocator`. I think if a user ever did that, it'd break the world.

@derekmauro "There are a handful of users that use the allocator functionality" — That sounds like you know "a handful of" people who would endorse `inplace_vector<T, N, A>`. Could you put me and/or Pablo Halpern in touch with them via email, please? There's no "engage with"; it's literally just finding people to say "yes, we use this feature in real life."

derekmauro:
@Quuxplusone I can probably gather the information you are looking for and say that there are users of this feature. Just let me know what information you need.

Issue #1760, opened 2024-09-19
[Bug]: combase.dll is loaded and unloaded in every call to `LocalTimeZone()` on Win10


daniel-eiband-snkeos:
### Describe the issue

When fetching the local time zone on Win10 (which happens by default if no time zone is given in calls to `FormatTime()`) combase.dll is loaded and the WinRT runtime is initialized for each call. This is very slow when profiling (we observed ~300ms per call in VTune) and and as a result makes profiling more or less unusable.

Even in scenarios other than profiling this seems to be excessive.

### Steps to reproduce the problem

We experienced this issue when profiling with VTune. Among others, the function `RoInitialize()` shows up as the most expensive function when doing a gRPC call. Even though the [abseil documentation](https://abseil.io/docs/cpp/guides/time#time-zones) says it is preferred to pass the time zone explicitly, in practice gRPC calls this function [here](https://github.com/grpc/grpc/blob/e0c6055d21b40ed0baccd38472f4c6d7acc517ea/src/core/lib/gprpp/status_helper.cc#L290) and [here](https://github.com/grpc/grpc/blob/e0c6055d21b40ed0baccd38472f4c6d7acc517ea/src/core/lib/security/authorization/stdout_logger.cc#L46). Also see: https://github.com/grpc/grpc/issues/37766

### What version of Abseil are you using?

20230802.1

### What operating system and version are you using?

The library was compiled on Windows Server 2022 and the program is executed on Windows 10 Enterprise 22H2.

### What compiler and version are you using?

MSVC build tools version 14.36.17.6

### What build system are you using?

CMake via Conan

malkia:
Was just going to post about this. While getting Heap Profiles, I found that (through) gRPC this function is called a lot

![image](https://github.com/user-attachments/assets/9b565651-9b4c-40f1-a702-075d0f936238)

malkia:
![image](https://github.com/user-attachments/assets/76b0129e-4ee8-4c84-809c-0e2dcc7923f1)

yukawa:
I finally get a time to rework google/cctz#242 and have sent out a fix to upstream cctz repository.

 * https://github.com/google/cctz/pull/316

Just fyi.

yukawa:
The above fix got imported from cctz to abseil-cpp at 9a89ea8714c06aee702058c6acc094bff02dfd42.
I think this can now be closed.

derekmauro:
Thank you, @yukawa!

PR #1791, opened 2024-11-25
Cache the local time zone calculation on Windows

malkia:
On Windows 10+, combase.dll gets loaded, unloaded, then subsequently various COM-related functions called to initialize/deinitialize the system leading to performance degradation https://github.com/abseil/abseil-cpp/issues/1760 

This "fix" here caches the value once. Obviously this won't work for long running apps, so ideally some kind of way knowing when to update the value are needed (like subscribing to time zone changes, or simply refreshing the value every once in a while).

In the mean time, this affects grpc for Windows where if grpc returns anything else than grpc::Status::OK it'll have to call this function. For processes running into grpc issues this means extra time added. It's also not a thing that normally would be caught, and I could not find benchmark for this case.

https://github.com/grpc/grpc/issues/37766

### DIFF
--- a/absl/time/internal/cctz/src/time_zone_lookup.cc
+++ b/absl/time/internal/cctz/src/time_zone_lookup.cc
@@ -67,7 +67,7 @@ namespace {
 #if defined(USE_WIN32_LOCAL_TIME_ZONE)
 // Calls the WinRT Calendar.GetTimeZone method to obtain the IANA ID of the
 // local time zone. Returns an empty vector in case of an error.
-std::string win32_local_time_zone(const HMODULE combase) {
+static std::string win32_local_time_zone(const HMODULE combase) {
   std::string result;
   const auto ro_activate_instance =
       reinterpret_cast<decltype(&RoActivateInstance)>(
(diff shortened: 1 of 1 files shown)


absl/time/internal/cctz/src/time_zone_lookup.cc
```diff
@@ -138,6 +138,43 @@ std::string win32_local_time_zone(const HMODULE combase) {
...
+        }
+      }
+    }
+    FreeLibrary(combase);
+  }
+  return winrt_tz;
+}
+
+static std::string& cached_win32_local_time_zone() {
+  static std::string cached_time_zone = get_once_win32_local_time_zone();
```
derekmauro, line 175: Note that this is not safe for global destruction. See https://google.github.io/styleguide/cppguide.html#Static_and_Global_Variables.

derekmauro (commented):
A couple things.

Changes to this code must be send to https://github.com/google/cctz. We automatically import this code from there.

However, I'm not sure the changes would be accepted. I am a maintainer of CCTZ but I usually defer to @devbww if he has an opinion on CCTZ code. In general though I think gRPC should be caching `absl::LocalTimeZone()` and passing that to `absl::FormatTime()`. As I think you understand from the text of this pull request, the local time zone can change while the program is running and caching would defeat that and may do something unexpected.

Also note that Abseil's logging library caches the local time: https://github.com/abseil/abseil-cpp/blob/c7cf999bda8390d2dd294ef903716a80135e6f4c/absl/log/initialize.cc

absl/time/internal/cctz/src/time_zone_lookup.cc
malkia: Good point! I wonder what else can be done here - maybe leaking the object? (Not ideal)

derekmauro: Yes that is what I would do, but I was just pointing that out. I think the approach here is wrong anyway. See above.

malkia:
Per @derekmauro 's comment - closing this as it needs to be really cached in `grpc` not "fix" it here (or in `cctz`).

devbww:
> I usually defer to @devbww if he has an opinion on CCTZ code. In general though I think gRPC should be caching `absl::LocalTimeZone()` and passing that to `absl::FormatTime()`.

I agree with Derek, and just made a comment to that effect on https://github.com/grpc/grpc/issues/37766, for what that's worth.

PR #1797, opened 2024-12-20
Remove bazel cpu constraints usage

keith:
This fixes a warning in an upcoming version of bazel. This use case
seems to solvable by using platform constraints directly. Fixes:

```
WARNING: @@abseil-cpp+//absl/random/internal:cpu_aarch64: select() on cpu is deprecated. Use platform constraints instead: https://bazel.build/docs/configurable-attributes#platforms.
```

### DIFF
--- a/absl/copts/configure_copts.bzl
+++ b/absl/copts/configure_copts.bzl
@@ -15,7 +15,6 @@ load(
     "ABSL_MSVC_FLAGS",
     "ABSL_MSVC_LINKOPTS",
     "ABSL_MSVC_TEST_FLAGS",
-    "ABSL_RANDOM_HWAES_ARM32_FLAGS",
     "ABSL_RANDOM_HWAES_ARM64_FLAGS",
     "ABSL_RANDOM_HWAES_MSVC_X64_FLAGS",
     "ABSL_RANDOM_HWAES_X64_FLAGS",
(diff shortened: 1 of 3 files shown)


absl/copts/configure_copts.bzl
```diff
@@ -45,38 +44,11 @@ ABSL_DEFAULT_LINKOPTS = select({
...
-    # APPLE
-    ":cpu_darwin_x86_64": ABSL_RANDOM_HWAES_X64_FLAGS,
-    ":cpu_darwin": ABSL_RANDOM_HWAES_X64_FLAGS,
-    ":cpu_x64_windows_msvc": ABSL_RANDOM_HWAES_MSVC_X64_FLAGS,
-    ":cpu_x64_windows": ABSL_RANDOM_HWAES_MSVC_X64_FLAGS,
-    ":cpu_k8": ABSL_RANDOM_HWAES_X64_FLAGS,
-    ":cpu_ppc": ["-mcrypto"],
-    ":cpu_aarch64": ABSL_RANDOM_HWAES_ARM64_FLAGS,
+    "@platforms//os:windows": ABSL_RANDOM_HWAES_MSVC_X64_FLAGS,
+    "@platforms//cpu:x86_64": ABSL_RANDOM_HWAES_X64_FLAGS,
```
derekmauro: I think my summary of this pull request is "thanks for trying, I don't think this works, but it does let us know we have a problem."

The problem is that `@platforms//os:windows` and `@platforms//cpu:x86_64` can both be true. The options controlled by this are really about the compiler (not the OS), and the CPU.

I'm actually thinking that this logic should not be in this file at all. Since for this code we use runtime dispatch when we are not sure at compile time if the CPU supports certain instructions, I'm thinking that using `[[target(foo)]]` in the source code might be a better solution.

cc: @laramiel

keith: > The problem is that @platforms//os:windows and @platforms//cpu:x86_64 can both be true.

Yea I was hoping CI could prove this out but I can't see the logs here. I've pushed a solution to that which uses `config_setting_group` + `match_all` to disambiguate (I don't have access to a windows machine to verify that it works)

keith:
thanks!

PR #1800, opened 2024-12-29
Make absl::from_chars_result compatible with C++20 and C++26

MBkkt:
https://en.cppreference.com/w/cpp/utility/from_chars_result


absl/strings/charconv.h
```diff
@@ -47,6 +47,10 @@ enum class chars_format {
 struct from_chars_result {
   absl::Nonnull<const char*> ptr;
   std::errc ec;
+
+  bool operator==(const from_chars_result&) = default;
```
derekmauro: This change doesn't work prior to C++20. You will have to implement the operator method.

```
/abseil-cpp/absl/strings/charconv.h:51:8: error: defaulted 'bool absl::from_chars_result::operator==(const absl::from_chars_result&)' only available with '-std=c++20' or '-std=gnu++20'
   51 |   bool operator==(const from_chars_result&) = default;
```

Also, I know this is a simple change, but it still needs test coverage.

derekmauro (changes requested):

absl/strings/charconv.h
MBkkt: @derekmauro Hi, I applied your suggestions to code

PR #1801, opened 2024-12-31
Remove unncessary <iostream> includes

jwnimmer-tri:
Including `<iostream>` means introducing the static (global) constructors and destructors for `std::cin`, `std::cerr`, and `std::cout`. That extra `init` and `fini` code is [undesirable](https://google.github.io/styleguide/cppguide.html#Static_and_Global_Variables) when those streams are not actually used.

### DIFF
--- a/absl/random/internal/randen_engine.h
+++ b/absl/random/internal/randen_engine.h
@@ -18,9 +18,10 @@
 #include <algorithm>
 #include <cinttypes>
 #include <cstdlib>
-#include <iostream>
+#include <istream>
 #include <iterator>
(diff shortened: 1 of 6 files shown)


derekmauro:
I've decided to accept a slightly modified version of this pull request (I've made the changes, no action necessary). The changes to `cord.cc` and `cord_rep_btree.cc` are slightly problematic for us (because there is code that isn't exported here so you can't see it), so I've reverted them. I think what I would like to do is change that code to output to a `std::string` and let the user dump it.

PR #1831, opened 2025-02-10
fix typo `iff`

LiAuTraver:


derekmauro:
This is not a typo. See https://en.wikipedia.org/wiki/If_and_only_if.

PR #1853, opened 2025-03-11
Add support for QCC compiler

pablorcum:
Since QCC (QNX C Compiler) is directly derived from the GNU C Compiler, same flags can be used.


abcd1424:
unsubscribe p

Issue #1866, opened 2025-04-03
[Bug]: Global static initializer RtlCaptureStackBackTrace_fn impacts startup


chrdavis:
### Describe the issue

RtlCaptureStackBackTrace_fn gets set (vig GetModuleHandle/GetProcAddress) at crt init and impacts app startup (in my case Chrome).  Since CaptureStackBackTrace points to RtlCaptureStackBackTrace since Vista can we just call CaptureStackBackTrace for desktop as well?  I doubt the loader lock comment is that impactful at runtime.

// It is not possible to load RtlCaptureStackBackTrace at static init time in
// UWP. CaptureStackBackTrace is the public version of RtlCaptureStackBackTrace
#if WINAPI_FAMILY_PARTITION(WINAPI_PARTITION_APP) && \
    !WINAPI_FAMILY_PARTITION(WINAPI_PARTITION_DESKTOP)
static RtlCaptureStackBackTrace_Function* const RtlCaptureStackBackTrace_fn =
    &::CaptureStackBackTrace;
#else
// Load the function we need at static init time, where we don't have
// to worry about someone else holding the loader's lock.
static RtlCaptureStackBackTrace_Function* const RtlCaptureStackBackTrace_fn =
    (RtlCaptureStackBackTrace_Function*)GetProcAddress(
        GetModuleHandleA("ntdll.dll"), "RtlCaptureStackBackTrace");
#endif  // WINAPI_PARTITION_APP && !WINAPI_PARTITION_DESKTOP

### Steps to reproduce the problem

Built chrome and examined static initializers in chrome.dll

### What version of Abseil are you using?

30.*

### What operating system and version are you using?

Windows 11

### What compiler and version are you using?

Clang 19.1.1

### What build system are you using?

cmake version 3.29.2

PR #1867, opened 2025-04-03
Remove global static in stacktrace_win32-inl.inc

chrdavis:
Removes the global static RtlCaptureStackBackTrace_Function and the GetModuleHandle/GetProcAddress calls so they do not impact module load at startup.


chrdavis:
Bug: #1866

PR #1872, opened 2025-04-14
hold original pointer along with aligned to get rid of possibly lost memory leak

TaeZStkyoht:
Unless pointers have been hold by an object, it's a lost on valgrind's logic. Even we don't free it, we must hold it. In this case, original pointers were not held.

Because gRPC uses this library, grpc usage was creating this memory leak.

before this change:
LEAK SUMMARY:
   definitely lost: 0 bytes in 0 blocks
   indirectly lost: 0 bytes in 0 blocks
     possibly lost: 1,408 bytes in 4 blocks
   still reachable: 1,408 bytes in 4 blocks
        suppressed: 0 bytes in 0 blocks

after this change:
LEAK SUMMARY:
   definitely lost: 0 bytes in 0 blocks
   indirectly lost: 0 bytes in 0 blocks
     possibly lost: 0 bytes in 0 blocks
   still reachable: 2,816 bytes in 8 blocks
        suppressed: 0 bytes in 0 blocks

Still reachable is not a problem for valgrind.

### DIFF
--- a/absl/random/internal/entropy_pool.cc
+++ b/absl/random/internal/entropy_pool.cc
@@ -96,9 +96,15 @@ void RandenPoolEntry::Fill(uint8_t* out, size_t bytes) {
 // Number of pooled urbg entries.
 static constexpr size_t kPoolSize = 8;
 
+// Struct to hold original and aligned pointers.
+struct AlignedPointer final {
+  char* original;            // Original pointer
+  RandenPoolEntry* aligned;  // Aligned pointer
+};
+
 // Shared pool entries.
 static absl::once_flag pool_once;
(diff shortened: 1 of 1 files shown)


TaeZStkyoht:
Why did "Windows MSVC 2022 C++17 Bazel (Optimized)" fail? I cannot see job logs. If I saw it, I would try to fix it.

derekmauro:
Thank you for letting us know about this issue. The failure is an infrastructure flake.

While this fix does work, I think using `alignas` is a simpler solution. I'm testing that now.

derekmauro:
https://github.com/abseil/abseil-cpp/commit/01228903e998c2be8069565c0f78d85d1c8a25cf should fix this issue.

PR #1880, opened 2025-05-01
Use constexpr in container algorithm

MBkkt:
https://github.com/llvm/llvm-project/blob/main/libcxx/include/algorithm
checked with llvm libc++19 and clang19 (only algorithms which support constexpr since c++20)

### DIFF
--- a/absl/algorithm/container.h
+++ b/absl/algorithm/container.h
@@ -522,7 +522,8 @@ ABSL_INTERNAL_CONSTEXPR_SINCE_CXX20
 // Container-based version of the <algorithm> `std::copy()` function to copy a
 // container's elements into an iterator.
 template <typename InputSequence, typename OutputIterator>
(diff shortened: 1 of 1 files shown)


derekmauro:
Interestingly this passes Abseil's tests but breaks almost our entire monorepo in certain cases where `operator<=>` is defined. I haven't had a chance to dig into why, and I'm not sure I will have time soon.

MBkkt:
> Interestingly this passes Abseil's tests but breaks almost our entire monorepo in certain cases where operator<=> is defined. I haven't had a chance to dig into why, and I'm not sure I will have time soon.

@derekmauro  Hi, can you share some example with compiler/stdlib/std version if possible?

derekmauro:
I'm pretty sure the problem is not the fault of your change. I think there are a handful of users that have non-compliant code that I will need to fix first.

MBkkt:
Ok, thanks for explaining

PR #1891, opened 2025-05-22
hash: fix ciso646 deprecated warning for gcc 15.1.1

cspiel1:
The deprecated warning for including <ciso646> breaks the build of
at least openvpn3 arch linux. With gcc 15.1.1 the pre-proc variables are

__cplusplus 201703
ABSL_INTERNAL_CPLUSPLUS_LANG 201703

Warning:
```
/usr/include/c++/15.1.1/ciso646:46:4: error: #warning "<ciso646> is deprecated in C++17, use <version> to detect implementation-specific macros" [-Werror=cpp]
   46 | #  warning "<ciso646> is deprecated in C++17, use <version> to detect implementation-specific macros"
```

Not sure if this is related or fixes also https://github.com/abseil/abseil-cpp/issues/1251.

### DIFF
--- a/absl/hash/internal/hash.h
+++ b/absl/hash/internal/hash.h
@@ -38,7 +38,7 @@
 #endif
 
 // For feature testing and determining which headers can be included.
-#if ABSL_INTERNAL_CPLUSPLUS_LANG >= 202002L || \
+#if ABSL_INTERNAL_CPLUSPLUS_LANG >= 2017003L || \
     ABSL_INTERNAL_VERSION_HEADER_AVAILABLE
 #include <version>
 #else


cspiel1:
I signed the CLA now. But currently have no time and desire to a "piper CL". So maybe somebody could do this. Or isn't it a correct solution?

derekmauro (commented):
What version of Abseil are you using? This should have already been fixed by the check on lines 31-38. Is it not working?

cspiel1:
Version: `abseil-cpp 20250127.1-2` which is the current version in Arch.

But the PR is bases on master HEAD of course. And I think that the check is not correct. Because for g++ 15.1.1 the 
```
__cplusplus = 201703
```
and thus
```
ABSL_INTERNAL_CPLUSPLUS_LANG = 201703
```
and `/usr/include/c++/15.1.1/ciso646` looks like this:
```c
#if __cplusplus >= 202002L && ! _GLIBCXX_USE_DEPRECATED
#  error "<ciso646> is not a standard header in C++20, use <version> to detect implementation-specific macros"
#elif __cplusplus >= 201703L && defined __DEPRECATED
#  pragma GCC diagnostic push
#  pragma GCC diagnostic ignored "-Wc++23-extensions"
#  warning "<ciso646> is deprecated in C++17, use <version> to detect implementation-specific macros"
#  pragma GCC diagnostic pop
#endif
```

This breaks the build for packages that are built with `-Werror`.

derekmauro:
I take it that you haven't actually tested your solution?

The full statement is 
```
#if ABSL_INTERNAL_CPLUSPLUS_LANG >= 202002L || \
    ABSL_INTERNAL_VERSION_HEADER_AVAILABLE
```

The second part of the or `ABSL_INTERNAL_VERSION_HEADER_AVAILABLE` should be true.

`<version>` is not a C++17 header, so it looks like your solution is incorrect.

Note that this fix is not present in abseil-cpp 20250127.1. The latest version should already work.

cspiel1:
Sorry, I tested without `ABSL_INTERNAL_VERSION_HEADER_AVAILABLE`. Didn't see your fix and the difference to the arch version.
Now I tested your fix. Of course it works. Sorry, for the confusion!

cspiel1:
I'll try to make an arch PR. Is it possible that you set a tag?

PR #1894, opened 2025-05-24
Fix usage of flat_hash_map with move only keys

gnusi:
This PR fixes compilation of `flat_hash_map` with move only keys, e.g. the following code won't compile:
```
absl::flat_hash_map<std::unique_ptr<int>, int> a;
absl::flat_hash_map<std::unique_ptr<int>, int> b;
b = std::move(a); // causes compilation error
```
The reason is that compiler tries to compile a wrong branch due to some values and `if`'s aren't marked with `constexpr`.

### DIFF
--- a/absl/container/internal/container_memory.h
+++ b/absl/container/internal/container_memory.h
@@ -389,7 +389,7 @@ struct map_slot_policy {
   template <class Allocator, class... Args>
   static void construct(Allocator* alloc, slot_type* slot, Args&&... args) {
     emplace(slot);
-    if (kMutableKeys::value) {
+    if constexpr (kMutableKeys::value) {
       absl::allocator_traits<Allocator>::construct(*alloc, &slot->mutable_value,
                                                    std::forward<Args>(args)...);
     } else {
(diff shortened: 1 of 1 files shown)


absl/container/internal/container_memory.h
```diff
@@ -438,29 +438,29 @@ struct map_slot_policy {
...
     // See https://github.com/llvm/llvm-project/pull/95444 for instance.
-    auto is_relocatable = typename std::conjunction<
+    static constexpr auto kIsRelocatable = typename std::conjunction<
         absl::is_trivially_relocatable<typename value_type::first_type>,
         absl::is_trivially_relocatable<typename value_type::second_type>>::
         type();
 
     emplace(new_slot);
-    if (is_relocatable) {
+    if constexpr (kIsRelocatable) {
```
derekmauro, line 447: This is now producing a warning:
```
./absl/container/internal/container_memory.h:434:35: error: parameter 'alloc' set but not used [-Werror=unused-but-set-parameter]
  434 |   static auto transfer(Allocator* alloc, slot_type* new_slot,
```

It seems that whatever analysis GCC is doing doesn't understand constexpr branches.

I'm guessing `[[maybe_unused]]` will fix this. Or perhaps `kIsRelocatable` could be a `bool` and the `std::conjunction` should use `::value` instead of `::type()`, maybe that would also give GCC a hint?

derekmauro (changes requested):
Could you also add a test case like the one in the example to prevent a regression?

absl/container/internal/container_memory.h
gnusi: Let me try, I'm going to also add a test case as requested.

PR #1910, opened 2025-07-25
Fix absl_demangle_test on ppc64

miladfarca:
We have recently started experimenting with using Clang to compile abseil (instead of gcc) on ppc64 and this test started failing:
```
# ./bin/absl_demangle_test
abseil-cpp/absl/debugging/internal/demangle_test.cc:1996: Failure
Expected: (stack_consumed) < (kStackConsumptionUpperLimit), actual: 9668 vs 8192

[  FAILED  ] Demangle.DemangleStackConsumption (5 ms)
```
This PR increases the value of `kStackConsumptionUpperLimit` to accommodate higher stack usage.

### DIFF
--- a/absl/debugging/internal/demangle_test.cc
+++ b/absl/debugging/internal/demangle_test.cc
@@ -1936,11 +1936,11 @@ static const char *DemangleStackConsumption(const char *mangled,
   return g_demangle_result;
 }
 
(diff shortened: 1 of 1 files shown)


miladfarca:
will reopen pr with another email address.

PR #1937, opened 2025-09-19
Simplify ConvertSpecialToEmptyAndFullToDeleted

klaren:
Take advantage of the fact that kEmpty is already the smallest signed char value, and using a saturated subtraction will not affect it.

All special values have the MSB set, so after an AND with MSBS, we are left with -128 for special values and 0 for full. After applying subs 2, we arrive at the result of -128(kEmpty) for special and -2(kDeleted) for full.

According to Intel specifications, the subs(PSUBSB) instruction has equal or better CPI and latency than shuffle(PSHUFB), so there is no need to keep the SSSE3 alternative.

### DIFF
--- a/absl/container/internal/hashtable_control_bytes.h
+++ b/absl/container/internal/hashtable_control_bytes.h
@@ -330,14 +330,8 @@ struct GroupSse2Impl {
 
   void ConvertSpecialToEmptyAndFullToDeleted(ctrl_t* dst) const {
     auto msbs = _mm_set1_epi8(static_cast<char>(-128));
-    auto x126 = _mm_set1_epi8(126);
-#ifdef ABSL_INTERNAL_HAVE_SSSE3
-    auto res = _mm_or_si128(_mm_shuffle_epi8(x126, ctrl), msbs);
-#else
-    auto zero = _mm_setzero_si128();
-    auto special_mask = _mm_cmpgt_epi8_fixed(zero, ctrl);
-    auto res = _mm_or_si128(msbs, _mm_andnot_si128(special_mask, x126));
-#endif
+    auto twos = _mm_set1_epi8(static_cast<char>(2));
+    auto res = _mm_subs_epi8(_mm_and_si128(msbs, ctrl), twos);
     _mm_storeu_si128(reinterpret_cast<__m128i*>(dst), res);
   }
 
--- a/absl/container/internal/hashtable_control_bytes.h
+++ b/absl/container/internal/hashtable_control_bytes.h
@@ -212,6 +212,9 @@ static_assert(
 static_assert(ctrl_t::kDeleted == static_cast<ctrl_t>(-2),
               "ctrl_t::kDeleted must be -2 to make the implementation of "
               "ConvertSpecialToEmptyAndFullToDeleted efficient");
+static_assert(ctrl_t::kEmpty == static_cast<ctrl_t>(-128),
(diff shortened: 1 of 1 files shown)


derekmauro:
How was this change tested? This change makes the microbenchmark `BM_DropDeletes` about 10% slower on my machine.

(@goldvitaly added the SSSE3 implementation in cl/164089769)

/cc @ezbr @goldvitaly

absl/container/internal/hashtable_control_bytes.h
```diff
@@ -330,14 +330,8 @@ struct GroupSse2Impl {
...
     auto msbs = _mm_set1_epi8(static_cast<char>(-128));
-    auto x126 = _mm_set1_epi8(126);
-#ifdef ABSL_INTERNAL_HAVE_SSSE3
-    auto res = _mm_or_si128(_mm_shuffle_epi8(x126, ctrl), msbs);
-#else
-    auto zero = _mm_setzero_si128();
-    auto special_mask = _mm_cmpgt_epi8_fixed(zero, ctrl);
-    auto res = _mm_or_si128(msbs, _mm_andnot_si128(special_mask, x126));
-#endif
+    auto twos = _mm_set1_epi8(static_cast<char>(2));
```
goldvitaly, line 342: Please, add to the line 225 another static_assert with a comment. We a collecting all the places we rely on specific values.

```
static_assert(
   ctrl_t::kEmpty == static_cast<ctrl_t>(-128),
   "ctrl_t::kEmpty must be -128 to use saturated subtraction in"
   " ConvertSpecialToEmptyAndFullToDeleted");
```

absl/container/internal/hashtable_control_bytes.h
```diff
@@ -330,14 +330,8 @@ struct GroupSse2Impl {
...
+    auto res = _mm_subs_epi8(_mm_and_si128(msbs, ctrl), twos);
```
goldvitaly, line 343: I know there was no comment before, but please, copy part of your commit description as a code comment.

goldvitaly:
> How was this change tested? This change makes the microbenchmark `BM_DropDeletes` about 10% slower on my machine.
> 
> (@goldvitaly added the SSSE3 implementation in cl/164089769)
> 
> /cc @ezbr @goldvitaly

Change looks very reasonable. On my machine it is ~0.5% faster.

https://godbolt.org/z/MYGrrbErz also suggest it is slightly faster with sse3 and 50% faster without sse3.

Given that it is simpler and faster I am in favor of that.

goldvitaly (commented):
Thanks. This looks simpler and I believe faster.

klaren:
> On my machine it is ~0.5% faster.

I couldn't measure any statistical difference on my machine, so that aligns with my findings.

I found this while porting the raw hash set to java using the new vector api, where shuffle is not available. I ended up using the equivalence of 
```c++
auto msbs = _mm_set1_epi8(static_cast<char>(-128));
auto twos = _mm_set1_epi8(static_cast<char>(-2));
auto res = _mm_min_epi8(_mm_and_si128(msbs, ctrl), twos);
```
which is ever so slightly faster, but `_mm_min_epi8` is sse4.1, so it's not as portable.

derekmauro:
> On my machine it is ~0.5% faster.

Sounds good, I could just have an outlier of a machine. I don't see any reason this should be slower.

PR #1939, opened 2025-09-20
Add missing rules_cc loads

keith:
This is required for use with bazel past this commit: https://github.com/bazelbuild/bazel/commit/71ca0ed111ff3d842a0d23bc3a46bd2e6745491d

Most files do this already these were just missing

### DIFF
--- a/absl/debugging/BUILD.bazel
+++ b/absl/debugging/BUILD.bazel
@@ -14,6 +14,9 @@
 # limitations under the License.
 #
 
+load("@rules_cc//cc:cc_binary.bzl", "cc_binary")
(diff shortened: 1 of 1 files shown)


derekmauro:
Thanks for keeping the Bazel ecosystem running! I kicked off the import procedure, it should be automatically imported by Monday sometime.

keith:
Thanks for the quick turnaround!

fmeum:
@derekmauro Could this be included in the next patch release?

derekmauro:
Yes. In fact I was planning on doing a patch release today or this week.

derekmauro:
I've released 20250814.1 containing this patch, and it is now in BCR as well.

PR #1943, opened 2025-09-25
Remove ABSL_MUST_USE_RESULT macro support only for clang

Malone-AI:
On line 454 of absl/base/attributes.h
```cpp
#if defined(__clang__) && ABSL_HAVE_ATTRIBUTE(warn_unused_result)
#define ABSL_MUST_USE_RESULT __attribute__((warn_unused_result))
#else
#define ABSL_MUST_USE_RESULT
#endif
```

ABSL_HAVE_ATTRIBUTE macro support only for clang, but this macro is also supported for g++

### DIFF
--- a/absl/base/attributes.h
+++ b/absl/base/attributes.h
@@ -451,7 +451,7 @@
 //
 // TODO(b/176172494): Use ABSL_HAVE_CPP_ATTRIBUTE(nodiscard) when all code is
 // compliant with the stricter [[nodiscard]].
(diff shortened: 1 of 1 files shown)


derekmauro:
If I recall correctly, there is some reason why we haven't done this. I think there was a lot of non-compliant code being built under GCC and no one got around to fixing it. There is also a relevant comment above:

https://github.com/abseil/abseil-cpp/blob/473a0c7530afb9ca5d107a64a3de46f84be757b5/absl/base/attributes.h#L422-L423

So is there still a good reason to persue this change?

PR #1968, opened 2025-11-03
Use std::move_backward within InlinedVector's Storage::Insert

bdash:
Both libc++ and libstdc++ have optimizations within `std::move_backward` that support lowering to `memmove` when it is appropriate for the types involved. Using it significantly speeds up insert on vectors of trivial types such as integers.

This completely addresses the [poor insert performance relative to other inline capacity vectors](https://github.com/martinus/svector#random-insert) that others have observed. See also https://github.com/abseil/abseil-cpp/discussions/1934.

### DIFF
--- a/absl/container/internal/inlined_vector.h
+++ b/absl/container/internal/inlined_vector.h
@@ -795,16 +795,9 @@ auto Storage<T, N, A>::Insert(ConstIterator<A> pos, ValueAdapter values,
                                    move_construction_values,
                                    move_construction.size());
 
-    for (Pointer<A>
-             destination = move_assignment.data() + move_assignment.size(),
-             last_destination = move_assignment.data(),
-             source = move_assignment_values + move_assignment.size();
-         ;) {
-      --destination;
-      --source;
-      if (destination < last_destination) break;
-      *destination = std::move(*source);
-    }
+    std::move_backward(move_assignment_values, 
+        move_assignment_values + move_assignment.size(),
+        move_assignment.data() + move_assignment.size());
 
     AssignElements<A>(insert_assignment.data(), values,
                       insert_assignment.size());


derekmauro:
I ran our benchmarks for this change [`//absl/containter:inlined_vector_benchmark`](https://github.com/abseil/abseil-cpp/blob/master/absl/container/inlined_vector_benchmark.cc) and the most relevant benchmark, `BM_InsertRange`, seems to show slightly worse performance.

Before this change:
```
-------------------------------------------------------------------------------------------------
Benchmark                                                       Time             CPU   Iterations
-------------------------------------------------------------------------------------------------
BM_InsertRange<TrivialType, kLargeSize, kLargeSize>          22.1 ns         22.1 ns     31695000
BM_InsertRange<TrivialType, kLargeSize, kSmallSize>          19.1 ns         19.1 ns     36663500
BM_InsertRange<TrivialType, kSmallSize, kLargeSize>          20.6 ns         20.7 ns     33790000
BM_InsertRange<TrivialType, kSmallSize, kSmallSize>          19.1 ns         19.1 ns     37114700
BM_InsertRange<NontrivialType, kLargeSize, kLargeSize>       34.8 ns         34.8 ns     20143500
BM_InsertRange<NontrivialType, kLargeSize, kSmallSize>       25.9 ns         25.9 ns     26989900
BM_InsertRange<NontrivialType, kSmallSize, kLargeSize>       29.8 ns         29.8 ns     23554900
BM_InsertRange<NontrivialType, kSmallSize, kSmallSize>       21.5 ns         21.5 ns     32569000

```

With this change:
```
-------------------------------------------------------------------------------------------------
Benchmark                                                       Time             CPU   Iterations
-------------------------------------------------------------------------------------------------
BM_InsertRange<TrivialType, kLargeSize, kLargeSize>          24.3 ns         24.4 ns     28610500
BM_InsertRange<TrivialType, kLargeSize, kSmallSize>          21.8 ns         21.8 ns     32299600
BM_InsertRange<TrivialType, kSmallSize, kLargeSize>          23.6 ns         23.7 ns     29154800
BM_InsertRange<TrivialType, kSmallSize, kSmallSize>          21.2 ns         21.2 ns     33401400
BM_InsertRange<NontrivialType, kLargeSize, kLargeSize>       35.6 ns         35.6 ns     19761600
BM_InsertRange<NontrivialType, kLargeSize, kSmallSize>       26.0 ns         26.1 ns     26798800
BM_InsertRange<NontrivialType, kSmallSize, kLargeSize>       29.3 ns         29.3 ns     24473100
BM_InsertRange<NontrivialType, kSmallSize, kSmallSize>       20.8 ns         20.8 ns     33148700
```

I think there is a strong possibility that our benchmark isn't great. Can you explain the benchmark results or fix the benchmarks somehow to be more representative of real-world performance?

bdash:
The benchmark looks to use an inline capacity of 4 elements, with the small and large vectors used in the test being 2 and 8 elements respectively. `BM_InsertRange` inserts elements at the midpoint of the destination vector so it only ever needs to  move 1 or 4 elements. It appears that for small numbers of elements the overhead from the call to `memmove` makes it slower than the loop. On my machine `memmove` becomes faster once the data being moved approaches 64 bytes. The gap widens as the size of the data being moved increases.

One option for this change would be to use `std::move_backward` only if `move_assignment.size() * sizeof(T)` is above some size threshold _and_ if the move assignment policy for `T` is `memcpy`. The latter condition should let the compiler elide the `std::move_backward` path entirely when the type is not trivially copyable, preventing an additional copy of the element-wise move assignment loop from being inlined into the function.

I don't have a sense of what representative use of `InlinedVector` looks like to know exactly what the benchmark should care about. In my usage it is often the case that the inline capacity covers 95% of usage, but in the other 5% of cases the number of elements may be significantly larger than the inline capacity. The benchmark testing with a maximum of 2x the inline capacity doesn't capture the performance characteristics for these larger sizes.

I think at a minimum the benchmarks should measure the performance with a size that is significantly larger than the inline capacity. This would make it possible to verify that it does not perform significantly worse than `std::vector` when it is in a state where it mostly behaves like one. Sadly adding additional size combinations seems a bit awkward with Google Benchmark due to the test functions being parameterized via template arguments.

derekmauro:
I was able to check the real world usage of `absl::InlinedVector::insert()` and it turns out that inserting into the middle of an `InlinedVector` just isn't something people do. It uses a trivial amount of CPU globally. I'm going to try accept this change as it seems to benefit most when it is used in the most pathological way.

PR #1971, opened 2025-11-12
Remove containers' dependency on cord library.

### LINKED ISSUE #1868: [Bug]: Cord causes slow builds all over project
### Describe the issue

While diagnosing my slow builds in a project that depends on Abseil, I concluded that Cord is the outstanding culprit. I don't use Cord, but I do use flags, and flags depend on flat_hash_map, and flat_hash_map comes with a header that _can_ hash a Cord, and thereby depends on Cord. It would be neat if there was a way to untie this knot so that only Cord users depended on Cord and its hasher.

According to `-ftime-trace` analysis these are the top three most expensive template instantiations for a small portion of my build that really should have been trivial but actually takes 55s to compile:

```
**** Templates that took longest to instantiate:
  3932 ms: std::deque<absl::crc_internal::CrcCordState::PrefixCrc>::operator= (31 times, avg 126 ms)
  3861 ms: std::deque<absl::crc_internal::CrcCordState::PrefixCrc>::assign<std:... (31 times, avg 124 ms)
  3829 ms: std::deque<absl::crc_internal::CrcCordState::PrefixCrc>::__assign_wi... (31 times, avg 123 ms)
```

Use clang's `-ftime-trace` flag in your project that needs Abseil-cpp flags, then analyze the output with ClangBuildAnalyzer.

HEAD

Ubuntu Linux 24.04

/opt/llvm-19/bin/clang -v
clang version 19.1.7
Target: aarch64-unknown-linux-gnu
Thread model: posix
InstalledDir: /opt/llvm-19/bin
Found candidate GCC installation: /usr/lib/gcc/aarch64-linux-gnu/11
Selected GCC installation: /usr/lib/gcc/aarch64-linux-gnu/11
Candidate multilib: .;@m64
Selected multilib: .;@m64

Bazel 7.6.1

### Additional context

These results are seen with libc++, if that makes a difference.

anpol:
Abseil associative containers having keys of a string type provide so-called heterogeneous lookup: they allow either `std::string`, `std::string_view`, or `absl::Cord` values in lookup.

On the other hand, using either of associative containers should not require compiling the cord library unless the use of `absl::Cord` is explicitly requested.

The solution is to forward-declare `absl::Cord` and a few functions just to support heterogeneous lookup, make the associative containers use these functions only, and require users of `absl::Cord` to link with the cord library explicitly.

fixes #1868

### DIFF
--- a/absl/container/internal/hash_function_defaults.h
+++ b/absl/container/internal/hash_function_defaults.h
@@ -56,7 +56,7 @@
 #include "absl/container/internal/common.h"
 #include "absl/hash/hash.h"
 #include "absl/meta/type_traits.h"
-#include "absl/strings/cord.h"
+#include "absl/strings/internal/cord_lookup_support.h"
 #include "absl/strings/string_view.h"
 
 namespace absl {
@@ -77,7 +77,7 @@ struct StringHash {
     return absl::Hash<absl::string_view>{}(v);
   }
   size_t operator()(const absl::Cord& v) const {
-    return absl::Hash<absl::Cord>{}(v);
+    return cord_internal::HashOfCord(v);
   }
 
  private:
@@ -88,8 +88,7 @@ struct StringHash {
         absl::Hash<absl::string_view>{}, v, seed);
   }
   size_t hash_with_seed(const absl::Cord& v, size_t seed) const {
-    return absl::hash_internal::HashWithSeed().hash(absl::Hash<absl::Cord>{}, v,
-                                                    seed);
+    return cord_internal::HashOfCordWithSeed(v, seed);
   }
 };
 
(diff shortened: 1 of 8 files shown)


absl/container/internal/hash_function_defaults.h
```diff
@@ -56,7 +56,7 @@
 #include "absl/container/internal/common.h"
 #include "absl/hash/hash.h"
 #include "absl/meta/type_traits.h"
-#include "absl/strings/cord.h"
+#include "absl/strings/internal/cord_lookup_support.h"
```
derekmauro, line 59: This doesn't work as is since the Bazel build now has an undeclared dependency on this header. The same goes for the other places where the dependency was removed.

derekmauro (changes requested):

PR #1976, opened 2025-11-23
fix(wasi): update wasi stubs

ricochet:
Today, WASI does not support atomics. Support for WASI in Abseil was added in #1509 and needs to be refreshed.

Note that there is a roadmap item for wasi that will give us access to atomics via the component model's [cooperative threads](https://github.com/WebAssembly/component-model/pull/557) feature. Implementations are in progress but this may take a few months and so in the mean time, we need this fix.

I created an example with the necessary stubs that are missing from upstream abseil and validated that these can be removed with the changes in this PR: https://github.com/ricochet/wasi-protobuf/blob/main/wasi-build/wasi_stubs.cpp

### DIFF
--- a/absl/base/internal/low_level_alloc.cc
+++ b/absl/base/internal/low_level_alloc.cc
@@ -659,3 +659,48 @@ ABSL_NAMESPACE_END
 }  // namespace absl
 
 #endif  // ABSL_LOW_LEVEL_ALLOC_MISSING
+
+// WASI stubs: WASI doesn't have mmap, so use malloc/free instead
+#if defined(__wasi__)
+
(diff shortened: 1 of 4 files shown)


derekmauro (changes requested):
It should be far simpler to achieve this. I believe all that needs to be done is a slight tweak here and everything else should just work:

https://github.com/abseil/abseil-cpp/blob/9d35bf52e774f231e3052fab34e93c2792f43755/absl/base/internal/low_level_alloc.h#L34-L41

ricochet:
Thank you that simplified it quite a bit 👍

PR #1632, opened 2024-02-29
inlined_vector: Use trivial relocation for `erase`

Quuxplusone:
Prior art for the `vector::erase` optimization:
https://github.com/AmadeusITGroup/amc/blob/efcb7be/include/amc/vectorcommon.hpp#L176-L180 https://github.com/bloomberg/bde/blob/e15f05be6/groups/bsl/bslalg/bslalg_arrayprimitives.h#L3787-L3799 https://github.com/facebook/folly/blob/d24bf04/folly/FBVector.h#L1254-L1262 https://github.com/qt/qtbase/blob/fbfee2d/src/corelib/tools/qarraydataops.h#L856-L861

(Opening a new PR as requested by @derekmauro in #1615. I would have reopened #1615, since this is the same branch and all, but I don't see a "reopen" button. Odd.)

PR #1985, opened 2025-12-19
Remove all attempts to use Clang builtins for trivial relocatability

### LINKED ISSUE #1885: [Bug]: builtin __is_trivially_relocatable is deprecated
### Describe the issue

On the newer (basically LLVM ToT branch) Clang compiler, abseil-cpp shows a lot of warnings, such as

```
../../third_party/abseil-cpp/absl/meta/type_traits.h:391:22: warning: builtin __is_trivially_relocatable is deprecated; use __builtin_is_cpp_trivially_relocatable instead [-Wdeprecated-builtins]
  391 |                     (__is_trivially_relocatable(T) &&
      |                      ^
../../third_party/libc++/src/include/__type_traits/conjunction.h:60:58: note: in instantiation of template class 'absl::is_trivially_relocatable<sk_sp<cc::PaintFilter>>' requested here
   60 | struct conjunction<_Arg, _Args...> : conditional_t<!bool(_Arg::value), _Arg, conjunction<_Args...>> {};
      |                                                          ^
../../third_party/abseil-cpp/absl/container/internal/inlined_vector.h:319:13: note: in instantiation of template class 'std::conjunction<absl::is_trivially_relocatable<sk_sp<cc::PaintFilter>>, std::is_same<std::allocator<sk_sp<cc::PaintFilter>>, std::allocator<sk_sp<cc::PaintFilter>>>>' requested here
  319 |       absl::conjunction<absl::is_trivially_relocatable<ValueType<A>>,
      |             ^
../../third_party/abseil-cpp/absl/container/inlined_vector.h:101:33: note: in instantiation of template class 'absl::inlined_vector_internal::Storage<sk_sp<cc::PaintFilter>, 2, std::allocator<sk_sp<cc::PaintFilter>>>' requested here
  101 |   using MemcpyPolicy = typename Storage::MemcpyPolicy;
      |                                 ^
../../cc/paint/paint_filter.h:596:46: note: in instantiation of template class 'absl::InlinedVector<sk_sp<cc::PaintFilter>, 2>' requested here
  596 |   absl::InlinedVector<sk_sp<PaintFilter>, 2> inputs_;
      |                                              ^
4 warnings generated.
```

This problem occurs when you compile Chromium from the ToT branch using a newer Clang compiler (i.e. a custom clang binary).

Compile Chromium from the ToT branch using a newer Clang compiler (i.e. a custom clang binary).

1b52dcb350289b262a105471a75ef6c001beecae

Debian Unstable

clang version 21.0.0git (https://chromium.googlesource.com/external/github.com/llvm/llvm-project 802d8d90776b476ca8f257ab2e4fa2db185c6b69)

Ninja

### Additional context

_No response_

Quuxplusone:
This simplifies the codepaths after a4950fba. We can't use Clang's current `__builtin_is_cpp_trivially_relocatable(T)` for the same reason we continue to be unable to use Clang's current `__is_trivially_relocatable(T)`: it returns true for polymorphic types and other types that aren't trivially relocatable in the Abseil/Folly/P1144 sense. We use memcpy for relocation; we use relocation to implement InlineVector::erase and InlineVector::swap. So, as the comment says, we care about wonky assignment as well as wonky copy-construction or destruction.

Use the public (not private/compiler-builtin!) API of P1144 wherever it's available. Never use the Clang builtins otherwise. Always fall back to the public (not private/compiler-builtin!) std::is_trivially_copyable when P1144 isn't available.

Remove references to P2786; it is not part of C++26 anymore, and even if it were, Abseil couldn't use it, because P2786 ignored move-assignment and also permitted "relocation" to fix up vptrs (which memcpy never does).

(Fixes #1885 in a different, shorter/correcter, way.)

Where P1144 is available, Abseil 20250127.0 has better performance than current Abseil trunk: i.e. trunk has regressed. This fixes that regression.

### DIFF
--- a/absl/meta/type_traits.h
+++ b/absl/meta/type_traits.h
@@ -347,16 +347,10 @@ using swap_internal::Swap;
 
 // absl::is_trivially_relocatable<T>
 //
-// https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2024/p2786r11.html
-//
 // Detects whether a type is known to be "trivially relocatable" -- meaning it
 // can be relocated from one place to another as if by memcpy/memmove.
 // This implies that its object representation doesn't depend on its address,
 // and also none of its special member functions do anything strange.
-//
-// Note that when relocating the caller code should ensure that if the object is
-// polymorphic, the dynamic type is of the most derived type. Padding bytes
-// should not be copied.
 //
 // This trait is conservative. If it's true then the type is definitely
 // trivially relocatable, but if it's false then the type may or may not be. For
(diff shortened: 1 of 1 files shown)


derekmauro:
I agree with removing the Clang builtins given the reported bugs. However, as far as I can tell, `std::is_trivially_relocatable` and `__cpp_lib_trivially_relocatable` are not yet standardized, I would prefer not to speculate on the final API shape in Abseil's mainline.

Could we please remove the `#if defined(__cpp_lib_trivially_relocatable)` block and stick strictly to the `std::is_trivially_copyable` fallback for now?

Quuxplusone:
> I agree with removing the Clang builtins given the reported bugs. However, as far as I can tell, `std::is_trivially_relocatable` and `__cpp_lib_trivially_relocatable` are not yet standardized, I would prefer not to speculate on the final API shape in Abseil's mainline.

`__cpp_lib_trivially_relocatable` is P1144's [feature-test macro](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2024/p1144r10.html#wording-version.syn). It, and the API of P1144, haven't changed since P1144 support was introduced into Abseil two years ago (in my own #1625) nor since `__is_trivially_relocatable(T)` was introduced into Abseil four years ago in b6a1039bfcd6163b923c84e23e11439f42c98bc7. I have often used Abseil as a good example of the kinds of performance wins you can get from trivial-relocatability; it's easy to see these wins on Godbolt (which uses Abseil 20250127.0, which predates a4950fba03217d0ea102c98ce795c189bba50383, so you can still see these wins today even though Abseil trunk has lost them).

For a test case, you can use the code from https://godbolt.org/z/xaa98qxhz , where Abseil version 20250127.0 used to use memmove inside `InlinedVector<unique_ptr<int>>::erase` on my P1144 reference implementation, but Abseil trunk (due to a4950fba03217d0ea102c98ce795c189bba50383) no longer uses memmove.

> Could we please remove the `#if defined(__cpp_lib_trivially_relocatable)` block and stick strictly to the `std::is_trivially_copyable` fallback for now?

That would be less buggy than the current post- a4950fba03217d0ea102c98ce795c189bba50383 status quo, but Abseil would still not be getting the memmove optimizations when compiled with a P1144-compatible compiler. Getting those optimizations has been the point of patches like #1625, #1632, #1618; it is a shame to break those optimizations now.

Using P1144's features _on systems that define them today_ isn't "speculating" — if C++29/32/... ever ships something different, or a second suitable-yet-differently-shaped API ever materializes, then Abseil could simply add an `#elif` branch conditioned on that other API's presence. In the meantime, re-enabling these optimizations (fixing a4950fba03217d0ea102c98ce795c189bba50383's brief regression versus Abseil 20250127.0) is appropriate IMO.

Note it's only these lines of code we're debating over:
```
#if defined(__cpp_lib_trivially_relocatable) // P1144
template <class T>
struct is_trivially_relocatable : std::is_trivially_relocatable<T> {
};
#else
```
Which is just, like, "if we have the feature then use it" (otherwise, emulate it).

derekmauro:
Regardless of the current compiler support or the performance benefits, trivial relocatability is not currently in the C++26 working draft, relying on this macro effectively treats a proposal as a standard feature. 

Even if P1144 is the, um, more useful approach, Abseil shouldn't claim support for a standard feature that doesn't legally exist yet. We should stick to the correct, non-speculative `std::is_trivially_copyable` baseline for now and revisit this in a couple of years, or whenever the feature is officially part of the standard.

Quuxplusone:
> relying on this macro effectively treats a proposal as a standard feature

I'd say it's a "non-standard extension" — that's why you have to place the code under the guard macro rather than just assume it's OK to use on all platforms. I don't see this as fundamentally different from e.g. [Abseil's conditional support for `gsl::Owner`](https://github.com/abseil/abseil-cpp/blob/master/absl/base/attributes.h#L972-L977).

However, while e.g. Folly and Parlay are happy to guard their usage directly with `__cpp_lib_trivially_relocatable`, there is precedent for using your own guard macro: [you can see here](https://github.com/STEllAR-GROUP/hpx/blob/e9739ca70da4ecff4bfdbaaa70926d0bdde75538/libs/core/type_support/include/hpx/type_support/uninitialized_relocation_primitives.hpp) that Ste||ar HPX used to test `__cpp_lib_trivially_relocatable` (that identifier remains in a comment) but has switched to testing `HPX_HAVE_P1144_RELOCATE_AT` instead. The downside is that someone has to `#define` that macro at build time; it's not just set automatically.

(1) Suppose we replace

    #if defined(__cpp_lib_trivially_relocatable) // P1144

with

    #if defined(ABSL_HAVE_P1144_LIBRARY_API)

and then add a comment somewhere, similar to how you do `ABSL_HAVE_ATTRIBUTE_TRIVIAL_ABI` today, indicating that this macro is not defined but people who want it _should_ define it?

(2) Or, again looking at how Abseil generally does this conditional-support/conditional-extension stuff — in Abseil's [attributes.h](https://github.com/abseil/abseil-cpp/blob/master/absl/base/attributes.h) — suppose you define something in there like

    #define ABSL_IS_TRIVIALLY_RELOCATABLE std::is_trivially_copyable

and then guard _that_ definition with compiler-support guards, e.g.

    #if __cpp_lib_trivially_relocatable // P1144
    #define ABSL_IS_TRIVIALLY_RELOCATABLE std::is_trivially_relocatable
    #else
    #define ABSL_IS_TRIVIALLY_RELOCATABLE std::is_trivially_copyable
    #endif

PR #1994, opened 2026-01-14
Implement ABSL_HASH_INTERNAL_CRC32_U64 for x86 builds

pps83:


pps83:
@goldvitaly ping

absl/hash/internal/hash.h
```diff
@@ -94,13 +94,16 @@
...
-#if defined(__SSE4_2__) && defined(__x86_64__)
+#if defined(__SSE4_2__)
 
 #include <x86intrin.h>
 #define ABSL_HASH_INTERNAL_HAS_CRC32
+#if defined(__x86_64__)
 #define ABSL_HASH_INTERNAL_CRC32_U64 _mm_crc32_u64
+#else
+#define ABSL_HASH_INTERNAL_CRC32_U64(crc, v) \
+  _mm_crc32_u32(_mm_crc32_u32((uint32_t)crc, (uint32_t)v), (uint32_t)(v >> 32))
```
goldvitaly, line 105: I afraid that double crc call will make it slower than alternative implementation. Although we do not consistently measure performance for 32 bit builds.

Did you run any benchmark?
Is there any additional motivation for this change?

pps83: I simply tried to build win32 build. it failed with errors trying to use _mm_crc32_u64 in 32builds. Unlikely alt impl could be faster than two 32bit crc calls. Most likely alt path also takes unoptimized path in 32bit builds.

absl/hash/internal/hash.h
```diff
@@ -94,13 +94,16 @@
...
+#endif
 #define ABSL_HASH_INTERNAL_CRC32_U32 _mm_crc32_u32
 #define ABSL_HASH_INTERNAL_CRC32_U8 _mm_crc32_u8
```
goldvitaly, line 109: Would it be enough to add `&& defined (_M_X64)` to the line below to fix win32 build?

I would prefer to go with such a simple change. There is an alternative implementation that is not using CRC all together. I also believe that we actually do not use CRC for 32 bits.

See 
```
CombineContiguousImpl(
    uint64_t state, const unsigned char* first, size_t len,
    std::integral_constant<int, 4> /* sizeof_size_t */)
```

pps83: I think CombineContiguousImpl thing has to use crc under the hood. Eventually, it falls back to c code if no crc32 hw is present. 
`&& defined (_M_X64)` fill fix vs with 32bit ms compiler, but I think it's better to simply emulate 64bit version with two 32bit calls.

@derekmauro would know better if it's better to exclude 32bit builds or emulate 64bit opcode with 2x32 bit crc

goldvitaly: 32bit version of `CombineContiguousImpl` doesn't use CRC for sure. So emulating `ABSL_HASH_INTERNAL_CRC32_U64` for 32 bit builds will be a dead code. I prefer to avoid that.

I have sent to @derekmauro a change inside of Google with just `&& defined (_M_X64)` and comment.
It will be instantly delivered to github once submitted.

derekmauro: @goldvitaly has the right idea to fallback to the non-CRC hash.

goldvitaly:
I hope that the following commit have fixed win32 build:
https://github.com/abseil/abseil-cpp/commit/5b086d78c4424ab03771e2eed85a356082d4c7ae

So I would propose to abandon this.
Thanks for reporting and helping to fix that!

PR #1996, opened 2026-01-18
Add flat_hash_map benchmarks

lihasgupta:
Adds benchmark suite for absl::flat_hash_map, the recommended default hash map that previously had no public benchmarks.

Benchmarks included:
- Basic operations (insert, lookup hit/miss) for int and string keys
- Load factor analysis at 25%, 50%, 75%, and 87% capacity
- Iteration performance at different load factors
- Reserve vs no-reserve comparison across multiple sizes

Key findings from benchmark results:
- reserve() provides 20-24% performance improvement for known sizes
- Load factor has minimal impact on lookup speed (~2.1-2.2M ops/sec)
- Higher load factors improve iteration by 30% (better cache locality)
- Failed lookups are 50% faster than successful lookups

Testing: All benchmarks compile and run successfully with Bazel.

### DIFF
--- a/absl/container/BUILD.bazel
+++ b/absl/container/BUILD.bazel
@@ -1350,6 +1350,22 @@ cc_binary(
     ],
 )
 
+cc_binary(
+    name = "flat_hash_map_benchmark",
+    testonly = True,
+    srcs = ["flat_hash_map_benchmark.cc"],
+    copts = ABSL_TEST_COPTS,
+    linkopts = ABSL_DEFAULT_LINKOPTS,
+    tags = ["benchmark"],
+    visibility = ["//visibility:private"],
+    deps = [
+        ":flat_hash_map",
+        "//absl/strings:str_format",
+        "//absl/strings:string_view",
+        "@google_benchmark//:benchmark_main",
+    ],
+)
+
 cc_library(
     name = "chunked_queue",
     srcs = ["internal/chunked_queue.h"],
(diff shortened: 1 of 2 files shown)


lihasgupta:
Hi team,

Polite follow-up on this PR. The benchmarks show some interesting findings on reserve() performance and load factor impacts.

Happy to make any changes or address feedback.

derekmauro:
Hi and thanks for your interest in Abseil.

We actually do have benchmarks for flat_hash_map, but they are a little hard to find.
https://github.com/abseil/abseil-cpp/blob/master/absl/container/internal/raw_hash_set_benchmark.cc
https://github.com/abseil/abseil-cpp/blob/master/absl/container/internal/raw_hash_set_probe_benchmark.cc

Since we spend so many cycles across our datacenters in this code, it is pretty critical that actual performance changes are validated on production workloads, not microbenchmarks. See https://abseil.io/fast/39.

lihasgupta:
> Hi and thanks for your interest in Abseil.
> 
> We actually do have benchmarks for flat_hash_map, but they are a little hard to find. https://github.com/abseil/abseil-cpp/blob/master/absl/container/internal/raw_hash_set_benchmark.cc https://github.com/abseil/abseil-cpp/blob/master/absl/container/internal/raw_hash_set_probe_benchmark.cc
> 
> Since we spend so many cycles across our datacenters in this code, it is pretty critical that actual performance changes are validated on production workloads, not microbenchmarks. See https://abseil.io/fast/39.

@derekmauro 

Thanks for the clarification! This makes total sense - I should have checked the internal/ directory more carefully.

I appreciate you taking the time to review. I'll close this PR.

PR #2000, opened 2026-01-22
Fix signed/unsigned mismatch warning with msvc

pps83:
> absl\time\internal\cctz\src\time_zone_name_win.cc(134,12): warning C4018: '>': signed/unsigned mismatch


derekmauro:
Pull requests for this code must be sent to https://github.com/google/cctz. They will be automatically imported here once accepted.

PR #2001, opened 2026-01-22
Switch to std::string::data() in Utf16ToUtf8 (C++17)

pps83:


derekmauro:
Pull requests for this code must be sent to https://github.com/google/cctz. They will be automatically imported here once accepted.

PR #2004, opened 2026-02-06
Construct {flat,node}_hash_{set,map} w/ from_range

victorvianna:
Adds constructors accepting std::from_range if building with C++23. See https://en.cppreference.com/w/cpp/ranges/from_range.html for context.


victorvianna:
I opened an internal change instead

PR #2012, opened 2026-02-11
Add support for CYGWIN

carlo-bramini:
This patch adds support for CYGWIN.
Hopefully, very few changes are needed.

1)
`absl/base/attributes.h`
Talking about shared libraries, CYGWIN shares the same behaviour of the traditional WIN32 binaries with DLLs.
So, exactly like MinGW, the compiler doesn't complain about the `weak` attribute until it fails at the link step.
Here, the solution is to avoid the `weak` attribute, exactly like it has been already done for the existing MinGW support.

2)
`absl/base/internal/low_level_alloc.h`
CYGWIN lacks support for low-level allocation of virtual memory, so here it follows again the same behaviour of WIN32.

3)
`absl/base/internal/sysinfo.cc`
`absl/base/internal/sysinfo.h`
On CYGWIN, the function `GetTID()` cannot be compiled because `pid_t` and the return value of `pthread_self()` are not compatible.
`pid_t` is an `int` type, so it's 32 bit on CYGWIN.
`pthread_self()` returns a `pthread_t` which is an opaque object.
And, on CYGWIN, `pthread_t` is a `typedef` to a pointer.
When compiling for x86_64, the size of the pointer is 64bit.
So, there is a mismatch because 32bit != 64bit.
Hopefully, it is already possible to support a different `pid_t` into `absl/base/internal/sysinfo.h`, so I did the same thing that it has been done for `_WIN32` and I added it to be an `uintptr_t`, which is good for all conditions.
However, into `absl/base/internal/sysinfo.cc`, the `static_cast` is not able to convert yet.
It can work if `pthread_self()` returns an `int`, but since `pthread_t` is an opaque object and it must be assumed to be like that, here the `reinterpret_cast` looks a better solution.

4)
`absl/base/policy_checks.h`
Since CYGWIN is supported after these changes, the lines for blocking the execution of the build process are not needed anymore and they have been removed from the code.

===
The library compiled fine in all conditions.
Tested build with both shared and static libraries.
Tested build with `ABSL_BUILD_MONOLITHIC_SHARED_LIBS` to be `ON` and `OFF`.

Sadly, I have no chance to execute bazel or Docker inside CYGWIN for running the tests because they seem to be not supported.
However, the chance that everything is working fine looks very high because CYGWIN is using code that already existed.
A direct test has been done with direct use, when building gstreamer-1.0, since abseil-cpp is a dependency.
If you think that this may be enough or something more can be done, please let me know.

I hope that you will find this patch useful.
Sincerely.

Carlo Bramini.

===


mkruskal-google:
Thanks for the PR, and it looks like a decent attempt.  But I think this is incomplete, and it sounds like maybe you didn't run tests?  We have the policy check because we don't [support](https://github.com/google/oss-policies-info/blob/main/foundational-cxx-support-matrix.md) Cygwin, don't test against it, and are *very* unlikely to work with it (much less so than other things we support "best-effort").

Even if we pushed these changes, they would probably regress in the near future, unexpectedly breaking people, and there could still be unknown runtime/behavior issues.  I strongly suspect this would hit serious problems in Cygwin even with these changes already though (notably in debugging, synchronization, flags, and threading code in base)

PR #2013, opened 2026-02-12
absl/hash: Fix ODR violation when building shared libraries

mochaaP:
Trivial change, fixes ASan warning about ODR violations.


genarorobles2314-cloud (commented):
24775f1

mochaaP:
not actually needed `inline` but probably needs a better way to suppress ASan warnings.

mochaaP:
https://github.com/google/sanitizers/issues/1017

Issue #2015, opened 2026-02-13
[Bug]: `absl::Base64Escape()` produces wrong output


T90REAL:
### Describe the issue

When calling the `absl::Base64Escape(absl::string_view src, std::string* dest)` with `src` backed by the same `std::string` as `*dest`, the function can:
- Produce an incorrect Base64 result (silent data corruption).
- Trigger a crash.

### Steps to reproduce the problem

1) Wrong output without sanitizers 
```cpp
#include <iostream>
#include <string>

#include "absl/strings/escaping.h"

int main() {
  std::string s = "hello";

  std::string out;
  absl::Base64Escape(s, &out);

  std::string inplace = s;
  absl::Base64Escape(inplace, &inplace);

  std::cout << "out_of_place=[" << out << "]\n";
  std::cout << "inplace=[" << inplace << "]\n";
  return 0;
}
```

```

out_of_place=[aGVsbG8=]
inplace=[aGVsc28=]
```

2) ASan heap-use-after-free crash 
```cpp
#include <string>
#include "absl/strings/escaping.h"

int main() {
  std::string s(1000, 'A');
  absl::Base64Escape(s, &s);  // src aliases dest
  return 0;
}
```
Build/run with AddressSanitizer, observed:
`AddressSanitizer: heap-use-after-free`
```
=================================================================
==53361==ERROR: AddressSanitizer: heap-use-after-free on address 0x619000001480 at pc 0x000100dd3284 bp 0x00016f9cdf50 sp 0x00016f9cd700
READ of size 4 at 0x619000001480 thread T0
    #0 0x000100dd3280 in __asan_memcpy+0x3f4 (libclang_rt.asan_osx_dynamic.dylib:arm64e+0x3b280)
    #1 0x00010047cbb4 in absl::base_internal::UnalignedLoad32(void const*)+0x1c (base64_escape_inplace_asan_repro:arm64+0x10004cbb4)
    #2 0x00010047cb38 in absl::big_endian::Load32(void const*)+0x14 (base64_escape_inplace_asan_repro:arm64+0x10004cb38)
    #3 0x00010047bc64 in absl::strings_internal::Base64EscapeInternal(unsigned char const*, unsigned long, char*, unsigned long, char const*, bool)+0xbc (base64_escape_inplace_asan_repro:arm64+0x10004bc64)
[...]

PR #2017, opened 2026-02-14
strings: fix data corruption in Base64Escape due to aliasing

Aspen-coder:
Description:
Addressed a data corruption bug in the pointer-based overloads of Base64Escape and WebSafeBase64Escape.
For a full description (as well as comments that led me to the solution), see: https://github.com/abseil/abseil-cpp/issues/2015

The Problem:
When the input string_view aliases the destination std::string, the internal implementation can _potentially_ overwrite the source data before the encoding is complete, leading to corrupted output / undefined behavior.

The Fix:
Modified the implementation to route these calls through the string-returning overloads.

Testing:
Added a regression test Base64Escape.AliasingSafety in absl/strings/escaping_test.cc. Verified passing with bazel test //absl/strings:escaping_test.

### DIFF
--- a/absl/strings/escaping.cc
+++ b/absl/strings/escaping.cc
@@ -954,17 +954,14 @@ bool WebSafeBase64Unescape(absl::string_view src,
   return Base64UnescapeInternal(src.data(), src.size(), dest, kUnWebSafeBase64);
 }
 
+
 void Base64Escape(absl::string_view src, std::string* absl_nonnull dest) {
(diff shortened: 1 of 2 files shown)


amtajkhan58-oss:
Hi

derekmauro:
Thanks for your interest in Abseil. A preferred fix for this issue has already been submitted: https://github.com/abseil/abseil-cpp/commit/586a3fbe2a985e8eaf8e54c5b80731a22af49ae1

PR #2020, opened 2026-02-24
Recognize `clang` targeting Windows as "MinGW like"

fmeum:
This is necessary to get abseil to build with https://github.com/cerisier/toolchains_llvm_bootstrapped. The old mappings for MinGW are arguably not semantically correct as MinGW is the runtime/sysroot, not the compiler, but are left in place for backwards compatibility.

### DIFF
--- a/absl/BUILD.bazel
+++ b/absl/BUILD.bazel
@@ -52,11 +52,23 @@ config_setting(
     visibility = [":__subpackages__"],
 )
 
+config_setting(
+    name = "windows-clang",
+    constraint_values = [
+        "@platforms//os:windows",
+    ],
+    flag_values = {
+        "@bazel_tools//tools/cpp:compiler": "clang",
+    },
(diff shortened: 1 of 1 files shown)


fmeum:
FYI @meteorcloudy

absl/BUILD.bazel
```diff
@@ -52,11 +52,23 @@ config_setting(
...
+    },
+    visibility = [":__subpackages__"],
+)
+
 selects.config_setting_group(
     name = "mingw_compiler",
     match_any = [
         ":mingw_unspecified_compiler",
         ":mingw-gcc_compiler",
+        ":windows-clang",
```
derekmauro, line 71: The seems to be a hack. In some cases hacks are fine, but there is no comment or any trace for a future reader that explains why this strange line is appearing here. Can you explain precisely what the problem is? Would renaming this group make it more clear what its purpose is?

PR #2046, opened 2026-05-06
Make absl::InitializeLog() safe to call more than once (#1656)

### LINKED ISSUE #1656: [Bug]: Make absl::InitializeLog safe to be called multiple times
### Describe the issue

I don't really use `absl::log`, but a library I depend on uses it. It logs a bunch of "garbage" (unrelated to my project) that appears on the stderr. It includes the message:

```
WARNING: All log messages before absl::InitializeLog() is called are written to STDERR
```

Now I'm adding an earlier call in my program to `absl::InitializeLog()` in an attempt to eliminate the noise. But I noticed the library I use also eventually calls `absl::InitializeLog()`. 

Checking the documentation of `absl::InitializeLog` it's an error to call it more than once. 

More generically, a project  and one or more of its dependencies may need to call `absl::InitializeLog`, and there is no way to coordinate this across completely independent libraries -- ideally, libraries should never call it but ... it happens -- see [Google's XLA](https://github.com/openxla/xla/blob/6a257dc8805d11c11d25de557dadcb4e248d0849/xla/pjrt/c/pjrt_c_api_gpu.cc#L27C1-L29C1) as an example.

Hence my bug/feature request to make `InitializeLog` safe to be called multiple times -- a no-op if it has already been called.

ps.: I'm no expert on `absl` library, it's very likely there is nuance I'm missing.

Use two libraries where both call `absl::InitializeLog()`

Head.

Ubuntu, various versions.

clang

bazel

### Additional context

_No response_

Othmane-Ch:
Fixes #1656.

Previously, calling `absl::InitializeLog()` more than once aborted the program: the underlying `absl::log_internal::SetTimeZone()` fires `ABSL_RAW_LOG(FATAL, "...has already been called")` on its second invocation. The header documented this as "an error to call this function twice", but in real-world systems multiple independent libraries each want to bring up the logging library and have no way to coordinate among themselves — Google's XLA project being the example cited in #1656.

## Change

Wrap the body of `InitializeLog()` in `absl::call_once`. Exactly the first invocation runs the underlying initialization; all subsequent calls (including concurrent ones from other threads) become no-ops. The first caller's `absl::LocalTimeZone()` value wins, matching the documented "only the first call has effect" intent — just made safe instead of UB.

Net diff: +106 / −2 across 5 files.

- `absl/log/initialize.cc` — `absl::call_once` guard around `InitializeLogImpl(absl::LocalTimeZone())`.
- `absl/log/initialize.h` — docstring updated to document the new contract (safe to call repeatedly; concurrent callers block until the first call completes).
- `absl/log/initialize_test.cc` — new gtest covering: first call initializes, second call is a no-op, concurrent calls from N threads are safe.
- `absl/log/BUILD.bazel`, `absl/log/CMakeLists.txt` — wire the new test target and add the `//absl/base` (CMake: `absl::base`) dep to the `log_initialize` library for `call_once`.

## Testing

Built and ran the new test locally with CMake/Ninja + MSVC 19.36 on Windows:

```
[==========] 3 tests from InitializeLogTest
[ RUN      ] InitializeLogTest.FirstCallInitializes      OK (0 ms)
[ RUN      ] InitializeLogTest.SecondCallIsNoOp          OK (0 ms)
[ RUN      ] InitializeLogTest.ConcurrentCallsAreSafe    OK (1 ms)
[==========] 3 tests from 1 test suite ran. (2 ms total)
[  PASSED  ] 3 tests.
```

Sibling regression check: `absl_log_globals_test` also passes after the BUILD/CMake edits.

Without the `call_once` guard, `ConcurrentCallsAreSafe` either races on the initialization globals or aborts via the FATAL raw log inside `SetTimeZone()`, so it doubles as a regression marker.

## Notes

- No public API surface change — same function name, same signature.
- The cost on every `InitializeLog()` call is one `absl::call_once` flag check, which after the first call is a single relaxed atomic load on the fast path.
- Out of scope: adding a `ShutdownLog()` / re-init capability (the issue does not request it)

### DIFF
--- a/absl/log/BUILD.bazel
+++ b/absl/log/BUILD.bazel
@@ -137,6 +137,7 @@ cc_library(
     visibility = ["//visibility:public"],
     deps = [
         ":globals",
+        "//absl/base",
         "//absl/base:config",
         "//absl/log/internal:globals",
         "//absl/time",
@@ -425,6 +426,20 @@ cc_test(
         "//absl/base:log_severity",
         "//absl/log/internal:globals",
         "//absl/log/internal:test_helpers",
+        "@googletest//:gtest",
+        "@googletest//:gtest_main",
+    ],
+)
+
+cc_test(
+    name = "initialize_test",
+    size = "small",
+    srcs = ["initialize_test.cc"],
+    copts = ABSL_TEST_COPTS,
+    linkopts = ABSL_DEFAULT_LINKOPTS,
+    deps = [
+        ":initialize",
+        "//absl/log/internal:globals",
         "@googletest//:gtest",
         "@googletest//:gtest_main",
     ],
--- a/absl/log/CMakeLists.txt
+++ b/absl/log/CMakeLists.txt
@@ -537,6 +537,7 @@ absl_cc_library(
   LINKOPTS
     ${ABSL_DEFAULT_LINKOPTS}
   DEPS
+    absl::base
     absl::config
     absl::log_globals
     absl::log_internal_globals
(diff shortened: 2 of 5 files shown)


rmjoona27-create:
- Othmane envió una solicitud de pull para hacer que
   absl::InitializeLog() se pueda llamar de forma segura más de una vez.
   - Google-cla[bot] indicó que Othmane debe firmar un Contributor License
   Agreement (CLA) antes de revisar la solicitud del Ministerio .

PR #2048, opened 2026-05-08
Allow WASM targets to pre-define ABSL_HAVE_MMAP

andreaTP:
Hi :wave: !
Thank you so much for maintaining Abseil, it's an amazing library and we really appreciate all the work that goes into it!

This is a tiny follow-up to #1976 . When building Abseil for WASM/WASI with wasi-sdk, build systems need to pass `-DABSL_HAVE_MMAP` alongside `-D_WASI_EMULATED_MMAN`. The #error guard on `ABSL_HAVE_MMAP` prevents this.

This PR exempts `__wasm__` targets from that guard (1 line changed), while keeping it intact for all other platforms. It also adds a CI script that builds Abseil with `wasi-sdk` to prevent regressions, I can remove it and leave the single line change when requested.

This patch (or equivalent) is already working in the wild across several projects, for example:

- [roastedroot/protobuf4j](https://github.com/roastedroot/protobuf4j/blob/main/buildtools-v3/patch-absl.txt)
- [wasilibs/go-protoc-gen-grpc-java](https://github.com/wasilibs/go-protoc-gen-grpc-java/blob/main/buildtools/wasm/patch-absl.txt)
- [wasilibs/go-protoc-gen-builtins](https://github.com/wasilibs/go-protoc-gen-builtins/blob/main/buildtools-descriptors/wasm/patch-absl.txt)

Having this upstream would let all these projects drop their local patches. Thank you for considering it!

### DIFF
--- a/absl/base/config.h
+++ b/absl/base/config.h
@@ -372,7 +372,7 @@ static_assert(ABSL_INTERNAL_INLINE_NAMESPACE_STR[0] != 'h' ||
 //
 // Checks whether the platform has an mmap(2) implementation as defined in
 // POSIX.1-2001.
-#ifdef ABSL_HAVE_MMAP
+#if defined(ABSL_HAVE_MMAP) && !defined(__wasm__)
 #error ABSL_HAVE_MMAP cannot be directly set
 #elif defined(__linux__) || defined(__APPLE__) || defined(__FreeBSD__) || \
     defined(_AIX) || defined(__ros__) || defined(__asmjs__) ||            \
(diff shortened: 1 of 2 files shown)


derekmauro:
Why is `-DABSL_HAVE_MMAP` needed if `-D_WASI_EMULATED_MMAN` will cause it to be defined?

andreaTP:
Thanks @derekmauro! I hadn't noticed `_WASI_EMULATED_MMAN` was already in the detection chain.
That covers my use case, closing this. Thanks for your time!

PR #2057, opened 2026-05-21
doc: clarify moved-from state for hash containers

### LINKED ISSUE #1928: [Bug]: Could you document something about the state of moved-from hash maps?
### Describe the issue

I couldn't find any documentation as to which methods, if any, are safe to call on a moved-from flat_hash_map.

I assume Abseil follows the STL convention that "moved-from objects shall be placed in a valid but unspecified state".

Does it mean that, at a minimum, `clear()` on such a container is safe to call and will work as expected (resulting in an empty container)? Or perhaps it's only safe to call `operator=(/* empty map */T{})`?

Empirically, it seems that calling `size()` is not safe, at least in ASAN'd code: 

https://godbolt.org/z/MTos6Kn5r

This is a documentation problem.

Example uses 20250127, but the documentation question applies to all recent versions.

N/A

N/A

N/A

### Additional context

_No response_

DrishtiTripathi2230:
Adds documentation to flat_hash_map, flat_hash_set, node_hash_map, and 
node_hash_set clarifying the moved-from contract for Swiss table containers.

After a move, only destruction, assignment, and clear() are guaranteed 
safe. Any other operation (e.g. size(), empty(), iteration) results in 
undefined behavior. Sanitizer builds actively catch such misuse via 
sentinel capacity values (kMovedFrom/kSelfMovedFrom).

This was confirmed by a maintainer in the linked issue.

Fixes #1928

### DIFF
--- a/absl/container/flat_hash_map.h
+++ b/absl/container/flat_hash_map.h
@@ -174,8 +174,7 @@ class ABSL_ATTRIBUTE_OWNER flat_hash_map
   //   // After the move, map4 is in a valid but unspecified state. The only
   //   // operations guaranteed to be safe on a moved-from map are destruction,
   //   // assignment, and clear(). Any other operation (e.g. size(), empty(),
-  //   // iteration) results in undefined behavior. Sanitizer builds will
-  //   // actively catch such misuse.
+  //   // iteration) results in undefined behavior.
   //
   // * Move assignment operator
   //
--- a/absl/container/flat_hash_map.h
+++ b/absl/container/flat_hash_map.h
@@ -171,11 +171,19 @@ class ABSL_ATTRIBUTE_OWNER flat_hash_map
   //   // Move is guaranteed efficient
   //   absl::flat_hash_map<int, std::string> map5(std::move(map4));
   //
+  //   // After the move, map4 is in a valid but unspecified state. The only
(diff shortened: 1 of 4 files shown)


absl/container/flat_hash_map.h
```diff
@@ -171,12 +171,20 @@ class ABSL_ATTRIBUTE_OWNER flat_hash_map
   //   // Move is guaranteed efficient
   //   absl::flat_hash_map<int, std::string> map5(std::move(map4));
   //
+  //   // After the move, map4 is in a valid but unspecified state. The only
+  //   // operations guaranteed to be safe on a moved-from map are destruction,
+  //   // assignment, and clear(). Any other operation (e.g. size(), empty(),
+  //   // iteration) results in undefined behavior. Sanitizer builds will
```
derekmauro: Please remove the part about sanitizer builds. The API contract should focus on defining correct usage and preconditions. If a user violates these preconditions, the behavior is undefined. Specifying which debug tools will catch that violation is an unnecessarily strong guarantee that leaks implementation details.

DrishtiTripathi2230: Done! Removed the sanitizer sentence from all 4 files.

PR #2059, opened 2026-05-23
Allow string_view as default value for string flags

### LINKED ISSUE #1540: [Bug]: `ABSL_FLAG(std::string)` does not accept `std::string_view` as a default value
### Describe the issue

When defining a flag of `std::string` type, the default value must be an `std::string` or `const char*`. It would be great if `std::string_view` were also accepted.

Following Google conventions, I frequently have string constants in the form:

```
constexpr std::string_view kMyStringConstant = "asdf";
```

which I tend to prefer over `constexpr const char*` because `std::string_view` has a `constexpr` constructor that allows explicit zeroes in strings.

The following does not compile:

```
constexpr std::string_view kMyStringConstant = "asdf";

ABSL_FLAG(std::string, my_flag_name, kMyStringConstant, "my message");
```

Of course, I can work around this with:

```
ABSL_FLAG(std::string, my_flag_name, std::string(kMyStringConstant), "my message");
```

[Pinned to commit 74eee2aff683cc7dcd2dbaa69b2c654596d8024e](https://github.com/abseil/abseil-cpp/commit/74eee2aff683cc7dcd2dbaa69b2c654596d8024e)

MacOS Ventura 13.5.2

```
$ clang -v
Apple clang version 14.0.3 (clang-1403.0.22.14.1)
Target: arm64-apple-darwin22.6.0
Thread model: posix
InstalledDir: /Applications/Xcode_14.3.1.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin
```

bazel 6.1.1

### Additional context

_No response_

DrishtiTripathi2230:
Fixes #1540

Allow std::string_view to be used as default value for ABSL_FLAG(std::string, ...).

Changed InitDefaultValue template to accept any convertible type U,
so that string_view is implicitly converted to string when used as
a default value.

### DIFF
--- a/absl/flags/internal/flag.h
+++ b/absl/flags/internal/flag.h
@@ -272,14 +272,20 @@ struct FlagDefaultArg {
 // TODO(rogeeff): Fix handling types with explicit constructors.
 struct EmptyBraces {};
 
-template <typename T>
-constexpr T InitDefaultValue(T t) {
-  return t;
+template <typename T, typename U = T>
+constexpr T InitDefaultValue(U t) {
+  return T(t);
 }
 
 template <typename T>
 constexpr T InitDefaultValue(EmptyBraces) {
   return T{};
+}
+
+
+// Allow std::string_view to be used as default value for std::string flags
+inline std::string InitDefaultValue(absl::string_view sv) {
+  return std::string(sv);
 }
 
 template <typename ValueT, typename GenT,
--- a/absl/flags/flag_test.cc
+++ b/absl/flags/flag_test.cc
@@ -294,6 +294,10 @@ ABSL_FLAG(uint64_t, test_flag_08, 9876543, "test flag 08");
 ABSL_FLAG(double, test_flag_09, -9.876e-50, "test flag 09");
 ABSL_FLAG(float, test_flag_10, 1.234e12f, "test flag 10");
 ABSL_FLAG(std::string, test_flag_11, "", "test flag 11");
+
+constexpr absl::string_view kStringViewDefault = "asdf";
+ABSL_FLAG(std::string, test_flag_sv_default, kStringViewDefault, "test flag with string_view default");
+
 ABSL_FLAG(absl::Duration, test_flag_12, absl::Minutes(10), "test flag 12");
 ABSL_FLAG(absl::int128, test_flag_13, absl::MakeInt128(-1, 0), "test flag 13");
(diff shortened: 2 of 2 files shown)


absl/flags/internal/flag.h
```diff
@@ -272,16 +272,22 @@ struct FlagDefaultArg {
 // TODO(rogeeff): Fix handling types with explicit constructors.
 struct EmptyBraces {};
 
-template <typename T>
-constexpr T InitDefaultValue(T t) {
-  return t;
+template <typename T, typename U = T>
+constexpr T InitDefaultValue(U t) {
```
derekmauro, line 276: By changing `InitDefaultValue()` to use a functional cast, the macro now allows any default value that is constructible to T, even if that constructor is marked explicit. Previously, the default value had to be implicitly convertible to T.

For example, `ABSL_FLAG(std::vector<int>, f, 10, "")` will now compile and create a vector of size 10. Previously, this would fail because `int` does not implicitly convert to `std::vector`.

absl/flags/internal/flag.h
```diff
@@ -272,16 +272,22 @@ struct FlagDefaultArg {
...
 }
 
 template <typename T>
 constexpr T InitDefaultValue(EmptyBraces) {
   return T{};
 }
 
+
+// Allow std::string_view to be used as default value for std::string flags
+inline std::string InitDefaultValue(absl::string_view sv) {
```
derekmauro, line 287: I think this has no effect because the `ABSL_FLAG` macro invokes `InitDefaultValue` with explicit template arguments. The functionality currently works only because the template catch-all allows the explicit conversion from `string_view` to `string`.

derekmauro (commented):
So I'm reading what I wrote a few years ago where I said I would consider accepting something like this, but seeing this code is making me think harder about it, and it is making me realize this is a bad idea.

Yes, this code does save a few key strokes. It also unintentionally allows explicit conversions implicitly. We probably could fix that and only allow the implicit conversion for `string_view`, but what do we gain from it? Saving a few keystrokes is a poor reason to add a convenience wrapper. When you read explicit, verbose C++, you can see exactly what is happening. The justification for a convenience wrapper should be to reduce complexity, not verbosity. We should introduce a utility only when the underlying operation is intrinsically hard to get right.

In short, I think it would be better to use an explicit conversion when you create your flag.

DrishtiTripathi2230:
I understand now that this is a bad approach. Thanks for explaining how it can introduce unintended implicit conversions. I'll close this PR.

PR #2060, opened 2026-05-24
Restrict in memory mmap arguments to glibc on s390x

jogemu:
DirectMmap fails for musl, because passing mmap arguments in memory is glibc specific.

### DIFF
--- a/absl/base/internal/direct_mmap.h
+++ b/absl/base/internal/direct_mmap.h
@@ -104,7 +104,7 @@ inline void* DirectMmap(void* start, size_t length, int prot, int flags, int fd,
       syscall(SYS_mmap2, start, length, prot, flags, fd,
               static_cast<unsigned long>(offset / pagesize)));  // NOLINT
 #endif
-#elif defined(__s390x__)
+#elif defined(__s390x__) && defined(__GLIBC__)
   // On s390x, mmap() arguments are passed in memory.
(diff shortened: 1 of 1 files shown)


mkruskal-google:
What about this code fails for musl?  I'm not seeing something glibc-specific here

jogemu:
`low_level_alloc.cc` uses DirectMmap during runtime

https://github.com/abseil/abseil-cpp/blob/3e0d9ff64abd250e4b78e3165745cfa296c1ee17/absl/base/internal/low_level_alloc.cc#L588-L589

which returns `MAP_FAILED` with `EINVAL`

https://github.com/abseil/abseil-cpp/blob/3e0d9ff64abd250e4b78e3165745cfa296c1ee17/absl/base/internal/low_level_alloc.cc#L598-L599

`[low_level_alloc.cc : 599] RAW: mmap error: 22` occurred during nodejs snapshot generation, which uses an older version that is permalinked above.

In musl `mmap` syscalls behave consistently with other 64-bit architectures, due to this condition: `if (n == SYS_mmap) return __syscall1(n, (long)(long[]){a,b,c,d,e,f});` https://git.musl-libc.org/cgit/musl/tree/arch/s390x/syscall_arch.h#n64

Just for the `s390x` architecture, `glibc` still uses `old_mmap` to maintain backward compatibility.

PR #2061, opened 2026-05-25
cast to unsigned char before std::isspace in ParseTime

dxbjavid:
std::isspace takes a value representable as unsigned char or EOF; sv->front() can yield a negative char when the input starts with a high-bit-set byte.


derekmauro:
Thanks for this report. I used a different solution consistent with Abseil code and added a test case.

PR #2064, opened 2026-05-28
docs: document MSVC compile-time format checking limitation in str_format.h

### LINKED ISSUE #1478: [Bug]: Compile-time string format checking not working with MSVC
### Describe the issue

The compile-time string format checking of `absl::StrFormat` doesn't work when compiling with MSVC.

I've looked through the docs of `StrFormat` and the C++ support matrix and couldn't find any mention of this. That's a pretty *big* footnote to leave out. And since the implementation isn't standards-compliant and instead defers to compiler extensions, it doesn't look like this is gonna change anytime soon.

`absl::StrFormat("%s", 1);` compiles instead of erroring.

lts_2023_01_25

Windows

cl 19.36.32534

Doesn't matter.

### Additional context

_No response_

DrishtiTripathi2230:
Fixes #1478

absl::StrFormat compile-time format string checking relies on
__attribute__((format(...))), a GCC/Clang-specific compiler extension
that MSVC does not support. This adds a note to the file-level
documentation so MSVC users are aware that format string errors will
only be caught at runtime, not at compile time.

### DIFF
--- a/absl/strings/str_format.h
+++ b/absl/strings/str_format.h
@@ -68,6 +68,13 @@
 //
 // In addition, the `str_format` library provides extension points for
 // augmenting formatting to new types.  See "StrFormat Extensions" below.
+//
+// Compiler Support Note:
+// Compile-time format string checking (e.g. detecting `absl::StrFormat("%s",
+// 1)` as an error at compile time) is supported on GCC and Clang only. MSVC
+// does not support the compiler extension required for this feature. When
+// compiling with MSVC, format string errors will only be caught at runtime,
+// not at compile time.
 
 #ifndef ABSL_STRINGS_STR_FORMAT_H_
 #define ABSL_STRINGS_STR_FORMAT_H_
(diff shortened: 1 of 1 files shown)


derekmauro (changes requested):
The added top-level "Compiler Support Note" is a bit prominent and interrupts the introductory flow of the file. It describes a platform limitation rather than focusing on how to use the library correctly.

A cleaner, less distracting approach is to integrate this platform note directly into the sections that discuss compile-time format checking. This keeps the documentation focused on the feature itself and naturally qualifies it for MSVC users.

How about around line 51: "The `FormatSpec<Args...>` template is evaluated at compile-time, providing type safety (supported on GCC and Clang; on MSVC, these checks are deferred to runtime)."

And around line 180: "Note: Compile-time format string checking is supported on GCC and Clang. On MSVC, these checks are performed at runtime instead."

DrishtiTripathi2230:
Thanks for the feedback! I've updated the PR to integrate the MSVC notes inline — one in the FormatSpec overview section and one just before the FormatSpec template declaration, as suggested.

PR #2066, opened 2026-05-29
simplify condition for ABSL_INTERNAL_STD_FILESYSTEM_PATH_HASH_AVAILABLE

h-vetinari:
The conditions for `ABSL_INTERNAL_STD_FILESYSTEM_PATH_HASH_AVAILABLE` are basically always true, since abseil now requires C++17 (and all compilers have implemented `__cpp_lib_filesystem` for a long time), plus the macOS [lower bound](https://github.com/google/oss-policies-info/blob/main/foundational-cxx-support-matrix.md) has long moved past the versions being considered here.

It could be deleted entirely depending on what's the status resp. support for XTENSA. I've left it in for now because I don't know the details there.

### DIFF
--- a/absl/hash/internal/hash.h
+++ b/absl/hash/internal/hash.h
@@ -683,10 +683,6 @@ H AbslHashValue(H hash_state, std::basic_string_view<Char> str) {
 }
 
 #if defined(__cpp_lib_filesystem) && __cpp_lib_filesystem >= 201703L && \
-    (!defined(__ENVIRONMENT_IPHONE_OS_VERSION_MIN_REQUIRED__) ||        \
-     __ENVIRONMENT_IPHONE_OS_VERSION_MIN_REQUIRED__ >= 130000) &&       \
-    (!defined(__ENVIRONMENT_MAC_OS_X_VERSION_MIN_REQUIRED__) ||         \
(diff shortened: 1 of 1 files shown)


absl/hash/internal/hash.h
```diff
@@ -682,12 +682,7 @@ H AbslHashValue(H hash_state, std::basic_string_view<Char> str) {
   return H::combine_contiguous(std::move(hash_state), str.data(), str.size());
 }
 
-#if defined(__cpp_lib_filesystem) && __cpp_lib_filesystem >= 201703L && \
```
derekmauro, line 685: Of all the PRs I've accepted, I don't think there is any I regret accepting more than adding hashing for `std::filesystem::path`. I must have wasted an entire week on workarounds for platforms that do not have a file system. I *think* it is fine to remove `ENVIRONMENT` stuff, but please restore the `__cpp_lib_filesystem` guards. I don't feel like researching it, but I think there are platforms that do not define this because they do not have a file system.

derekmauro (changes requested):

absl/hash/internal/hash.h
h-vetinari: Done.

PR #2067, opened 2026-05-29
remove ABSL_HAVE_PTHREAD_CPU_NUMBER_NP

h-vetinari:
The conditions are always satisfied based on the macOS [lower bound](https://github.com/google/oss-policies-info/blob/main/foundational-cxx-support-matrix.md) that has long moved past the versions being considered here.

### DIFF
--- a/absl/debugging/failure_signal_handler.cc
+++ b/absl/debugging/failure_signal_handler.cc
@@ -71,15 +71,7 @@
 // Checks whether pthread_cpu_number_np is available.
 #ifdef ABSL_HAVE_PTHREAD_CPU_NUMBER_NP
 #error ABSL_HAVE_PTHREAD_CPU_NUMBER_NP cannot be directly set
-#elif defined(__APPLE__) && defined(__has_include) &&              \
-    ((defined(__ENVIRONMENT_MAC_OS_X_VERSION_MIN_REQUIRED__) &&    \
-      __ENVIRONMENT_MAC_OS_X_VERSION_MIN_REQUIRED__ >= 110000) ||  \
-     (defined(__ENVIRONMENT_IPHONE_OS_VERSION_MIN_REQUIRED__) &&   \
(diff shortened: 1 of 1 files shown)


absl/debugging/failure_signal_handler.cc
```diff
@@ -66,23 +66,6 @@
 #endif
 #endif
 
-// ABSL_HAVE_PTHREAD_CPU_NUMBER_NP
-//
-// Checks whether pthread_cpu_number_np is available.
-#ifdef ABSL_HAVE_PTHREAD_CPU_NUMBER_NP
-#error ABSL_HAVE_PTHREAD_CPU_NUMBER_NP cannot be directly set
-#elif defined(__APPLE__) && defined(__has_include) &&              \
```
derekmauro, line 74: Don't we still need the `__APPLE__` test?

derekmauro (changes requested):

absl/debugging/failure_signal_handler.cc
h-vetinari: Depends a bit on the pthread setup. I can restore it, but then ABSL_HAVE_PTHREAD_CPU_NUMBER_NP won't be removable

h-vetinari:
It looks like this was reverted in https://github.com/abseil/abseil-cpp/commit/39519d75c2958969686eba78a7707d37af41960f? It's a bit confusing because the commit message is exactly the same as https://github.com/abseil/abseil-cpp/commit/bbca5fe5b0155b14f27ef962965417d787e5102a

derekmauro:
Yeah, so the story here is that now there is a bot running on our internal codebase that reverts changes that break tests. And of course there were people testing unsupported OS versions. Unfortunately the bot did not understand how to edit the part of the commit message that actually makes it to GitHub, so it just copied the original message. We will have to get that fixed.

Your other related changes are stalled for related reasons.

PR #2071, opened 2026-06-07
Include x86gprintrin.h instead of bmi2intrin.h

musicinmybrain:
GCC 16 produces an error when including `bmi2intrin.h` directly:

```
Never use <bmi2intrin.h> directly; include <x86gprintrin.h> instead.
```

See https://github.com/gcc-mirror/gcc/blob/cd33e5ad1945f19b5b1b2429ee62d1b3c8919914/gcc/config/i386/bmi2intrin.h#L24-L26.

This PR fixes failure to build with GCC 16 when the BMI2 extensions are enabled, e.g., when targeting `x86_64-v3`, by following the compiler’s advice.

### DIFF
--- a/absl/container/internal/raw_hash_set.h
+++ b/absl/container/internal/raw_hash_set.h
@@ -231,7 +231,7 @@
 #endif
 
 #ifdef __BMI2__
-#include <bmi2intrin.h>
+#include <x86gprintrin.h>
 #endif  // __BMI2__
 
 namespace absl {


absl/container/internal/raw_hash_set.h
```diff
@@ -231,7 +231,7 @@
 #endif
 
 #ifdef __BMI2__
-#include <bmi2intrin.h>
+#include <x86gprintrin.h>
```
derekmauro, line 234: This doesn't seem to work for clang. The clang version doesn't transitively include bmi2intrin.h. It looks like <immintrin.h> works on all compilers. I'll fix that for you.

musicinmybrain: That seems reasonable. Thanks for the review and revision.

Mizux:
spotted also with gcc 14 on almalinux and rockylinux in our docker based ci of google/or-tools...
see: https://github.com/google/or-tools/actions/runs/28590300063/job/84771871935

@derekmauro would be cool if you can add it to the release's "Known Issues"
https://github.com/abseil/abseil-cpp/releases/tag/20260526.0 (or enable discussion on release so we can comment it...)

derekmauro:
Added it to the known issues. We may do a patch release to include this fix.

PR #2074, opened 2026-06-10
docs: warn against using &v[0] on empty vectors in MakeSpan

### LINKED ISSUE #1334: absl::MakeSpan for vector, and the common incorrect usages
absl's MakeSpan takes only pointers, but since vector is very commonly used in C++, I've come across many use-cases where people tried to make a span for a vector in a similar way as array, and the resultant code was actually a UB. The example in the comments is for arrays: 

```
int array[3] = { 0, 0, 0 };
// Call with a [begin, end) pair.
ProcessInts(absl::MakeConstSpan(&array[0], &array[3]));
```

I've seen many people do the same for vectors, and access the element at index size(), which is UB. A better way to do that might be to use data() method of vectors, like MakeConstSpan(v.data(), v.data() + 3).

Can we have a comment saying this, or maybe we can add a method with iterators. I think even in the example in the comments, &array[3] might as well be UB, since I'm not sure about accessing the element at the size() index for an array.

Can you please share your thoughts on this?

Madhav1729:
This PR adds a warning note to the documentation of  absl::MakeSpan  and  absl::MakeConstSpan  in  absl/types/span.h  regarding a common undefined
  behavior (UB) pitfall when constructing spans from  std::vector  (or other dynamic containers).

  #### The Problem

  Developers frequently try to initialize spans using array-like pointer boundaries:

    absl::MakeSpan(&v[0], v.size());

  However, if  v  is empty, dereferencing  v[0]  to obtain its address is undefined behavior in C++. Similarly, attempting to use  &v[v.size()]  to obtain a range-end pointer also triggers UB.

  #### The Fix

  This documentation update advises developers to use  .data()  instead:

    absl::MakeSpan(v.data(), v.size());

  Or simply pass the container directly to the container overload of  MakeSpan :

    absl::MakeSpan(v);

  This addresses the common pitfall discussed in Issue #1334.

  ### Related Issues:

  Closes #1334

### DIFF
--- a/absl/types/span.h
+++ b/absl/types/span.h
@@ -719,16 +719,8 @@ constexpr bool operator>=(Span<T> a, const U& b) {
 //     return absl::MakeSpan(&array[0], num_elements_);
 //   }
 //
-// NOTE: Do not use `&v[0]` or `&v[v.size()]` to construct spans from an empty
-// or potentially empty std::vector (or other dynamic container), as
-// dereferencing at those boundaries is undefined behavior. Use `.data()`
-// instead:
-//
-//   absl::MakeSpan(my_vector.data(), my_vector.size());
-//
-// Or simply pass the container directly:
-//
-//   absl::MakeSpan(my_vector);
+// NOTE: To avoid undefined behavior if the container is empty, use `.data()`
+// or pass the container directly instead of using `&v[0]` or `&v[v.size()]`.
 //
 template <int&... ExplicitArgumentBarrier, typename T>
 constexpr Span<T> MakeSpan(T* absl_nullable ptr ABSL_ATTRIBUTE_LIFETIME_BOUND,
--- a/absl/types/span.h
+++ b/absl/types/span.h
@@ -719,6 +719,17 @@ constexpr bool operator>=(Span<T> a, const U& b) {
 //     return absl::MakeSpan(&array[0], num_elements_);
 //   }
 //
+// NOTE: Do not use `&v[0]` or `&v[v.size()]` to construct spans from an empty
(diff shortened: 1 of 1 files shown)


absl/types/span.h
```diff
@@ -719,6 +719,17 @@ constexpr bool operator>=(Span<T> a, const U& b) {
 //     return absl::MakeSpan(&array[0], num_elements_);
 //   }
 //
+// NOTE: Do not use `&v[0]` or `&v[v.size()]` to construct spans from an empty
```
derekmauro: Thanks for submitting this PR and pointing out this common undefined behavior.

To keep the headers clean and concise, we prefer to avoid large code blocks in comments when the usage is already covered by existing examples. Also, if we had to educate users about all the possible undefined behavior in C++ everywhere it might be a problem, our documentation would be overwhelming.

How do you feel about condensing the warning to a more concise version? Something like:
``` 
// NOTE: To avoid undefined behavior if the container is empty, use `.data()`
// or pass the container directly instead of using `&v[0]` or `&v[v.size()]`.
```
This still warns developers about the UB risk with  &v[0]  /  &v[v.size()]  and guides them to the safe alternatives without taking up too much vertical space in the header.

derekmauro (changes requested):

absl/types/span.h
Madhav1729: Sounds good to me. My main goal was to call attention to the UB risk with `&v[0]` / `&v[v.size()]`, and I agree the shorter version communicates that without adding too much detail to the header.

I'll update the comment accordingly. Thanks for the feedback!

PR #2080, opened 2026-06-12
Capture and emit function name in log messages

chenBright:
This PR adds the function name (`__func__`) to absl log messages, making it appear in the log prefix right after the line number, which gives a more precise callsite identifier.

```
# before
I0102 03:04:05.678900     451 foo.cc:1234] RAW: hello world

# after
I0102 03:04:05.678900     451 foo.cc:1234 MyFunc] RAW: hello world
```

### DIFF
--- a/absl/log/die_if_null.cc
+++ b/absl/log/die_if_null.cc
@@ -24,8 +24,9 @@ ABSL_NAMESPACE_BEGIN
 namespace log_internal {
 
 void DieBecauseNull(const char* absl_nonnull file, int line,
+                    const char* absl_nonnull function,
                     const char* absl_nonnull exprtext) {
-  LOG(FATAL).AtLocation(file, line)
+  LOG(FATAL).AtLocation(file, line, function)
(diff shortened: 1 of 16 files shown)


chenBright:
I have signed the CLA.

derekmauro:
Thanks for your interest in Abseil. Unfortunately there is no chance we can accept this, even though the code might be technically fine.

The major reasons why we can't accept this are:
1) Abseil's primary purpose is to be used as a dependency by Google's open source C++ projects. Since the logging library dates back to the early 2000s, the format has ossified. No doubt there is code that is parsing the output. While parsing the output of a logging library is certainly a bad idea, we don't have time to fix that.
2) Adding function names is going to cause binary bloat. This matters a lot when binaries consist of many thousands of files. This is actually one of the reasons `absl::SourceLocation` does not guarantee to expose the function name. For identifying the call site, it is mostly a nice to have. The filename and line number should be sufficient. If you want to add a function name, you can do `LOG(INFO) << __func__`.

PR #2082, opened 2026-06-15
make RoundTripDoubleToBuffer locale-independent

dxbjavid:
RoundTripDoubleToBuffer formats with snprintf %g, whose radix character follows the global C locale's LC_NUMERIC category, so in a process that has called setlocale() to something like de_DE or fr_FR absl::HighPrecision(d) comes out as e.g. "0,1". SimpleAtod only accepts '.', so the value no longer parses back to itself even though round-tripping through SimpleAtod is exactly what HighPrecision promises, and the rest of the float formatting here (RoundTripFloatToBuffer, SixDigitsToBuffer) is already locale-independent. This rewrites the radix character back to '.' in the produced buffer, and adds a regression test that exercises HighPrecision under a comma-radix locale (skipped when no such locale is installed).

### DIFF
--- a/absl/strings/numbers.cc
+++ b/absl/strings/numbers.cc
@@ -21,6 +21,7 @@
 #include <array>
 #include <cassert>
 #include <cfloat>  // for DBL_DIG and FLT_DIG
+#include <clocale>  // for localeconv
 #include <cmath>   // for HUGE_VAL
 #include <cstdint>
 #include <cstdio>
@@ -447,6 +448,18 @@ char* absl_nonnull numbers_internal::RoundTripDoubleToBuffer(
     // Should never overflow; see above.
     ABSL_ASSERT(snprintf_result > 0 &&
                 snprintf_result < numbers_internal::kFastToBufferSize);
+  }
+
+  // snprintf() writes the radix character chosen by the global C locale's
+  // LC_NUMERIC category, so a process that has called setlocale() can end up
+  // with a separator other than '.' here. The rest of Abseil's float
+  // formatting (RoundTripFloatToBuffer, SixDigitsToBuffer) is locale-
+  // independent and SimpleAtod() only accepts '.', so rewrite the radix back
+  // to '.' to keep absl::HighPrecision(double) locale-independent and
+  // round-trippable through SimpleAtod().
+  const char radix = *localeconv()->decimal_point;
+  if (radix != '.') {
+    if (char* p = strchr(buffer, radix)) *p = '.';
   }
   return buffer;
 }
--- a/absl/strings/numbers.cc
+++ b/absl/strings/numbers.cc
@@ -457,9 +457,21 @@ char* absl_nonnull numbers_internal::RoundTripDoubleToBuffer(
   // independent and SimpleAtod() only accepts '.', so rewrite the radix back
   // to '.' to keep absl::HighPrecision(double) locale-independent and
   // round-trippable through SimpleAtod().
-  const char radix = *localeconv()->decimal_point;
-  if (radix != '.') {
-    if (char* p = strchr(buffer, radix)) *p = '.';
+  // TODO: Once all supported compilers ship std::to_chars with floating-point
+  // support, use it here for inherent locale independence.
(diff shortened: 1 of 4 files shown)


absl/strings/numbers.cc
```diff
@@ -448,6 +449,18 @@ char* absl_nonnull numbers_internal::RoundTripDoubleToBuffer(
...
+  // snprintf() writes the radix character chosen by the global C locale's
+  // LC_NUMERIC category, so a process that has called setlocale() can end up
+  // with a separator other than '.' here. The rest of Abseil's float
+  // formatting (RoundTripFloatToBuffer, SixDigitsToBuffer) is locale-
+  // independent and SimpleAtod() only accepts '.', so rewrite the radix back
+  // to '.' to keep absl::HighPrecision(double) locale-independent and
+  // round-trippable through SimpleAtod().
+  const char radix = *localeconv()->decimal_point;
+  if (radix != '.') {
+    if (char* p = strchr(buffer, radix)) *p = '.';
```
derekmauro: If `localeconv()->decimal_point` is empty (which might happen in some minimal environments), `radix` will be `\0`. In this case, `strchr(buffer, '\0')` will find the NUL terminator and replace it with `.`, corrupting the string.

I also wonder if there is any chance that `localeconv()->decimal_point` will be multibyte.

absl/strings/numbers_test.cc
```diff
@@ -1717,6 +1718,36 @@ class SimpleDtoaTest : public testing::Test {
...
+  }
+  EXPECT_EQ(absl::StrCat(absl::HighPrecision(0.5)), "0.5");
+  EXPECT_EQ(absl::StrCat(absl::HighPrecision(-1.25)), "-1.25");
+  EXPECT_EQ(absl::StrCat(absl::HighPrecision(3.14159265358979)),
+            "3.14159265358979");
+  double parsed = 0;
+  EXPECT_TRUE(
+      absl::SimpleAtod(absl::StrCat(absl::HighPrecision(0.1)), &parsed));
+  EXPECT_EQ(parsed, 0.1);
+  setlocale(LC_NUMERIC, old_locale.c_str());
```
derekmauro: It would be better to use `absl::Cleanup` as an RAII to restore this around line 1728.

absl/strings/numbers.cc
```diff
@@ -448,6 +449,18 @@ char* absl_nonnull numbers_internal::RoundTripDoubleToBuffer(
     ABSL_ASSERT(snprintf_result > 0 &&
                 snprintf_result < numbers_internal::kFastToBufferSize);
   }
+
+  // snprintf() writes the radix character chosen by the global C locale's
```
derekmauro, line 453: In theory we should be able to use `std::to_chars` for locale independence, but compilers were late to implement full floating-point support, and we still support some of these older compilers. Maybe a TODO would be nice here.

derekmauro (changes requested):
This is a good catch, but I have a few comments. Let me know if you want me to make the suggested improvements, or if you would like to handle them.

absl/strings/numbers.cc
dxbjavid: good catch, handled both. it now skips an empty decimal_point so the NUL terminator can't get clobbered, and matches the whole radix string with strstr rather than a single char. for the multibyte case it rewrites the first byte to '.' and memmoves the trailing bytes away so the output collapses to a single '.'.

absl/strings/numbers_test.cc
dxbjavid: done, switched to absl::MakeCleanup so the locale is restored even if an assertion bails out early.

absl/strings/numbers.cc
dxbjavid: agreed, added a TODO noting we should move to std::to_chars once all the compilers we support have floating-point support for it.

dxbjavid:
thanks, took care of all three. pushed the radix guarding (empty + multibyte), the std::to_chars TODO, and the absl::Cleanup in the test. test still passes under a comma-radix locale here.

PR #2088, opened 2026-06-21
Initial support for GNU/Hurd

pinotree:
Tweak configuration & tests so that most of abseil works on the Hurd:
- enable available TLS (`__thread`, in practice)
- enable available mmap
- enable available POSIX semaphores
- enable available POSIX `write()`
- disable the ELF symbolizer: while the object format is ELF, the symbolizer uses bits (i.e. `mmap` files in `/proc`) not currently available
- use the available `/proc/self/exe`
- tweak a log test for the different value of `EBADF`
- tweak the strerror test for the GNU libc Hurd string representation of `-1`

### DIFF
--- a/absl/debugging/internal/symbolize.h
+++ b/absl/debugging/internal/symbolize.h
@@ -30,7 +30,7 @@
 #ifdef ABSL_INTERNAL_HAVE_ELF_SYMBOLIZE
 #error ABSL_INTERNAL_HAVE_ELF_SYMBOLIZE cannot be directly set
 #elif defined(__ELF__) && defined(__GLIBC__) && !defined(__asmjs__) \
-      && !defined(__wasm__)
+      && !defined(__wasm__) && !defined(__GNU__)
 #define ABSL_INTERNAL_HAVE_ELF_SYMBOLIZE 1
 
 #include <elf.h>
--- a/absl/log/log_modifier_methods_test.cc
+++ b/absl/log/log_modifier_methods_test.cc
@@ -218,13 +218,14 @@ TEST(TailCallsModifiesTest, WithPerror) {
       Send(AllOf(
           TextMessage(AnyOf(Eq("hello world: Bad file number [9]"),
                             Eq("hello world: Bad file descriptor [9]"),
-                            Eq("hello world: Bad file descriptor [8]"))),
+                            Eq("hello world: Bad file descriptor [8]"),
+                            Eq("hello world: Bad file descriptor [1073741833]"))),
           ENCODED_MESSAGE(HasValues(ElementsAre(
               ValueWithLiteral(Eq("hello world")), ValueWithLiteral(Eq(": ")),
               AnyOf(ValueWithStr(Eq("Bad file number")),
                     ValueWithStr(Eq("Bad file descriptor"))),
               ValueWithLiteral(Eq(" [")),
-              AnyOf(ValueWithStr(Eq("8")), ValueWithStr(Eq("9"))),
+              AnyOf(ValueWithStr(Eq("8")), ValueWithStr(Eq("9")), ValueWithStr(Eq("1073741833"))),
               ValueWithLiteral(Eq("]"))))))));
 
   test_sink.StartCapturingLogs();
--- a/absl/base/config.h
+++ b/absl/base/config.h
@@ -231,11 +231,11 @@ static_assert(ABSL_INTERNAL_INLINE_NAMESPACE_STR[0] != 'h' ||
 #endif
 
 // ABSL_HAVE_TLS is defined to 1 when __thread should be supported.
-// We assume __thread is supported on Linux when compiled with Clang or
+// We assume __thread is supported on Linux or Hurd when compiled with Clang or
 // compiled against libstdc++ with _GLIBCXX_HAVE_TLS defined.
 #ifdef ABSL_HAVE_TLS
 #error ABSL_HAVE_TLS cannot be directly set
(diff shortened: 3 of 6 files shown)


mkruskal-google:
How many more changes will be needed for this?  This looks fairly minimal, but Hurd is not in our [support matrix](https://github.com/google/oss-policies-info/blob/main/foundational-cxx-support-matrix.md), and we don't test against it.  So it may be accidentally broken again by future changes.

To evaluate whether or not this is a good idea, it would be good to see the full scope of the necessary changes.  The more you need to touch, the less likely it is to be stable

pinotree:
> How many more changes will be needed for this?

It is mostly this; I just rebased this PR with an additional fix for `absl_strerror_test` that I missed.

As for things to do:
- `NumbersTest.Atod` in `absl_numbers_test` fails, I have not investigated the reason
- I did not go through the list of skipped/disabled tests to see whether there is something important/missing; I do not expect negative surprises here though
- the `StrError` implementation and its test assume that the values of the POSIX errnos start from 1; on the Hurd they start from 0x40000001 instead, so
  - the caching done by `NewStrErrorTable` is not effective for actual errnos; that said, the fallback that calls `strerror_r()` works just fine
  - `StrErrorTest.MultipleThreads` is not really testing actual errnos

My plan is to send changes for each separately, so it is easier to review them.

> So it may be accidentally broken again by future changes.

That is fine for me. Once abseil is changed to build/work on the Hurd, I do not expect a constant stream of changes to keep it up-to-date -- followup fixes should be easy to do and to review (IMHO).

mkruskal-google:
I think for something like this it might be easier to review it in totality.  My worry is that we accept this PR (which seems pretty minimal) and then the next one seems too invasive.

pinotree:
> I think for something like this it might be easier to review it in totality. My worry is that we accept this PR (which seems pretty minimal) and then the next one seems too invasive.

Sorry for the late answer. I checked a bit more, and I think this is the biggest set of changes needed for the the Hurd. Hence it should be ready for review.

absl/debugging/internal/symbolize.h
```diff
@@ -30,7 +30,7 @@
 #ifdef ABSL_INTERNAL_HAVE_ELF_SYMBOLIZE
 #error ABSL_INTERNAL_HAVE_ELF_SYMBOLIZE cannot be directly set
 #elif defined(__ELF__) && defined(__GLIBC__) && !defined(__asmjs__) \
-      && !defined(__wasm__)
+      && !defined(__wasm__) && !defined(__GNU__)
```
mkruskal-google, line 33: Isn't this going to disable elf symbolization for clang/linux?

absl/log/log_modifier_methods_test.cc
```diff
@@ -218,13 +218,14 @@ TEST(TailCallsModifiesTest, WithPerror) {
...
-                            Eq("hello world: Bad file descriptor [8]"))),
+                            Eq("hello world: Bad file descriptor [8]"),
+                            Eq("hello world: Bad file descriptor [1073741833]"))),
           ENCODED_MESSAGE(HasValues(ElementsAre(
               ValueWithLiteral(Eq("hello world")), ValueWithLiteral(Eq(": ")),
               AnyOf(ValueWithStr(Eq("Bad file number")),
                     ValueWithStr(Eq("Bad file descriptor"))),
               ValueWithLiteral(Eq(" [")),
-              AnyOf(ValueWithStr(Eq("8")), ValueWithStr(Eq("9"))),
+              AnyOf(ValueWithStr(Eq("8")), ValueWithStr(Eq("9")), ValueWithStr(Eq("1073741833"))),
```
mkruskal-google, line 228: This feels a little magical, where does this number come from??  Do we need *tests* to pass on Hurd?  We don't run these in CI anyway and they're likely to backslide, so I'd be tempted to say we should revert all these test changes

absl/debugging/internal/symbolize.h
mkruskal-google: whoops, confused this with `__GNUC__`

mkruskal-google: whoops, confused this with `__GNUC__`

absl/log/log_modifier_methods_test.cc
pinotree: > This feels a little magical, where does this number come from??

It's simply the decimal value of `EBADF`, as I wrote in the commit/PR message:

> tweak a log test for the different value of `EBADF`

In the Hurd implementation of GNU libc, the POSIX errnos start at 0x40000001 (the first). This is because there are different ranges of error values, and the POSIX values are one of those ranges (in addition to e.g. the Mach errors, IPC errors, etc).

```
$ grep -r EBADF /usr/include/i386-gnu/
/usr/include/i386-gnu/bits/errno.h:  EBADF                          = 0x40000009,       /* Bad file descriptor */
/usr/include/i386-gnu/bits/errno.h:#define EBADF                          0x40000009

$ errno EBADF
EBADF 1073741833 Bad file descriptor
```
See also the GNU libc Hurd sources: https://sourceware.org/git/?p=glibc.git;a=blob;f=sysdeps/mach/hurd/bits/errno.h;h=3b54f5855e21091123ab0669402550968f0746cf;hb=d70dd7d72273ac1aa53b435156de3f50f0c5a868#l37

mkruskal-google: Yea my larger point is: is there any significant value to submitting fixes to get *tests* working on Hurd?  I get that it was useful to produce this PR, but you only need the prod code to work right?

The test changes seem particularly confusing without context, and like they (marginally) weaken the tests in the platforms we *do* support.

The non-test changes all look fine to me

PR #2092, opened 2026-06-27
Fix msan false positive in the swiss table

AustinSchuh:
The swiss table implementation is well optimized and uses SIMD instructions.  This doesn't play well with MSAN.  There is already a macro to ignore uninitialized warnings, treat MSAN the same way.

### DIFF
--- a/absl/container/internal/container_memory.h
+++ b/absl/container/internal/container_memory.h
@@ -479,6 +479,21 @@ struct map_slot_policy {
   _Pragma("GCC diagnostic pop")
 #define ABSL_SWISSTABLE_IGNORE_UNINITIALIZED_RETURN(x) \
   ABSL_SWISSTABLE_IGNORE_UNINITIALIZED(return x)
+#elif defined(ABSL_HAVE_MEMORY_SANITIZER)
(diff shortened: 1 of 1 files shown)


mkruskal-google:
Thanks for the PR!  Can you add a test case reproducing this problem as well?  We don't have MSAN tests for the OSS code atm, but I can verify that internally

AustinSchuh:
Thanks for the prod.  I've been trying to isolate this down to a single test I can upstream, and I think I've conflated 2 issues.  A decent number of absl tests fail when I'm building optimized with msan, but pass under fastbuild.  I think I conflated that with this fix.  This makes progress under -c opt, but has probably unintended consequences on fastbuild (ignoring more issues).  Happy to close this out if you want.

The latest https://github.com/bazel-contrib/toolchains_llvm now has support for building with msan.  That's how I'm doing my tests.

mkruskal-google:
Yea I'm just going to close this for now, feel free to reopen or file an issue if you have more info on where this is needed (or have any msan fixes).  FWIW, we probably *should* be testing msan in OSS.  I'm working on refactoring our test suite, and will try to add those in.  That should turn up any lingering issues you may have hit too

PR #2099, opened 2026-07-02
cast char to unsigned char before ctype calls in cctz parser

nabhan06:
The parse() and format() helpers in time_zone_format.cc feed a raw char straight into std::isspace() and std::isdigit() while scanning the input and format strings. When char is signed and a byte is 0x80 or higher the value is negative, which is undefined behavior for the ctype functions since the argument has to be representable as unsigned char or equal to EOF, and it can index outside the classification table on some libc implementations. This is reachable from ParseTime with any input or format containing a high-bit byte. Casting each argument to unsigned char fixes every site, the same guard already used for the ctype calls in absl/strings/escaping.cc.


derekmauro:
Please send this fix to https://github.com/google/cctz. We only import this code directly from this repository.

nabhan06:
Got it, didn't realize this was a straight import. I'll send the patch to google/cctz instead. Thanks for the pointer.

PR #2101, opened 2026-07-04
reject unpaired high surrogate in wchar_t string formatting

nabhan06:
The wide-string overload of ConvertStringArg in str_format runs each wchar_t through WideToUtf8 but never inspects the ShiftState after the loop, so a wide string ending on an unpaired UTF-16 high surrogate has only the first two bytes of a 4-byte sequence written while StrFormat("%ls", ...) still reports success, emitting truncated invalid UTF-8. The single-character path in ConvertWCharTImpl already rejects a lone high surrogate, so this adds the same saw_high_surrogate check after the loop to make the string path fail the same way. The regression test in convert_test.cc covers both a trailing high surrogate and a valid non-BMP code point so surrogate pairs keep working.

### DIFF
--- a/absl/strings/internal/str_format/arg.cc
+++ b/absl/strings/internal/str_format/arg.cc
@@ -322,6 +322,10 @@ inline bool ConvertStringArg(const wchar_t *v,
     if (chars == static_cast<size_t>(-1)) { return false; }
     chars_written += chars;
   }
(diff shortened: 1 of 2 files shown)


mkruskal-google:
This looks like a duplicate of https://github.com/abseil/abseil-cpp/pull/2093, but is missing some logic

PR #2102, opened 2026-07-08
guard size overflow in BytesToHexString and UrlEscape

nabhan06:
BytesToHexString sizes its output as 2 * from.size() and UrlEscapeInternal as initial_portion + 3 * (input.size() - initial_portion), neither of which checks for a size_t overflow, so a large enough input wraps the length and StringResizeAndOverwrite allocates a buffer smaller than the number of bytes the following loop writes into it. CEscapedLength and CalculateBase64EscapedLenInternal in this same file already guard their expansions against size_t max with ABSL_INTERNAL_CHECK, so these two newer helpers just add the matching check with the appropriate divisor.

### DIFF
--- a/absl/strings/escaping.cc
+++ b/absl/strings/escaping.cc
@@ -1177,6 +1177,8 @@ std::string HexStringToBytes(absl::string_view from) {
 
 std::string BytesToHexString(absl::string_view from) {
   std::string result;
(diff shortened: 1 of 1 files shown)


derekmauro:
Thanks!

PR #2103, opened 2026-07-09
reject invalid symbol table entry size in FindSymbol

nabhan06:
**Unchecked symbol-table entry size in FindSymbol**
`symtab->sh_entsize` is read verbatim from the object file and used as the divisor for `num_symbols` and as the per-entry read stride. A crafted ELF whose `SHT_SYMTAB`/`SHT_DYNSYM` header carries an entry size of zero divides by zero here; any other value that differs from `sizeof(ElfW(Sym))` slides the fixed-size symbol reads off the real entries. Added the same exact-size check the section-header paths already apply to `e_shentsize`, so a malformed table is skipped instead of trusted.


mkruskal-google:
Can you add a test case that demonstrates this bug (and fails without the fix)?

mkruskal-google (changes requested):
requesting changes*

nabhan06:
Added one in symbolize_test.cc (InvalidSymtabEntSizeDoesNotCrash). It builds a minimal ELF whose SHT_SYMTAB header has sh_entsize == 0, maps an executable page, and points the symbolizer at that file via a mapping hint so FindSymbol actually runs over it. Without the guard the num_symbols = sh_size / sh_entsize divide-by-zero SIGFPEs during symbolization; with it the malformed table is skipped and the address just fails to resolve. Reused the same RegisterFileMappingHint approach as SymbolizeWithMultipleMaps, and it's guarded to the ELF symbolizer.

nabhan06:
any update?

nabhan06:
any update?

PR #2106, opened 2026-07-11
cap varint length in DecodeVarint to avoid undefined shift

nabhan06:
DecodeVarint in absl/log/internal/proto.cc accumulates a base-128 varint with `value |= (byte & 0x7f) << 7 * s` but never caps the byte index, unlike the sibling Decode32Bit and Decode64Bit helpers that stop at sizeof(value). A malformed or over-long buffer whose first 10 bytes all set the continuation bit drives the shift to `<< 70` on a 64-bit value, which UBSan flags as an undefined shift exponent. This caps the loop at the 10-byte maximum of a 64-bit varint, matching the sibling decoders, with no change for well-formed input.

### DIFF
--- a/absl/log/internal/proto.cc
+++ b/absl/log/internal/proto.cc
@@ -163,12 +163,17 @@ void EncodeMessageLength(absl::Span<char> msg, const absl::Span<char> *buf) {
 
 namespace {
 uint64_t DecodeVarint(absl::Span<const char> *buf) {
+  // A base-128 varint for a 64-bit value occupies at most 10 bytes.  Stop after
+  // that many so the shift below can never reach the width of `value`, which
+  // would be undefined behavior on a malformed or over-long input.  The 64-bit
+  // and 32-bit decoders below cap their loops for the same reason.
+  constexpr size_t kMaxVarintBytes = 10;
   uint64_t value = 0;
   size_t s = 0;
   while (s < buf->size()) {
     value |= static_cast<uint64_t>(static_cast<unsigned char>((*buf)[s]) & 0x7f)
              << 7 * s;
-    if (!((*buf)[s++] & 0x80)) break;
+    if (!((*buf)[s++] & 0x80) || s == kMaxVarintBytes) break;
   }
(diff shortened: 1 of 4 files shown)


mkruskal-google:
Looks reasonable, can you add a test case that would have caught this too?

nabhan06:
Added proto_test.cc with a DecodeVarint case: a varint field whose bytes all set the continuation bit. Without the cap that drives the shift to `<< 70` on a uint64_t, so it fails under UBSan (shift exponent 70 too large, abort); with the cap it decodes fine and consumes at most the 10 varint bytes. Also wired the target into BUILD.bazel and CMakeLists.txt, plus a round-trip case as a sanity check.

absl/log/internal/proto.cc
mkruskal-google, line 17: Sorry just noticed this comment when I was wondering where this was even called from.  What issue did you hit with this?  This is definitely not a hardened protobuf parser

mkruskal-google: When I was asking for a test I was more wondering if this was reachable from public APIs.  This unit-test illustrated the theoretical problem, but it uses internal APIs that I suspect aren't actually reachable

nabhan06: No production crash behind it. I hit the `<< 70` while running the log internal code under UBSan and feeding DecodeFrom a deliberately malformed span (a varint field followed by 10 bytes that all set the continuation bit). Fair point that this isn't a hardened parser, but the loop already runs up to buf->size() specifically to tolerate junk input, and Decode32Bit/Decode64Bit right below already cap their loops to avoid the same over-wide shift. This just makes DecodeVarint consistent with them. If you'd rather not touch this path at all I'm fine closing it.

mkruskal-google: Yea I think I'd rather leave this alone if it's not reachable.  I don't think it's worth hardening this code as long as it remains unreachable.  At some point it would just become a duplicate protobuf parser that needs conformance testing :P

Thanks for the PR tho!

PR #2111, opened 2026-07-14
Use `volatile` instead of `benchmark::DoNotOptimize` as GCC optimizes awa…

SaiCharanMarrivada:
GCC optimizes away the write when we use `benchmark::DoNotOptimize` in [`stacktrace_benchmark.cc` ](https://github.com/abseil/abseil-cpp/blob/master/absl/debugging/stacktrace_benchmark.cc#L48) file.  Using volatile for the array pointer fixes this.  Compiler explorer link: https://godbolt.org/z/jzbKK3M1n


mkruskal-google:
This seems like a bug in google benchmark.  We can't accept this without setting precedent to remove every use of `DoNotOptimize` in all of our OSS repos that support gcc.

I think it would be better to file an issue at https://github.com/google/benchmark and maybe propose a fix to what seems like a gcc bug

mkruskal-google:
FWIW, it seems like this was introduced with gcc 12.  The godbolt link is useful :)

PR #2119, opened 2026-07-19
avoid partial UTF-8 sequences in wide-string AppendTruncated

nabhan06:
AppendTruncated encodes each wchar_t with WideToUtf8, which splits a surrogate pair across two calls and emits only the first two bytes of a four-byte sequence for the high half. Nothing checks that the pair completes, so a trailing unpaired high surrogate, a high surrogate followed by any other unit, or a valid pair straddling the truncation boundary each leave a partial sequence in the log record. This encodes U+FFFD for an unpaired high surrogate and lets a pending low surrogate use the two bytes its high half already reserved; the new test covers all three cases and fails on each without the fix.

### DIFF
--- a/absl/log/internal/append_truncated_test.cc
+++ b/absl/log/internal/append_truncated_test.cc
@@ -1,0 +1,77 @@
+// Copyright 2025 The Abseil Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "absl/log/internal/append_truncated.h"
+
+#include <array>
+#include <cstddef>
+#include <string>
+#include <string_view>
+
+#include "gtest/gtest.h"
+#include "absl/types/span.h"
+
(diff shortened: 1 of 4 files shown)


absl/log/internal/append_truncated_test.cc
```diff
@@ -0,0 +1,77 @@
...
+  const std::wstring src =
+      std::wstring(L"a") + wchar_t{0xD83D} + wchar_t{0xDE00};
+  // Five bytes is exactly enough for "a" plus the four-byte sequence.
+  EXPECT_EQ(Append(src, 5), "a\xF0\x9F\x98\x80");
+  EXPECT_EQ(Append(src, 6), "a\xF0\x9F\x98\x80");
+  // Four bytes is not, so the pair is dropped rather than half-written.
+  EXPECT_EQ(Append(src, 4), "a");
+}
+
+TEST(AppendTruncatedTest, PlainCharactersAreUnaffected) {
```
mkruskal-google, line 72: This test appears to be broken on windows

nabhan06: Good catch. The raw é中 literal was being decoded with the system code page on MSVC (no /utf-8), so the wchar_t values came out wrong. Switched it to universal character names (L"\u00E9\u4E2D") so the encoding no longer depends on the source file code page. The rest of the tests only use ASCII \x escapes, so they weren't affected.

nabhan06: Good catch. The é中 literal was the problem: MSVC decodes the narrow source bytes with the system code page unless /utf-8 is passed, so those wchar_t values came out wrong on Windows. I switched it to universal character names (L"\u00E9\u4E2D"), which pin the code points regardless of source or execution charset. Pushed.

nabhan06:
any update?

PR #2120, opened 2026-07-20
Fix hash_instantiated_test and hash_test on big-endian

miladfarca:
The two tests are failing on big-endian platforms after 63c77e406.

The new bit-packing code stores packed bits in the LSB of a uint64_t, then feeds raw bytes via reinterpret_cast. On big-endian, the LSB is at the end of the memory layout, so the partial-word path reads leading zero bytes instead of the actual data, causing hash collisions between distinct inputs.

Fix by byte-swapping the word to little-endian layout before feeding it to the combiner.

### DIFF
--- a/absl/hash/internal/hash.h
+++ b/absl/hash/internal/hash.h
@@ -809,7 +809,9 @@ AbslHashValue(H hash_state, const std::vector<T, Allocator>& vector) {
     for (size_t j = 0; j < 64; ++j) {
       word |= static_cast<uint64_t>(vector[i + j]) << j;
     }
-    word = absl::little_endian::FromHost64(word);
+    if constexpr (absl::endian::native == absl::endian::big) {
+      word = absl::byteswap(word);
+    }
     hash_state = combiner.add_buffer(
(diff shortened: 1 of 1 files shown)


absl/hash/internal/hash.h
```diff
@@ -809,6 +809,7 @@ AbslHashValue(H hash_state, const std::vector<T, Allocator>& vector) {
     for (size_t j = 0; j < 64; ++j) {
       word |= static_cast<uint64_t>(vector[i + j]) << j;
     }
+    word = absl::little_endian::FromHost64(word);
```
derekmauro: Thanks for the fix. The new preferred way to do this is something like:
```
if constexpr (absl::endian::native == absl::endian::big) {
  word = absl::byteswap(word);
}
```

derekmauro (changes requested):

absl/hash/internal/hash.h
miladfarca: Thank you for reviewing, I've updated it to use the new method.

PR #2125, opened 2026-08-11
fix off-by-one truncation detection in VADoRawLog and DoRawLog

nabhan06:
treat n == size as truncation in VADoRawLog and DoRawLog: vsnprintf writes only size-1 chars at that boundary, so an exactly-buffer-filling raw log message was losing its last byte and being emitted without the `(message truncated)` marker or trailing newline, letting a following line concatenate onto it.

### DIFF
--- a/absl/base/raw_logging_test.cc
+++ b/absl/base/raw_logging_test.cc
@@ -20,6 +20,13 @@
 
 #include <tuple>
 
+#if (defined(__unix__) || (defined(__APPLE__) && defined(__MACH__))) && \
+    !defined(__EMSCRIPTEN__)
+#include <unistd.h>
+
+#include <string>
+#endif
+
 #include "gtest/gtest.h"
 #include "absl/strings/str_cat.h"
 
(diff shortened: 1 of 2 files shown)


absl/base/raw_logging_test.cc
```diff
@@ -74,6 +74,31 @@ TEST(RawLoggingDeathTest, LogFatal) {
                             kExpectedDeathOutput);
 }
 
+// A message whose formatted length exactly equals the remaining buffer size is
+// still truncated: vsnprintf() only writes size-1 characters plus the NUL, so
+// truncation must be detected when the would-be length is >= the size, not only
+// when it is strictly greater.  With the off-by-one, such a message loses its
+// final character and is emitted without the "(message truncated)" marker (and
+// without the trailing newline).  A custom zero-length prefix hook makes the
+// whole raw-log buffer available so the boundary is hit deterministically.
```
mkruskal-google: This comment is describing the old buggy state, and isn't particularly useful.  I'd just delete the whole thing

nabhan06: Good call, dropped it.

absl/base/raw_logging_test.cc
```diff
@@ -74,6 +74,24 @@ TEST(RawLoggingDeathTest, LogFatal) {
                             kExpectedDeathOutput);
 }
 
+TEST(RawLoggingDeathTest, TruncationMarkerAtExactBufferBoundary) {
```
mkruskal-google: This test passes without the other changes.  Is there a bug here that's reproducible at head?

nabhan06: Yeah, the bug is real at head. With the zero-length prefix hook the whole 3000-byte buffer is free, so an exactly-3000-char message makes vsnprintf return 3000 == size. The old `n > *size` check reads that as a clean fit, so the buffer gets 2999 'x' plus the embedded NUL and is flushed with no `(message truncated)` and no trailing newline, which lets the next raw log line concatenate onto it. `n >= *size` is what reserves room for the marker.

The reason the old test passed without the fix is that it leaned on EXPECT_DEATH_IF_SUPPORTED, which no-ops when death tests aren't run in a given config, so the log statement never actually executed. I've reworked it to not depend on death tests at all: it dups STDERR_FILENO onto a pipe, logs at ERROR, and asserts on the captured bytes. That version fails at head (no marker, no newline) and passes with the fix. Verified both directions locally.

absl/base/raw_logging_test.cc
```diff
@@ -74,6 +81,50 @@ TEST(RawLoggingDeathTest, LogFatal) {
...
+  if (!absl::raw_log_internal::RawLoggingFullySupported()) {
+    GTEST_SKIP() << "Raw logging output is not supported on this platform.";
+  }
+  // kLogBufSize in raw_logging.cc; a message of exactly this length fills the
+  // buffer, so vsnprintf() reports a would-be length equal to the buffer size.
+  constexpr int kLogBufSize = 3000;
+
+  // Install a prefix hook that writes nothing so the whole raw-log buffer is
+  // available to the message and the exact boundary is hit deterministically.
+  absl::raw_log_internal::RegisterLogFilterAndPrefixHook(
```
mkruskal-google: Unfortunately this call doesn't work internally (it asserts).  We have internal hooks that invoke this and conflict here.  Can you reproduce it without an internal hook?

nabhan06: Good point, dropped the hook. The test now leaves the default prefix in place and just sizes the message to the space left after it, reconstructing "[basename : line] RAW: " from the same basename and source line the ABSL_RAW_LOG reports so the would-be length lands exactly on the buffer boundary. Confirmed it still fails at head (no marker, buffer ends in NUL instead of a newline) and passes with the fix.

absl/base/raw_logging_test.cc
mkruskal-google: This test still passes without the fix.  It might have to do with your recent changes to avoid the prefix hook, but it makes me wonder how reproducible this if w/o using internal APIs

PR #2132, opened 2026-08-18
fnmatch internal: added immediate checks to prevent unnecessary loops

Mher-DeLight:
In the FNMatch function, added a quick match-character check to prevent an unnecessary `while loop -> switch -> default -> block check` maneuver. Hopefully should make the code slightly easier to read. All tests pass.

### DIFF
--- a/absl/log/internal/fnmatch.cc
+++ b/absl/log/internal/fnmatch.cc
@@ -24,6 +24,11 @@ ABSL_NAMESPACE_BEGIN
 namespace log_internal {
 bool FNMatch(absl::string_view pattern, absl::string_view str) {
   bool in_wildcard_match = false;
+  if ((pattern.find_first_of('*') == pattern.npos) &&
+      (pattern.find_first_of('?') ==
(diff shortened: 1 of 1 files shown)


absl/log/internal/fnmatch.cc
```diff
@@ -24,6 +24,11 @@ ABSL_NAMESPACE_BEGIN
 namespace log_internal {
 bool FNMatch(absl::string_view pattern, absl::string_view str) {
   bool in_wildcard_match = false;
+  if ((pattern.find_first_of('*') == pattern.npos) &&
```
derekmauro, line 27: You can use `find_first_of("*?")` to do this in a single pass.

derekmauro (changes requested):
I actually think this is likely to be slower on realistic inputs. This is now going to pay the cost of scanning the entire string at the beginning. In realistic scenarios, the most common case is that the first character is not going to match.

Do you have data or benchmarks contradicting this?

PR #2133, opened 2026-08-18
fnmatch internal: added escape character

Mher-DeLight:
Before this change, it was impossible to match `*` and `?` literally with `FNMatch`. For example, the pattern `i love the * character` would also match `i love the foo character`.
This change adds backslash (`\`) escaping: placing a backslash before a character causes that character to be matched literally and not interpreted according to its special meaning. For example, `i love the \* character` matches the literal `*` character, not any string. (Although, since backslashes themselves need to be escaped in C++ strings, this would be written as `i love the \\* character` in source code.)
Because this is a behavioral change to an internal function and not a public API change, I wasn't sure whether this change is significant enough to require its own separate issue, so I opened this PR directly. All tests pass.

### DIFF
--- a/absl/log/internal/fnmatch.cc
+++ b/absl/log/internal/fnmatch.cc
@@ -43,10 +43,21 @@ bool FNMatch(absl::string_view pattern, absl::string_view str) {
         pattern.remove_prefix(1);
         str.remove_prefix(1);
         break;
+      case '\\':
+        pattern.remove_prefix(1);
+        if (pattern.empty()) {
+          return !str.empty() && str.front() == '\\' && str.size() == 1;
+        }
+        if (pattern.front() != str.front()) {
+          return false;
+        }
+        str.remove_prefix(1);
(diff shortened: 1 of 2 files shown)


derekmauro (changes requested):
The problem addressed by this PR is purely theoretical. I've never seen a C++ project that decided to use `*` or `?` in a filename. Besides that, there are a few critical correctness issues with this implementation:

On Windows, file paths use `\` as directory separators.

`FNMatch("*\\*", "foo*")` returns false because `case '\\'` compares against `str.front()` directly rather than searching within `str` when `in_wildcard_match == true`.

`in_wildcard_match` is never reset to `false` in `case '\\'`, causing `FNMatch("*\\ab", "aXYZb")` to return true.

The condition `!str.empty() && str.front() == '\\' && str.size() == 1` causes patterns like `*\\` matching `foo\\` to fail.

I don't think we should proceed with this PR. Let me know if you disagree.

PR #2135, opened 2026-08-22
fix greedy star matching in log fnmatch

dxbjavid:
FNMatch backs the --vmodule module patterns, and after a '*' it bound the following literal run to its first occurrence in the module name and never backtracked. So a pattern like *_test failed to match a module whose stem is unit_test_test, and the requested verbosity was silently not applied; the matcher can only produce false negatives in this case, never spurious matches. This swaps the greedy single-find loop for the usual two-pointer glob match that remembers the position just after the most recent '*' and retries on a later mismatch, and adds a regression test covering recurring literal runs after a star.

### DIFF
--- a/absl/log/internal/fnmatch.cc
+++ b/absl/log/internal/fnmatch.cc
@@ -23,50 +23,38 @@ namespace absl {
 ABSL_NAMESPACE_BEGIN
 namespace log_internal {
 bool FNMatch(absl::string_view pattern, absl::string_view str) {
-  bool in_wildcard_match = false;
-  while (true) {
-    if (pattern.empty()) {
-      // `pattern` is exhausted; succeed if all of `str` was consumed matching
-      // it.
-      return in_wildcard_match || str.empty();
-    }
-    if (str.empty()) {
-      // `str` is exhausted; succeed if `pattern` is empty or all '*'s.
-      return pattern.find_first_not_of('*') == pattern.npos;
-    }
-    switch (pattern.front()) {
-      case '*':
-        pattern.remove_prefix(1);
-        in_wildcard_match = true;
-        break;
-      case '?':
-        pattern.remove_prefix(1);
-        str.remove_prefix(1);
-        break;
-      default:
-        if (in_wildcard_match) {
-          absl::string_view fixed_portion = pattern;
-          const size_t end = fixed_portion.find_first_of("*?");
-          if (end != fixed_portion.npos) {
-            fixed_portion = fixed_portion.substr(0, end);
-          }
-          const size_t match = str.find(fixed_portion);
-          if (match == str.npos) {
-            return false;
-          }
-          pattern.remove_prefix(fixed_portion.size());
-          str.remove_prefix(match + fixed_portion.size());
-          in_wildcard_match = false;
-        } else {
-          if (pattern.front() != str.front()) {
-            return false;
-          }
-          pattern.remove_prefix(1);
-          str.remove_prefix(1);
-        }
-        break;
+  // Two-pointer glob matcher: '?' matches exactly one character and '*' matches
+  // any run of characters (including the empty run). We remember the position
+  // just after the most recent '*' so that, on a later mismatch, that '*' can
+  // consume one more character of `str` and the match be retried. The previous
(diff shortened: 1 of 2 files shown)


absl/log/internal/fnmatch.cc
```diff
@@ -23,50 +23,38 @@ namespace absl {
...
-            return false;
-          }
-          pattern.remove_prefix(1);
-          str.remove_prefix(1);
-        }
-        break;
+  // Two-pointer glob matcher: '?' matches exactly one character and '*' matches
+  // any run of characters (including the empty run). We remember the position
+  // just after the most recent '*' so that, on a later mismatch, that '*' can
+  // consume one more character of `str` and the match be retried. The previous
```
derekmauro: I've noticed that AI seems to love to add comments about "the previous implementation".

In-code comments should describe the current state and behavior of the code for future readers, unless there is a warning about how future authors should avoid bugs. Explanations of what was broken, how the previous code behaved, and why the change was made belong in the commit message and PR description.

absl/log/internal/fnmatch.cc
```diff
@@ -23,50 +23,38 @@ namespace absl {
...
+  // never backtracked, so a run that recurs later (e.g. "*test" against
+  // "testtest") was reported as a non-match even though it matches.
+  size_t p = 0;  // Current position in `pattern`.
+  size_t s = 0;  // Current position in `str`.
+  // `pattern` position just after the most recent '*', and the `str` position
+  // when it was seen; `npos` until a '*' has been encountered.
+  size_t star_p = absl::string_view::npos;
+  size_t star_s = 0;
+  while (s < str.size()) {
+    if (p < pattern.size() && pattern[p] == '*') {
```
derekmauro, line 37: I think adding some comments to the if statements would make this easier to quickly understand:

```
bool FNMatch(absl::string_view pattern, absl::string_view str) {
  // Two-pointer glob matcher: '?' matches exactly one character and '*' matches                                                                                               
  // any run of characters (including the empty run). We remember the position                                                                                                 
  // just after the most recent '*' so that, on a later mismatch, that '*' can                                                                                                 
  // consume one more character of `str` and the match be retried.                                                                                                             
  size_t p = 0;  // Current position in `pattern`.                                                                                                                             
  size_t s = 0;  // Current position in `str`.                                                                                                                                 
  // `pattern` position just after the most recent '*', and the `str` position                                                                                                 
  // when it was seen; `npos` until a '*' has been encountered.                                                                                                                
  size_t star_p = absl::string_view::npos;
  size_t star_s = 0;
  while (s < str.size()) {
    if (p < pattern.size() && pattern[p] == '*') {
      // Found '*'. Record checkpoint after '*' and advance pattern index only.                                                                                                                  
      star_p = ++p;
      star_s = s;
    } else if (p < pattern.size() &&
               (pattern[p] == '?' || pattern[p] == str[s])) {
      // Literal character match or single-character wildcard '?'.  Advance both                                                                                               
      // pattern and string pointers.                                                                                                                                          
      ++p;
      ++s;
    } else if (star_p != absl::string_view::npos) {
      // Mismatch, but a preceding '*' exists. Backtrack: reset pattern to after                                                                                               
      // the '*', and let that '*' consume one more character.                                                                                                                 
      p = star_p;
      s = ++star_s;
    } else {
      // Mismatch and no preceding '*' exists to absorb it.                                                                                                                    
      return false;
    }
  }

  // `str` is exhausted; the remainder of `pattern` must be all '*'s.                                                                                                          
  while (p < pattern.size() && pattern[p] == '*') ++p;
  return p == pattern.size();
}
```

derekmauro (changes requested):

absl/log/internal/fnmatch.cc
dxbjavid: fair point, that history belongs in the description rather than the code. i have dropped the before/after part so the comment just describes what the matcher does now.

absl/log/internal/fnmatch.cc
dxbjavid: agreed, the per-branch comments make it much easier to follow. i have taken your annotated version pretty much as-is.

PR #2138, opened 2026-08-25
Exclude Windows-specific CCTZ source files in podspec

aslam9691:
## Summary
- This PR excludes Windows-specific CCTZ source files (`absl/time/internal/cctz/src/*_win.cc`) in the generated `abseil.podspec`.

## Description
- The CCTZ time zone library in Abseil contains platform-specific source files like `absl/time/internal/cctz/src/time_zone_name_win.cc` intended only for Windows builds. When consuming Abseil via CocoaPods on Apple platforms (iOS, macOS, tvOS, watchOS, visionOS), these Windows-specific files are unnecessary and cause build errors.

## End-to-End Build Test
- Tested integration with CocoaPods client project (`gRPC Sample` on iOS Simulator `arm64` and `x86_64`).
- `xcodebuild -project Pods.xcodeproj -target abseil` passed with **`** BUILD SUCCEEDED **`**.

### DIFF
--- a/absl/abseil.podspec.gen.py
+++ b/absl/abseil.podspec.gen.py
@@ -200,8 +200,6 @@ def write_podspec_rule(f, rule, depth):
       srcs)
   # Writes dependencies of this rule.
   for dep in sorted(rule.deps):
-    if not dep.startswith("//absl/"):
-      continue
     name = get_spec_name(dep.replace(":", "/"))
     f.write("{indent}{var}.dependency '{dep}'\n".format(
         indent=indent, var=spec_var, dep=name))
--- a/absl/abseil.podspec.gen.py
+++ b/absl/abseil.podspec.gen.py
@@ -49,6 +49,7 @@ Pod::Spec.new do |s|
   s.tvos.deployment_target = '12.0'
   s.watchos.deployment_target = '4.0'
   s.visionos.deployment_target = '1.0'
+  s.exclude_files = [ 'absl/time/internal/cctz/src/*_win.cc' ]
   s.subspec 'xcprivacy' do |ss|
     ss.resource_bundles = {
       ss.module_name => 'PrivacyInfo.xcprivacy',
@@ -189,12 +190,18 @@ def write_podspec_rule(f, rule, depth):
   # Since CocoaPods treats header_files a bit differently from bazel,
   # this won't generate a header_files field so that all source_files
   # are considered as header files.
-  srcs = sorted(set(rule.hdrs + rule.textual_hdrs + rule.srcs))
+  srcs = [
+      s
+      for s in sorted(set(rule.hdrs + rule.textual_hdrs + rule.srcs))
+      if not s.endswith("_win.cc")
+  ]
   write_indented_list(
       f, "{indent}{var}.source_files = ".format(indent=indent, var=spec_var),
       srcs)
   # Writes dependencies of this rule.
   for dep in sorted(rule.deps):
+    if not dep.startswith("//absl/"):
+      continue
     name = get_spec_name(dep.replace(":", "/"))
     f.write("{indent}{var}.dependency '{dep}'\n".format(
         indent=indent, var=spec_var, dep=name))


absl/abseil.podspec.gen.py
```diff
@@ -189,12 +190,18 @@ def write_podspec_rule(f, rule, depth):
...
+      s
+      for s in sorted(set(rule.hdrs + rule.textual_hdrs + rule.srcs))
+      if not s.endswith("_win.cc")
+  ]
   write_indented_list(
       f, "{indent}{var}.source_files = ".format(indent=indent, var=spec_var),
       srcs)
   # Writes dependencies of this rule.
   for dep in sorted(rule.deps):
+    if not dep.startswith("//absl/"):
```
derekmauro: This change looks unrelated to the problem being fixed. Can you explain it to me?

aslam9691: If a dependency with a different prefix was passed to get_spec_name, it would fail the assertion and throw an AssertionError. for the safety added this check.

derekmauro: Is that actually happening? Is that preferred to a silent error?

aslam9691: It's not happening right now, I just added it defensively for edge cases. I originally thought skipping it was safer than crashing the script, but if you prefer an explicit assertion failure to catch bad inputs early, I am happy to remove this check.

derekmauro: In general, "defensive programming" is for defending against things outside of your control. Here the assertion is a good thing. We are checking our own invariants. An error would be our fault, and it is best to stop immediately with a loud error.

It also also a good rule not to make unrelated changes like this one.

Please revert this part of the change.

aslam9691: My Bad

after removing podspec generation is failing with the following error.

  File "/Users/aslamansari/Desktop/g-workspace/abseil-cpp/absl/abseil.podspec.gen.py", line 250, in <module>
    main()
  File "/Users/aslamansari/Desktop/g-workspace/abseil-cpp/absl/abseil.podspec.gen.py", line 246, in main
    generate(args)
  File "/Users/aslamansari/Desktop/g-workspace/abseil-cpp/absl/abseil.podspec.gen.py", line 225, in generate
    write_podspec(f, rules, vars(args))
  File "/Users/aslamansari/Desktop/g-workspace/abseil-cpp/absl/abseil.podspec.gen.py", line 154, in write_podspec
    write_podspec_map(f, rule_dir, 0)
  File "/Users/aslamansari/Desktop/g-workspace/abseil-cpp/absl/abseil.podspec.gen.py", line 179, in write_podspec_map
    write_podspec_map(f, value, depth + 1)
  File "/Users/aslamansari/Desktop/g-workspace/abseil-cpp/absl/abseil.podspec.gen.py", line 181, in write_podspec_map
    write_podspec_rule(f, value, depth + 1)
  File "/Users/aslamansari/Desktop/g-workspace/abseil-cpp/absl/abseil.podspec.gen.py", line 203, in write_podspec_rule
    name = get_spec_name(dep.replace(":", "/"))
  File "/Users/aslamansari/Desktop/g-workspace/abseil-cpp/absl/abseil.podspec.gen.py", line 140, in get_spec_name
    assert label.startswith("//absl/"), "{} doesn't start with //absl/".format(
AssertionError: @do_not_use_for_gloop_visibility_only//gloop/base/fprint doesn't start with //absl/

derekmauro: This is why we don't suppress errors like this. That dependency is a mistake and should be removed. 

I also just noticed that you are a Googler. Googler's are required to send changes using the internal code process. You should have noticed this in the pull request template. https://github.com/abseil/abseil-cpp/blob/master/.github/PULL_REQUEST_TEMPLATE.md

The internal change to remove the dependency is cl/973833051. After that is submitted and you remove the check I will merge this, but in the future, please use the internal code review process.

aslam9691: I have removed the check in the latest commit. Understood regarding the internal code review process—I will make sure to use it for future changes. Thanks for the review and for handling the internal dependency removal.

PR #2144, opened 2026-08-26
validate log_uniform_int_distribution stream input and harden preconditions

nabhan06:
log_uniform_int_distribution's param_type constructor computes log_range_ as static_cast<int>(ceil((1/log(base)) * log(range))). A base of 0 or 1, or a negative base for a signed IntType, violates the base > 1 precondition and makes that math non-finite: 1/log(1) divides by zero, and the final cast of inf/NaN to int is undefined behavior (UBSan: "inf is outside the range of representable values of type 'int'"). It was also reachable through operator>>, which handed whatever it parsed straight to param_type.

Changes:
- Check the param_type preconditions with ABSL_HARDENING_ASSERT.
- Only take the log() branch for base_ > 2, so an out-of-contract base never reaches the floating-point math.
- Have operator>> validate max >= min and base > 1 and set failbit on bad input instead of constructing an invalid param_type, with a test covering that rejection.

### DIFF
--- a/absl/random/log_uniform_int_distribution.h
+++ b/absl/random/log_uniform_int_distribution.h
@@ -80,7 +80,18 @@ class log_uniform_int_distribution {
         // which can eliminate some values depending on where the bounds fall.
         const double inv_log_base = 1.0 / std::log(static_cast<double>(base_));
         const double log_range = std::log(static_cast<double>(range()) + 0.5);
-        log_range_ = static_cast<int>(std::ceil(inv_log_base * log_range));
+        const double result = std::ceil(inv_log_base * log_range);
+        // A base_ of 0 or 1, or a negative base_, violates the base_ > 1
+        // precondition and leaves inv_log_base non-finite, so `result` can be
+        // inf or NaN. Casting such a value to int is undefined behavior; guard
+        // it so an out-of-contract base yields a defined (if meaningless)
+        // log_range_ instead. For a valid base_ (> 1), result is a small
+        // non-negative integer and this guard is a no-op.
+        log_range_ =
+            (result >= 0 &&
+             result < static_cast<double>((std::numeric_limits<int>::max)()))
+                ? static_cast<int>(result)
+                : 0;
       }
     }
 
--- a/absl/random/log_uniform_int_distribution_test.cc
+++ b/absl/random/log_uniform_int_distribution_test.cc
@@ -116,6 +116,33 @@ TYPED_TEST(LogUniformIntDistributionTypeTest, SerializeTest) {
   }
 }
 
+// A base of 1 (or, for signed types, a negative base) violates the base > 1
+// precondition. In debug builds the constructor asserts; in opt builds it must
+// still yield a defined object rather than invoking undefined behavior while
+// computing log_range_, which casts 1/log(base) * log(range) to int -- a cast
+// of a non-finite double (inf/NaN) to int is UB. The bad base can also reach
+// the object through operator>> reading an untrusted stream.
+TYPED_TEST(LogUniformIntDistributionTypeTest, InvalidBaseIsDefinedInOptMode) {
+#if defined(NDEBUG)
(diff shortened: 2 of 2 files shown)


absl/random/log_uniform_int_distribution.h
derekmauro, line 71: How about instead we simply change this to `base_ > 2`? This avoids all language level UB.

absl/random/log_uniform_int_distribution.h
```diff
@@ -80,7 +80,18 @@ class log_uniform_int_distribution {
         // which can eliminate some values depending on where the bounds fall.
         const double inv_log_base = 1.0 / std::log(static_cast<double>(base_));
```
derekmauro, line 81: Note that this fix fails to handle `base_ == 1`, which will produce a divide by 0 here.

absl/random/log_uniform_int_distribution.h
derekmauro, line 63: We should use ABSL_HARDENING_ASSERT to check preconditions.

absl/random/log_uniform_int_distribution_test.cc
```diff
@@ -116,6 +116,33 @@ TYPED_TEST(LogUniformIntDistributionTypeTest, SerializeTest) {
...
+#if defined(NDEBUG)
+  absl::InsecureBitGen gen;
+
+  // Direct construction with an out-of-contract base must not invoke UB.
+  absl::log_uniform_int_distribution<TypeParam> dist(0, 100, 1);
+  auto sample = dist(gen);
+  EXPECT_GE(sample, dist.min());
+  EXPECT_LE(sample, dist.max());
+
+  // The same bad base arriving through deserialization must also stay defined.
```
derekmauro, line 135: The proper fix for this is for operator>> to validate input and set std::ios_base::failbit. The wrong fix is for param_type to quietly sanitize bad input and leave the distribution in a silently failing state.

derekmauro (changes requested):
/cc @laramiel 

There is a fundamental difference between C++ language-level UB (compiler optimizations exploiting float overflow, which we want to avoid) and API contract violations.

Contract violations should fail fast and loudly. It should be very uncomfortable to have such a violation, otherwise users will eventually come to depend on them (Hyrum's Law).

We should avoid the language UB so UBSan and optimizers remain happy, but we should not promote an out-of-contract state into a tested feature.

absl/random/log_uniform_int_distribution.h
nabhan06: Done, the else is now `else if (base_ > 2)` and the cast guard is gone.

absl/random/log_uniform_int_distribution.h
nabhan06: Right, the guard only covered the cast. With the branch gated on `base_ > 2` the log() math is not reached for base 1 (or 0 / negative) at all. Verified with -fsanitize=float-divide-by-zero on top of the usual UBSan checks.

absl/random/log_uniform_int_distribution.h
nabhan06: Done, both preconditions use ABSL_HARDENING_ASSERT now.

absl/random/log_uniform_int_distribution_test.cc
nabhan06: Makes sense. operator>> now checks max >= min and base > 1 on the parsed values and sets failbit without touching the distribution. The opt-mode test is gone; the replacement only checks that operator>> rejects that input.

nabhan06:
Agreed on the contract vs. language UB distinction. Pushed a revision along those lines: ABSL_HARDENING_ASSERT for the preconditions, the log() branch gated on base_ > 2 with the cast guard dropped, and operator>> validating and setting failbit instead of handing bad input to param_type. The test no longer exercises out-of-contract construction, it just covers the operator>> rejection. Passes in fastbuild and -c opt with UBSan (including float-divide-by-zero).

PR #2154, opened 2026-09-04
debugging: define ElfW when the target has no <link.h>

### LINKED ISSUE #2153: [Bug]: ABSL_HAVE_ELF_MEM_IMAGE assumes every ELF target ships <link.h>
### Describe the issue

`ABSL_HAVE_ELF_MEM_IMAGE` is switched on for every `__ELF__` target except an explicit deny-list, and the header then unconditionally includes `<link.h>`:

  https://github.com/abseil/abseil-cpp/blob/71330b939f8b6b92d7f1b0a83e04c1288583737b/absl/debugging/internal/elf_mem_image.h#L36-L45

  `__ELF__` says the object format is ELF. It does not say a dynamic loader exists, and `<link.h>` is a glibc/dynamic-linker header, not part of the ELF format. Bare-metal ELF toolchains produce ELF objects and ship no `<link.h>` at all — `arm-none-eabi-g++` with newlib is the common one. There `ABSL_HAVE_ELF_MEM_IMAGE` is defined, `#include <link.h>` is reached, and the build stops with a fatal error.

  The deny-list is the symptom: `__QNX__`, `__asmjs__`, `__wasm__`, `__HAIKU__`, `__VXWORKS__`, `__hexagon__`, `__XTENSA__` are, for the most part, ELF targets that were added one at a time as each one hit this. Every further bare-metal ELF target has to send another PR to add itself.

  I would expect the guard to test for what the code actually needs — the availability of `<link.h>` — rather than enumerating the platforms that lack it.

Reproduces on unmodified `master` with a stock `arm-none-eabi` toolchain. No FreeRTOS, RTOS or other platform code is involved — a plain `-fsyntax-only` on Abseil's own sources is enough:

  ```
  $ git rev-parse HEAD
  71330b939f8b6b92d7f1b0a83e04c1288583737b

  $ arm-none-eabi-g++ -std=c++17 -I. -fsyntax-only absl/debugging/internal/elf_mem_image.cc
  In file included from absl/debugging/internal/elf_mem_image.cc:18:
  ./absl/debugging/internal/elf_mem_image.h:45:10: fatal error: link.h: No such file or directory
     45 | #include <link.h>  // for ElfW
        |          ^~~~~~~~
  compilation terminated.

  $ arm-none-eabi-g++ -std=c++17 -I. -fsyntax-only absl/debugging/internal/vdso_support.cc
  In file included from ./absl/debugging/internal/vdso_support.h:45,
                   from absl/debugging/internal/vdso_support.cc:19:
  ./absl/debugging/internal/elf_mem_image.h:45:10: fatal error: link.h: No such file or directory
     45 | #include <link.h>  // for ElfW
        |          ^~~~~~~~
  compilation terminated.

  $ g++ -std=c++17 -I. -fsyntax-only absl/debugging/internal/elf_mem_image.cc   # host, for contrast: compiles
  ```

  The two preconditions can be confirmed separately — the target is ELF, and it has no `<link.h>`:

  ```cpp
  #if !defined(__ELF__)
  #error NOT_ELF
  #endif
  #if __has_include(<link.h>)
  #error HAS_LINK_H
  #else
  #error NO_LINK_H
  #endif
  ```

  ```
  $ arm-none-eabi-g++ -std=c++17 -fsyntax-only t.cc
  t.cc:7:2: error: #error NO_LINK_H

  $ g++ -std=c++17 -fsyntax-only t.cc
  t.cc:5:2: error: #error HAS_LINK_H
  ```

71330b939f8b6b92d7f1b0a83e04c1288583737b (master)

Ubuntu 24.04.4 LTS, x86-64 (kernel 6.8.0-138-generic).

The failure is compile-time and host-independent; it depends only on the target toolchain being ELF without a <link.h>.

arm-none-eabi-g++ (15:13.2.rel1-2) 13.2.1 20231009 -- the toolchain that fails.

  g++ (Ubuntu 13.3.0-6ubuntu2~24.04.1) 13.3.0 -- host compiler, shown above for contrast.

None -- the errors above come from a direct -fsyntax-only compiler invocation on Abseil's own sources.

### Additional context

The one-line fix is to add the header check to the existing condition, keeping the deny-list untouched:

  ```cpp
      !defined(__XTENSA__) && __has_include(<link.h>)
  ```

  `absl/debugging/internal/stacktrace_config.h` already guards this way in the same directory (`#if __has_include(<execinfo.h>) && defined(ABSL_HAVE_THREAD_LOCAL)`), and `__has_include` is used unconditionally across Abseil since "Use __has_include unconditionally as it is supported by C++17" (2fb25663).

  Verified on `71330b93`: with this line, `elf_mem_image.cc` and `vdso_support.cc` both compile under `arm-none-eabi-g++`, and on the host `ABSL_HAVE_ELF_MEM_IMAGE` is still defined and both files still compile — so no supported platform changes behaviour.

vasko-110:
Fixes #2153.

  Per review: instead of disabling `ABSL_HAVE_ELF_MEM_IMAGE` on targets without a dynamic loader, define `ElfW` ourselves when `<link.h>` is missing. newlib ships `<elf.h>`, so that is enough to build.

  Two details differ from the snippet in the review. `__SIZEOF_POINTER__ == 8` replaces the architecture list, which covers ppc64/s390x/mips64/loongarch64 as well. And the FreeBSD branch keeps  `defined(__FreeBSD__)`: `__ELF_NATIVE_CLASS` is a glibc macro, and by the time we reach this block glibc's `<link.h>` has already defined `ElfW`, so it would never fire — while FreeBSD would silently lose `__ElfN`.

  The platform deny-list is untouched.

  Verified: `elf_mem_image.cc` and `vdso_support.cc` compile under `arm-none-eabi-g++`, and the host is unchanged — `ElfW` still comes from `<link.h>` there.

### DIFF
--- a/absl/debugging/internal/elf_mem_image.h
+++ b/absl/debugging/internal/elf_mem_image.h
@@ -42,10 +42,20 @@
 
 #ifdef ABSL_HAVE_ELF_MEM_IMAGE
 
+#if __has_include(<link.h>)
 #include <link.h>  // for ElfW
+#else
+#include <elf.h>
+#endif
 
-#if defined(__FreeBSD__) && !defined(ElfW)
-#define ElfW(x) __ElfN(x)
+#ifndef ElfW
+#if defined(__FreeBSD__)
+#define ElfW(type) __ElfN(type)
+#elif __SIZEOF_POINTER__ == 8
+#define ElfW(type) Elf64_##type
+#else
+#define ElfW(type) Elf32_##type
+#endif
 #endif
 
 namespace absl {


absl/debugging/internal/elf_mem_image.h
```diff
@@ -36,7 +36,7 @@
 #if defined(__ELF__) && !defined(__OpenBSD__) && !defined(__QNX__) &&    \
     !defined(__asmjs__) && !defined(__wasm__) && !defined(__HAIKU__) &&  \
     !defined(__sun) && !defined(__VXWORKS__) && !defined(__hexagon__) && \
-    !defined(__XTENSA__)
+    !defined(__XTENSA__) && __has_include(<link.h>)
```
mkruskal-google: Instead of just disabling this logic for bare-metal, we could add support like we did for FreeBSD.  Something like this seems like it should work, if we also condition the include of link.h:

```
#ifndef ElfW
#  if defined(__ELF_NATIVE_CLASS)
     /* FreeBSD native infrastructure uses __ElfN */
#    define ElfW(type) __ElfN(type)
#  elif defined(__x86_64__) || defined(__aarch64__) || defined(__riscv) && (__riscv_xlen == 64)
     /* Fallback block for 64-bit bare-metal environments */
#    define ElfW(type) Elf64_##type
#  else
     /* Fallback block for 32-bit bare-metal environments */
#    define ElfW(type) Elf32_##type
#  endif
#endif```

vasko-110: You're right, it builds — newlib ships `<elf.h>` even though it has no `<link.h>`, so this is enough:

  ```cpp
  #if __has_include(<link.h>)
  #include <link.h>  // for ElfW
  #else
  #include <elf.h>
  #endif

  #ifndef ElfW
  #if defined(__FreeBSD__)
  #define ElfW(type) __ElfN(type)
  #elif __SIZEOF_POINTER__ == 8
  #define ElfW(type) Elf64_##type
  #else
  #define ElfW(type) Elf32_##type
  #endif
  #endif
  ```

  Two changes from yours. `__ELF_NATIVE_CLASS` is glibc's, not FreeBSD's — glibc's `link.h` defines `ElfW` before we get here, so that branch never fires, and FreeBSD would quietly lose `__ElfN`. And   `__SIZEOF_POINTER__` saves listing ppc64/s390x/mips64/loongarch64.

  I'd still rather keep it off here, though. `VDSOSupport::Init()` needs either `getauxval` or `/proc/self/auxv`, so `IsPresent()` is always false on bare metal — that's 5 KB of .text and `open`/`read`/`close` in the link for a lookup that can't succeed.

  Either way works for me, which do you prefer?

mkruskal-google: I think either making it work for bare metal, or going your original route with some cleanup (instead of lumping on more complexity) would be preferable.  I don't have a strong preference between those, but I'd rather not add this catch-all and leave this list of platforms that presumably don't have link.h either

vasko-110: Updated.

Went with your approach. I can't test VxWorks/Hexagon/Xtensa/wasm/Haiku here, so pruning their guards would have been guesswork — this way the list doesn't need to change at all.

Kept two things from my last comment: `__SIZEOF_POINTER__ == 8` instead of the arch list, and `defined(__FreeBSD__)` for the `__ElfN` branch.

vasko-110: Anything else you'd like me to change here?

Also flagging in case it's useful: #2156 targets the same issue (#2153) with the other approach —`__has_include` in the guard, deny-list left in place. No preference from me on which lands, just didn't want you reviewing the same fix twice :)

vasko-110:
@derekmauro could you take a look? Mike reviewed this, I reworked it the way he asked, and it's been quiet since.

derekmauro (commented):
Look good, I just need to run some tests.

PR #2160, opened 2026-09-10
Recognize DragonFly BSD, and fix what keeps the tests from running on the BSDs

zakinko:
Abseil does not build on DragonFly BSD, and on the other BSDs its test
suite fails in places that have nothing to do with the code under test.
This fixes those, and reports what is left.

  - DragonFly is a FreeBSD descendant but is spelled __DragonFly__, so
    it missed every FreeBSD branch: GetTID, the auxv type names in
    vdso_support, two config macros, and cctz's _XOPEN_SOURCE, which
    hides the C99 wide functions from libstdc++'s <cwchar> there.
  - raw_logging.cc did not list NetBSD and DragonFly among the systems
    with write(), so a FATAL there aborted without its message and every
    death test failed with "died but not with expected error".
  - stripping_test could not open its own executable on the BSDs. It
    now uses KERN_PROC_PATHNAME on FreeBSD, DragonFly and NetBSD. OpenBSD
    has no such call, and the tests are skipped there.
  - convert_test skipped NaN on Apple by name because "%+f" prints "nan"
    with no sign there; the BSDs do the same, so that is now a probe.
    The hex-float rounding probe gains the case where rounding carries
    into the exponent, which FreeBSD prints as "0x1p+1" for 1.5.
  - OpenBSD links with --execute-only, which refuses the data that
    symbolize_test places in .text on purpose; that one test links with
    --no-execute-only there.
  - GrowExtremelyLargeTable reserves close to 8 GB of address space and
    now skips itself where RLIMIT_DATA is below that, which is OpenBSD's
    default.

Measured with CMake (ABSL_BUILD_TESTING, googletest at head, Release,
C++17):

    NetBSD 11.0     gcc 12.5    241 of 242
    FreeBSD 15.1    clang 19    240 of 242
    OpenBSD 7.9     clang 19    240 of 242 (a third, clock_interface_test,
                                failed once under load and passes alone)
    DragonFly 6.4   gcc 14      240 of 242 (gcc 14 from dports; Abseil
                                requires gcc 10 or later, base has 8.3)

The tests still failing are str_format_convert_test and charconv_test,
and in every case the failure is in the platform's libc, which those
tests use as the reference: FreeBSD, NetBSD and DragonFly do not type
the argument of a positional %F ("%1$*2$F"), and on FreeBSD the
vsnprintf return value then makes the test's helper allocate hundreds
of megabytes at a time, so I ran that test alone under a 4 GB address
space limit; OpenBSD prints nothing for "%#.0o" of zero; DragonFly's
"%lc" encodes a negative wchar_t instead of failing; and nan(3) and
strtod disagree on NaN payloads on FreeBSD, OpenBSD and DragonFly. Each
was confirmed with a small C program, and StrFormat and from_chars give
the C answer in every case. I have left those comparisons in place
rather than skip them; the bugs are for the BSDs to fix, and I will
report them there.

### DIFF
--- a/absl/base/config.h
+++ b/absl/base/config.h
@@ -345,7 +345,7 @@ static_assert(ABSL_INTERNAL_INLINE_NAMESPACE_STR[0] != 'h' ||
     defined(__myriad2__) || defined(__HAIKU__) || defined(__OpenBSD__) || \
     defined(__NetBSD__) || defined(__QNX__) || defined(__VXWORKS__) ||    \
     defined(__hexagon__) || defined(__XTENSA__) ||                        \
-    defined(_WASI_EMULATED_MMAN)
+    defined(_WASI_EMULATED_MMAN) || defined(__DragonFly__)
 #define ABSL_HAVE_MMAP 1
 #endif
 
@@ -357,7 +357,7 @@ static_assert(ABSL_INTERNAL_INLINE_NAMESPACE_STR[0] != 'h' ||
 #error ABSL_HAVE_PTHREAD_GETSCHEDPARAM cannot be directly set
 #elif defined(__linux__) || defined(__APPLE__) || defined(__FreeBSD__) || \
     defined(_AIX) || defined(__ros__) || defined(__OpenBSD__) ||          \
-    defined(__NetBSD__) || defined(__VXWORKS__)
+    defined(__NetBSD__) || defined(__VXWORKS__) || defined(__DragonFly__)
 #define ABSL_HAVE_PTHREAD_GETSCHEDPARAM 1
 #endif
 
--- a/absl/base/internal/raw_logging.cc
+++ b/absl/base/internal/raw_logging.cc
@@ -43,6 +43,7 @@
 // this, consider moving both to config.h instead.
 #if defined(__linux__) || defined(__APPLE__) || defined(__FreeBSD__) ||     \
     defined(__hexagon__) || defined(__Fuchsia__) || defined(__OpenBSD__) || \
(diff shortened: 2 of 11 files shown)


brad0:
> OpenBSD has no such call, and the tests are skipped there.

That is no longer true. OpenBSD now has getexecpath().

zakinko:
getexecpath() is the right call -- thanks for the correction.

It landed in 8.0, so I kept the skip on older releases and use
getexecpath() on 8.0 and later, keyed on OpenBSD >= 202610.  On 7.9 the
tests build and skip as before; on -current (OpenBSD 202610), with
getexecpath the ten StrippingTest cases run and pass instead of
skipping.  If the 8.0 release freezes OpenBSD to a lower number, tell me
and I'll move the threshold.

brad0:
> getexecpath() is the right call -- thanks for the correction.
> 
> It landed in 8.0, so I kept the skip on older releases and use getexecpath() on 8.0 and later, keyed on OpenBSD >= 202610. On 7.9 the tests build and skip as before; on -current (OpenBSD 202610), with getexecpath the ten StrippingTest cases run and pass instead of skipping.

Thanks.

> If the 8.0 release freezes OpenBSD to a lower number, tell me and I'll move the threshold.

It won't. That's the release month.

zakinko:
Rebased onto master.  The only conflict was ABSL_HAVE_MMAP, which now keys on
`__has_include(<sys/mman.h>)`, so the DragonFly line there is no longer needed
and I dropped it -- `<sys/mman.h>` is present on DragonFly, so the macro still
comes out defined.  The pthread_getschedparam line is unchanged, and the commit
message no longer claims two macros.

PR #2165, opened 2026-09-15
strings: use floating-point to_chars in RoundTripDoubleToBuffer when available

VietCT04:
RoundTripDoubleToBuffer currently uses snprintf and a parse-back check for
finite doubles. On standard libraries that provide floating-point
std::to_chars, this change uses std::chars_format::general with the existing
15/17-digit precision policy and absl::from_chars for the retry check.

The fast path is gated by standard-library support rather than compiler
version. The existing snprintf implementation remains the fallback for
unsupported standard libraries and for non-FE_TONEAREST rounding modes, where
std::to_chars would not preserve the current runtime rounding behavior.

The existing behavior for NaNs, signed zero, locale-independent output, the
DBL_MAX guard, buffer sizing, and round-trip retry semantics is preserved.

This work originated from protobuf/protobuf#29763, where a Protobuf maintainer
suggested moving the optimization into Abseil so downstream users can share the
same formatting primitive.

Correctness validation on GCC 15/libstdc++ 15:
- 10,035,982 differential double inputs: 0 byte mismatches
- 0 retry-decision differences
- C, en_US.UTF-8, and de_DE.UTF-8 locales: 0 mismatches
- FE_TONEAREST, FE_DOWNWARD, FE_UPWARD, and FE_TOWARDZERO preserve current output
- 1,035,982-value Protobuf compatibility corpus: 0 mismatches

Performance
-----------

Pinned to one CPU on an Intel i5-12500H, seven repetitions, median ns/op:

| Distribution       | Current | to_chars | Speedup |
|--------------------|--------:|---------:|--------:|
| Random finite      | 1255.17 |   224.70 | 5.59x   |
| Common decimal     |  307.36 |   108.68 | 2.83x   |
| Small magnitude    |  521.32 |   186.90 | 2.79x   |
| Large magnitude    | 1366.33 |   205.11 | 6.66x   |
| Subnormal          |  869.80 |   164.84 | 5.28x   |
| Retry-required     | 1288.46 |   234.08 | 5.50x   |
| Threshold boundary |  759.46 |   164.27 | 4.62x   |

Representative perf stat measurements also showed substantial reductions in
retired work. Across the tested distributions, CPU cycles fell by roughly
57–84% and retired instructions by roughly 59–86%.


absl/strings/numbers.cc
```diff
@@ -426,6 +477,18 @@ char* absl_nonnull numbers_internal::RoundTripDoubleToBuffer(
...
     return buffer;
   }
+
+#if ABSL_INTERNAL_HAVE_STD_TO_CHARS_FLOAT
+  // std::to_chars is locale-independent and substantially faster than the
+  // snprintf/strtod path below. It is not required to honor a runtime
+  // rounding mode, so retain the old path when callers select one other than
+  // round-to-nearest. The old path is also the fallback for standard libraries
+  // without floating-point to_chars.
+  if (std::isfinite(d) && std::fegetround() == FE_TONEAREST) {
```
derekmauro, line 487: When we wrote this I don't think we considered floating point rounding modes. `<cfenv>` is actually a banned header at Google.

I actually think `std::fegetround() == FE_TONEAREST` is the opposite of what we want. `absl::SimpleAtod()` is rounding-mode independent, so if we fallback using a non-default rounding mode, we will actually break round-tripping.

So I think this is actually broken today under non-default rounding modes. Removing the `std::fegetround() == FE_TONEAREST` test will fix this on all platforms that support floating-point to_chars.

I think we can just accept the old broken behavior on older compilers until they are deprecated and the fallback is removed.

absl/strings/numbers_test.cc
```diff
@@ -1750,6 +1750,36 @@ TEST(SimpleDtoa, HighPrecisionIsLocaleIndependent) {
...
+       "1.7976931348623157e+308"},
+  };
+
+  for (const TestCase& test_case : test_cases) {
+    char buffer[absl::numbers_internal::kFastToBufferSize];
+    absl::numbers_internal::RoundTripDoubleToBuffer(test_case.value, buffer);
+    EXPECT_STREQ(test_case.expected, buffer);
+
+    double parsed = 0;
+    EXPECT_TRUE(SimpleAtod(buffer, &parsed));
```
derekmauro, line 1778: This doesn't compile. It needs to be absl::SimpleAtod.

absl/strings/numbers_benchmark.cc
```diff
@@ -43,6 +44,33 @@ void BM_FastIntToBuffer(benchmark::State& state) {
...
+  };
+  const RoundTripDoubleBenchmarkCase& test_case =
+      kCases[static_cast<size_t>(state.range(0))];
+  state.SetLabel(test_case.label);
+
+  char buffer[absl::numbers_internal::kFastToBufferSize];
+  for (auto _ : state) {
+    benchmark::DoNotOptimize(
+        absl::numbers_internal::RoundTripDoubleToBuffer(test_case.value,
+                                                        buffer));
```
derekmauro, line 69: Likely needs `benchmark::DoNotOptimize(buffer)` too.

absl/strings/numbers.cc
```diff
@@ -411,6 +413,55 @@ char* absl_nonnull numbers_internal::FastIntToBuffer(
...
+#else
+#define ABSL_INTERNAL_HAVE_STD_TO_CHARS_FLOAT 0
+#endif
+
+#if ABSL_INTERNAL_HAVE_STD_TO_CHARS_FLOAT
+static char* absl_nonnull RoundTripDoubleToBufferWithToChars(
+    double d, char* absl_nonnull buffer) {
+  auto to_chars = [buffer, d](int precision) {
+    std::to_chars_result result = std::to_chars(
+        buffer, buffer + numbers_internal::kFastToBufferSize, d,
```
derekmauro, line 437: I think this needs to be `numbers_internal::kFastToBufferSize - 1`, because unlike `snprintf`, it doesn't NUL-terminate. Line 444 would write outside the buffer.

This isn't an issue today because `numbers_internal::kFastToBufferSize` has headroom, but we should still practice correctness.

absl/strings/numbers.cc
```diff
@@ -411,6 +413,55 @@ char* absl_nonnull numbers_internal::FastIntToBuffer(
...
     std::numeric_limits<double>::max() / 1.000000000000001;
 
+// The C++17 __cpp_lib_to_chars feature-test macro covers the integer overloads
+// too, so it cannot be used to detect floating-point to_chars.  Detect the
+// standard library implementation instead.  The compiler and standard
+// library can be selected independently, especially when Clang uses libstdc++.
+#if defined(__GLIBCXX__) && defined(_GLIBCXX_RELEASE) && \
+    _GLIBCXX_RELEASE >= 11
+#define ABSL_INTERNAL_HAVE_STD_TO_CHARS_FLOAT 1
+#elif defined(_LIBCPP_VERSION) && _LIBCPP_VERSION >= 14000
```
derekmauro, line 423: I can't tell you how many times we've broken Apple because their standard library uses availability attributes. We should make this:

```
#elif defined(_LIBCPP_VERSION) && _LIBCPP_VERSION >= 14000 && \
    (!defined(_LIBCPP_AVAILABILITY_HAS_TO_CHARS_FLOATING_POINT) || _LIBCPP_AVAILABILITY_HAS_TO_CHARS_FLOATING_POINT)
```

derekmauro (changes requested):
This is a nice improvement. I actually have a TODO for this towards the bottom of the function. We should reword it to remove the fallback when all platforms support it.

PR #2173, opened 2026-09-21
Fix crash appending/prepending large string to empty checksummed Cord

### LINKED ISSUE #2167: [Bug]: absl::Cord: Append/Prepend of a large std::string crashes after SetExpectedChecksum() on an empty cord
### Describe the issue

If an empty cord has an expected checksum set, appending or prepending an rvalue `std::string` longer than `kMaxBytesToCopy` (511 bytes) passes a null `CordRep` to `CordRepBtree::Create()`. With assertions enabled this fails `edge != nullptr` in `IsDataEdge()`; with `NDEBUG` it segfaults. I expected the append to work, as it does with 511 bytes.

It goes through `AppendTreeToTree()` (cord.cc:186), `ForceBtree()` (cord.cc:169) and `CordRepBtree::Create()`, which ends up calling `IsDataEdge(nullptr)`. `Prepend()` fails the same way through `PrependTreeToTree()`.

On an empty cord, `SetCrcCordState()` stores the checksum as a `CordRepCrc` with a null child. Most mutating paths call `MaybeRemoveEmptyCrcNode()` first, including the `string_view` overloads used for 511 bytes or less. The large-string branches of `Cord::Append(T&&)` and `Cord::Prepend(T&&)` go straight to `AppendTree()`/`PrependTree()` instead. (The header says `SetExpectedChecksum()` has no effect on an empty cord, but `ExpectedChecksum()` does return a value afterwards, so the checksum is kept.)

These two branches look like they were missed in 1db72eb03e ("Support empty Cords with an expected checksum"), which added `MaybeRemoveEmptyCrcNode()` to the other append and prepend paths. Before that commit `SetExpectedChecksum()` returned early for an empty cord, which is what the header comment still describes.

Adding the same call to both branches fixes it:

```diff
--- a/absl/strings/cord.cc
+++ b/absl/strings/cord.cc
@@ -558,6 +558,7 @@ void Cord::Append(T&& src) {
   if (src.size() <= kMaxBytesToCopy) {
     Append(absl::string_view(src));
   } else {
+    contents_.MaybeRemoveEmptyCrcNode();
     CordRep* rep = CordRepFromString(std::forward<T>(src));
     contents_.AppendTree(rep, CordzUpdateTracker::kAppendString);
   }
@@ -637,6 +638,7 @@ inline void Cord::Prepend(T&& src) {
   if (src.size() <= kMaxBytesToCopy) {
     Prepend(absl::string_view(src));
   } else {
+    contents_.MaybeRemoveEmptyCrcNode();
     CordRep* rep = CordRepFromString(std::forward<T>(src));
     contents_.PrependTree(rep, CordzUpdateTracker::kPrependString);
   }
```

With this change the example (both Append and Prepend) prints `size=512` in ASan+UBSan, assert-enabled and `NDEBUG` builds, and `absl_cord_test` passes under ASan.

`repro.cc`:

```cpp
#include <cstdio>
#include <string>

#include "absl/strings/cord.h"

int main() {
  absl::Cord cord;
  cord.SetExpectedChecksum(1);
  cord.Append(std::string(512, 'x'));
  std::printf("size=%zu\n", cord.size());
}
```

`CMakeLists.txt`, next to an `abseil-cpp` checkout:

```cmake
cmake_minimum_required(VERSION 3.16)
project(repro CXX)
set(CMAKE_CXX_STANDARD 17)
add_subdirectory(abseil-cpp)
add_executable(repro repro.cc)
target_link_libraries(repro absl::cord)
```

```sh
cmake -S . -B build -DCMAKE_CXX_COMPILER=clang++
cmake --build build --target repro
./build/repro
```

```
repro: .../abseil-cpp/absl/strings/internal/cord_data_edge.h:33: bool absl::cord_internal::IsDataEdge(const absl::cord_internal::CordRep *): Assertion `edge != nullptr' failed.
```

With `-DCMAKE_BUILD_TYPE=Release` the same program dies with a segmentation fault. `cord.Prepend(...)` instead of `cord.Append(...)` fails the same way, and `std::string(511, 'x')` works for both.

3a80a7794c7405b95dfa1f1afa26b37adf817636 (master)

Ubuntu 22.04.5 LTS, x86_64 (kernel 6.8.0)

```
Ubuntu clang version 15.0.7
Target: x86_64-pc-linux-gnu
Thread model: posix
InstalledDir: /usr/lib/llvm-15/bin
Selected GCC installation: /usr/lib/gcc/x86_64-linux-gnu/12
```

cmake version 3.22.1

### Additional context

Also reproduced with GCC 11.4 (assert-enabled and `NDEBUG` builds). On an empty cord with a checksum, the other ways of adding data that I tried (`string_view`, `const Cord&`, `Cord&&` and `CordBuffer` appends, `Prepend(const Cord&)`, `GetAppendBuffer()`, `operator=(std::string&&)`) work; only the two `std::string&&` overloads crash.

iliasabk:
## Issue

Fixes #2167 — `Cord::Append(T&&)` and `Cord::Prepend(T&&)` crash when a large rvalue `std::string` (> `kMaxBytesToCopy` = 511 bytes) is appended/prepended to an empty cord that has an expected checksum set (reported by Fengxiaoxx with assert and `NDEBUG` builds).

## Root cause

On an empty cord, `SetExpectedChecksum()` stores the checksum as a `CordRepCrc` node with a null child. The large-string branches of `Append(T&&)`/`Prepend(T&&)` go straight to `AppendTree()`/`PrependTree()`, which reach `CordRepBtree::Create()` → `IsDataEdge(nullptr)` — an assertion failure in debug builds and a segfault under `NDEBUG`. These two branches were missed in 1db72eb03e ("Support empty Cords with an expected checksum"), which added `MaybeRemoveEmptyCrcNode()` to the other append/prepend paths.

## Fix

Add the same `MaybeRemoveEmptyCrcNode()` call to both large-string branches, matching the existing `string_view` and `Cord` paths:

```cpp
} else {
  contents_.MaybeRemoveEmptyCrcNode();
  CordRep* rep = CordRepFromString(std::forward<T>(src));
  contents_.AppendTree(rep, CordzUpdateTracker::kAppendString);
}
```

This gives the >511-byte path identical semantics to the <=511-byte path: the empty CRC node is dropped and the data is appended normally.

## Test

New `AppendLargeStringToEmptyCrcCord` and `PrependLargeStringToEmptyCrcCord` in `cord_test.cc`: `SetExpectedChecksum(1)` on an empty cord, then append/prepend a 512-byte string. On the unfixed code both abort in `IsDataEdge()` (`cord_data_edge.h:33`).

## Verification

- Reporter's reproducer (assert-enabled build):
  - Unpatched: `Assertion failed: (edge != nullptr), function IsDataEdge, cord_data_edge.h:33` (abort); same for `Prepend`.
  - Patched: prints `size=512` for both `Append` and `Prepend`.
- `absl_cord_test`: 217/217 pass, 0 failures (2 pre-existing skips), including the 2 new tests.


derekmauro:
Thank you for trying to fix this. This was fixed by https://github.com/abseil/abseil-cpp/commit/ecb078564503dad2d4cd8658e59c064c4e0e5f87. Your fix is correct, but the added tests are different.

nvnzno-a11y:
Thanks for the heads-up — glad it's fixed upstream now. Appreciate the review!

